PR #11, opened 2017-06-29
Add ssl_ca_cert option to backend

tomyan:
Make it possible to configure the CA cert used to verify the backend
for a Fastly service.

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -271,6 +271,12 @@ func resourceServiceV1() *schema.Resource {
 							Description: "SSL certificate hostname",
 							Deprecated:  "Use ssl_cert_hostname and ssl_sni_hostname instead.",
 						},
+						"ssl_ca_cert": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Default:     "",
+							Description: "CA certificate attached to origin.",
+						},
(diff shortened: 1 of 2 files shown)


fastly/resource_fastly_service_v1.go
```diff
@@ -271,6 +271,12 @@ func resourceServiceV1() *schema.Resource {
 							Description: "SSL certificate hostname",
 							Deprecated:  "Use ssl_cert_hostname and ssl_sni_hostname instead.",
 						},
+						"ssl_ca_cert": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Default:     "",
```
catsby, line 277: Is the empty string default here needed?

catsby (commented):
Question about the empty default, does it serve a purpose or was a test failing without it?

fastly/resource_fastly_service_v1.go
tomyan: The tests pass without the default value. I put it in to match the style of similar optional parameters - e.g. `ssl_cert_hostname`.

Thanks

Tom

tomyan:
@catsby just bumping the main thread in case you didn't see my response to your comment in the code. Let me know if it looks okay, or whether any changes are needed. Thanks

catsby (approved):
Thanks!

PR #15, opened 2017-07-06
gcs_logging - Optionally use env variable for credentials (originally hashicorp/terraform#14587)

catsby:
Manually copied over the diff from hashicorp/terraform#14587


catsby:
This is just a copy-over of a user PR that I approve 👍 so I'm going to merge

Issue #18, opened 2017-07-11
There is no way to associate health checks with backend origin servers


dbeckham:
### Terraform Version

Terraform v0.9.11

### Affected Resource(s)
- fastly_service_v1

### Terraform Configuration Files
```hcl
resource "fastly_service_v1" "test" {
  name = "test"

  domain {
    name = "test.example.com"
  }

  backend {
    address          = "1.1.1.1"
    name             = "east"
    port             = 80
  }

  backend {
    address          = "2.2.2.2"
    name             = "west"
    port             = 80
  }

  healthcheck {
    name              = "beacon"
    host              = "test.example.com"
    path              = "/beacon.gif"
    method            = "HEAD"
    expected_response = 200
  }

}
```
### Expected Behavior
There should be a parameter to allow a health check to be associated with a specific backend.

### Actual Behavior
There is no way to associate a defined health check with a backend. The backend definitions in VCL end up being generated with no `.probe = { }` block.  Health check definitions are effectively ignored.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Define a `fastly_service_v1` resource similar to the example above, and run `terraform apply`.

dbeckham:
I'm closing this as it's a documentation problem, not a code problem. PR #20 has fixes for the backend healthcheck docs.

PR #19, opened 2017-07-11
fastly_service_v1 documentation fixes

dbeckham:
These changes fix two documentation issues:

* The documentation sidebar incorrectly lists the `fastly_service_v1` resource as `service_v1`.
* The `host` argument for the `healthcheck` block does not match the Fastly configuration interface, nor it's actual behavior. (This may be why issue #18 exists as the author misunderstood the host argument's usage.)


radeksimko (approved):
Thanks and sorry for the delay.

PR #37, opened 2017-08-17
Add support for Fastly dictionaries

msuterski:
This is an attempt to add support for Fastly dictionaries to the provider (https://github.com/terraform-providers/terraform-provider-fastly/issues/5). Any change to the dictionary, including changing values, is forcing the dictionary to be recreated for the new version of the service. It has it's pros and cons. I can list them here if it was helpful.

The acceptance test is failing due to provider trying to add a dictionary with an ID of the one that was previously deleted and without a name. It seems like if the diff between the new Set and the Old Set is not being calculated correctly (ID is a value provided in the API response and marked as `Computed` field in the Schema).

I'd appreciate any help if possible...

The scenario is:
1. Add a dictionary
2. Add 2 items to the dictionary (forces new dictionary to be created)
3. Add 3rd item to the dictionary (forces new dictionary to be created)

Initially, the TF seem to be calculating the diff correctly

```
UPDATE: fastly_service_v1.foo
  dictionary.3135423306.items.%:    "2" => "0"
  dictionary.3135423306.items.key1: "item1" => ""
  dictionary.3135423306.items.key2: "item2" => ""
  dictionary.3135423306.name:       "dict_w" => ""
  dictionary.48981542.id:           "" => "<computed>"
  dictionary.48981542.items.%:      "0" => "3"
  dictionary.48981542.items.key1:   "" => "item1"
  dictionary.48981542.items.key2:   "" => "item2"
  dictionary.48981542.items.key3:   "" => "item3"
  dictionary.48981542.name:         "" => "dict_w"
```

Once we get to actually get the diffs between new set and the old set (to remove and add dictionaries), the `new` set includes an object that only has the `ID` value in it. The `ID` value is of the previously removed dictionary from the delete portion of the run.

```
2017/08/17 14:45:52 [DEBUG] Fastly Service dictionary: *Set(map[string]interface {}{"3161630230":map[string]interface {}{"name":"", "id":"1rSqbdegBmmvqSufvWe4E4", "items":map[string]interface {}{}}, "48981542":map[string]interface {}{"name":"dict_w", "id":"", "items":map[string]interface {}{"key3":"item3", "key1":"item1", "key2":"item2"}}})
2017/08/17 14:45:52 [DEBUG] Fastly Dictionary old: *Set(map[string]interface {}{"3135423306":map[string]interface {}{"id":"1rSqbdegBmmvqSufvWe4E4", "items":map[string]interface {}{"key2":"item2", "key1":"item1"}, "name":"dict_w"}})
2017/08/17 14:45:52 [DEBUG] Fastly Dictionary new: *Set(map[string]interface {}{"48981542":map[string]interface {}{"name":"dict_w", "id":"", "items":map[string]interface {}{"key3":"item3", "key1":"item1", "key2":"item2"}}, "3161630230":map[string]interface {}{"items":map[string]interface {}{}, "name":"", "id":"1rSqbdegBmmvqSufvWe4E4"}})
2017/08/17 14:45:52 [DEBUG] Fastly Dictionary to remove: []interface {}{map[string]interface {}{"id":"1rSqbdegBmmvqSufvWe4E4", "items":map[string]interface {}{"key1":"item1", "key2":"item2"}, "name":"dict_w"}}
2017/08/17 14:45:52 [DEBUG] Fastly Dictionary to add: []interface {}{map[string]interface {}{"items":map[string]interface {}{}, "name":"", "id":"1rSqbdegBmmvqSufvWe4E4"}, map[string]interface {}{"name":"dict_w", "id":"", "items":map[string]interface {}{"key2":"item2", "key3":"item3", "key1":"item1"}}}
2017/08/17 14:45:52 [DEBUG] Fastly Dictionary removal opts: fastly.DeleteDictionaryInput{Service:"1PHcWaNGskDpGrUTC9BsV6", Version:3, Name:"dict_w"}
2017/08/17 14:45:52 [DEBUG] Fastly Dictionary creation opts: fastly.CreateDictionaryInput{Service:"1PHcWaNGskDpGrUTC9BsV6", Version:3, Name:""}
2017/08/17 14:45:52 [DEBUG] root: eval: *terraform.EvalWriteState
2017/08/17 14:45:52 [DEBUG] root: eval: *terraform.EvalApplyProvisioners
2017/08/17 14:45:52 [DEBUG] root: eval: *terraform.EvalIf
2017/08/17 14:45:52 [DEBUG] root: eval: *terraform.EvalWriteState
2017/08/17 14:45:52 [DEBUG] root: eval: *terraform.EvalWriteDiff
2017/08/17 14:45:52 [DEBUG] root: eval: *terraform.EvalApplyPost
2017/08/17 14:45:52 [ERROR] root: eval: *terraform.EvalApplyPost, err: 1 error(s) occurred:

* fastly_service_v1.foo: 400 - Bad Request
Message: Name can't be blank
2017/08/17 14:45:52 [ERROR] root: eval: *terraform.EvalSequence, err: 1 error(s) occurred:

* fastly_service_v1.foo: 400 - Bad Request
Message: Name can't be blank
2017/08/17 14:45:52 [TRACE] [walkApply] Exiting eval tree: fastly_service_v1.foo

```

The same sort of thing happens when I'm removing dictionaries. When I update my `.tf` file from 2 dictionaries to 1, the `add new` set of dictionaries includes one with no name and old ID (I did not add a test for it with this commit since yet).

I could check if the dictionary name filed is not empty before creating any, but that seems to be missing the point of having correct diffs. 

I also tried wrapping the items map in a Set. Same behavior. 

I'm probably doing something wrong here and it would be great if someone could point me in a right direction.

Even though it's not per TF standards, I included part of the functionality in a separate "dictionaries" go file for clarity. I'll move it to the main file once I can figure out the Schema sets diff problem.

Thanks a lot.


msuterski:
@radeksimko hey, is there a place where I could get some help figuring out why those schema diffs don't work for this implementation?

I'm sort of bumping my head against a wall with this and can't find what I should do differently.

Thanks.

msuterski:
OK. I got this thing working, but I'm still not sure why the `items` object needs to be set as `Calculated`. Is there a place where I could find more details about what this flag actually does? All the schema helper doc says is that when an object is marked as `Calculated`, it's going to be calculated. Not very helpful ;)

This should be ready for review now

msuterski:
```
% make testacc                                                                                                                                                                      2 ↵
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v  -timeout 120m

=== RUN   TestAccFastlyIPRanges
--- PASS: TestAccFastlyIPRanges (0.61s)
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestProvider_impl
--- PASS: TestProvider_impl (0.00s)
=== RUN   TestAccFastlyServiceV1CacheSetting_basic
--- PASS: TestAccFastlyServiceV1CacheSetting_basic (45.74s)
=== RUN   TestAccFastlyServiceV1_conditional_basic
--- PASS: TestAccFastlyServiceV1_conditional_basic (24.43s)
=== RUN   TestResourceFastlyFlattenDictionary
--- PASS: TestResourceFastlyFlattenDictionary (0.00s)
=== RUN   TestAccFastlyServiceV1_dictionary
--- PASS: TestAccFastlyServiceV1_dictionary (148.78s)
=== RUN   TestResourceFastlyFlattenGCS
--- PASS: TestResourceFastlyFlattenGCS (0.00s)
=== RUN   TestAccFastlyServiceV1_gcslogging
--- PASS: TestAccFastlyServiceV1_gcslogging (23.32s)
=== RUN   TestAccFastlyServiceV1_gcslogging_env
--- PASS: TestAccFastlyServiceV1_gcslogging_env (22.38s)
=== RUN   TestFastlyServiceV1_FlattenGzips
--- PASS: TestFastlyServiceV1_FlattenGzips (0.00s)
=== RUN   TestAccFastlyServiceV1_gzips_basic
--- PASS: TestAccFastlyServiceV1_gzips_basic (49.84s)
=== RUN   TestFastlyServiceV1_BuildHeaders
--- PASS: TestFastlyServiceV1_BuildHeaders (0.00s)
=== RUN   TestAccFastlyServiceV1_headers_basic
--- PASS: TestAccFastlyServiceV1_headers_basic (47.20s)
=== RUN   TestAccFastlyServiceV1_healthcheck_basic
--- PASS: TestAccFastlyServiceV1_healthcheck_basic (45.58s)
=== RUN   TestAccFastlyServiceV1_logentries_basic
--- PASS: TestAccFastlyServiceV1_logentries_basic (47.61s)
=== RUN   TestAccFastlyServiceV1_logentries_formatVersion
--- PASS: TestAccFastlyServiceV1_logentries_formatVersion (21.57s)
=== RUN   TestAccFastlyServiceV1_papertrail_basic
--- PASS: TestAccFastlyServiceV1_papertrail_basic (44.88s)
=== RUN   TestAccFastlyServiceV1RequestSetting_basic
--- PASS: TestAccFastlyServiceV1RequestSetting_basic (24.89s)
=== RUN   TestAccFastlyServiceV1_response_object_basic
--- PASS: TestAccFastlyServiceV1_response_object_basic (45.69s)
=== RUN   TestAccFastlyServiceV1_s3logging_basic
--- PASS: TestAccFastlyServiceV1_s3logging_basic (49.87s)
=== RUN   TestAccFastlyServiceV1_s3logging_domain_default
--- PASS: TestAccFastlyServiceV1_s3logging_domain_default (23.61s)
=== RUN   TestAccFastlyServiceV1_s3logging_s3_env
--- PASS: TestAccFastlyServiceV1_s3logging_s3_env (21.43s)
=== RUN   TestAccFastlyServiceV1_s3logging_formatVersion
--- PASS: TestAccFastlyServiceV1_s3logging_formatVersion (20.63s)
=== RUN   TestResourceFastlyFlattenSumologic
--- PASS: TestResourceFastlyFlattenSumologic (0.00s)
=== RUN   TestAccFastlyServiceV1_sumologic
--- PASS: TestAccFastlyServiceV1_sumologic (24.73s)
=== RUN   TestAccFastlyServiceV1_syslog_basic
--- PASS: TestAccFastlyServiceV1_syslog_basic (45.90s)
=== RUN   TestAccFastlyServiceV1_syslog_formatVersion
--- PASS: TestAccFastlyServiceV1_syslog_formatVersion (24.34s)
=== RUN   TestResourceFastlyFlattenDomains
--- PASS: TestResourceFastlyFlattenDomains (0.00s)
=== RUN   TestResourceFastlyFlattenBackend
--- PASS: TestResourceFastlyFlattenBackend (0.00s)
=== RUN   TestAccFastlyServiceV1_updateDomain
--- PASS: TestAccFastlyServiceV1_updateDomain (45.94s)
=== RUN   TestAccFastlyServiceV1_updateBackend
--- PASS: TestAccFastlyServiceV1_updateBackend (45.46s)
=== RUN   TestAccFastlyServiceV1_basic
--- PASS: TestAccFastlyServiceV1_basic (23.72s)
=== RUN   TestAccFastlyServiceV1_disappears
--- PASS: TestAccFastlyServiceV1_disappears (9.81s)
=== RUN   TestAccFastlyServiceV1_defaultTTL
--- PASS: TestAccFastlyServiceV1_defaultTTL (69.29s)
=== RUN   TestAccFastlyServiceV1_VCL_basic
--- PASS: TestAccFastlyServiceV1_VCL_basic (45.92s)
=== RUN   TestValidateLoggingFormatVersion
--- PASS: TestValidateLoggingFormatVersion (0.00s)
=== RUN   TestValidateLoggingMessageType
--- PASS: TestValidateLoggingMessageType (0.00s)
PASS
ok  	github.com/terraform-providers/terraform-provider-fastly/fastly	1043.335s
```

catsby:
Hello! Thanks for the work here. I have a question or 2:

- reading the API docs it seems Dictionaries can be edited without requiring a new VCL version to be validated/activated, is this correct? 
- Assuming so, I would imagine we'd want to allow users to update these dictionaries _outside_ of the current resource, e.g. make our own `fastly_dictionary` Resource, with required `service_id` and likely(?) `version_number` parameters

The way I see the code, now, changes to dictionaries require complete new service version, which I don't know is the best model. What do you think?

msuterski:
@catsby what you're proposing would be ideal. Same would apply to snippets, etc. 

The problem is with creating/destroying those objects. You can manage their contents without touching versions, but you can not create or attach them without creating new version of the service. The workflow for creating a dictionary is:
1) Clone an active version of the service (requires touching of the `fastly_service_v1`)
2) Create a dictionary for the new version of the service
3) Activate the new version (changes the state of the `fastly_service_v1`). 
4) Add items to the dictionary....

If we were to make dictionaries a standalone resource it means that creating and destroying them would need to apply changes to the `fastly_service_v1` state in the background. I'm not sure if it would be a good practice (?). I was under the impression that resources _should_ stay contained and not affect other resources. Though, the Fastly service is somewhat "different" in general.

We certainly can call the clone and activate endpoints when creating/destroying dictionaries and get the service ID + increase the version.

If you think it would be a better approach, I'd happy to rework this PR 👍  It would certainly "fit" better into the Fastly workflow.

thommahoney:
@msuterski @catsby Thank you both for putting so much thought into this implementation.

After talking it over with @phamann, we think it makes sense to follow @catsby's suggestion of making dictionaries a top-level resource.

As @msuterski suggests, the Dictionary itself will need to be managed (create, update, delete) within the fastly_service_v1 resource. We can then expose the IDs of these service-versioned Dictionaries as exported attributes of the resource.

The DictionaryItems will then become a new top-level resource that references the Dictionary by ID.

As an example (pseudocode):

```hcl
resource "fastly_service_v1" "demo" {
  name = "demofastly"

  dictionary {
    name = "my_dictionary"
  }
}

resource "fastly_dictionary_items_v1" "my_dictionary_items" {
  dictionary_id = "${fastly_service_v1.demo.dictionaries["my_dictionary"]}"

  items {
    foo = "bar"
  }
}
```

Our only concern of this pattern is that dictionary items mutated outside of terraform (ie. via our API directly) might cause confusion. We'll just need to document this warning that if you're using Terraform to manage dictionaries, then you should only use Terraform to manage your dictionaries.

One other concern is about write-only dictionaries (which are very new). We won't be able to manage write-only dictionaries with Terraform because you simply cannot see what's currently in the API at all.

msuterski:
Thanks @thommahoney It looks to me like a good plan.

The TF refreshes its state before each action, so it's going to import "current" state of the resources before anything happens. The plan output should reflect/include items that were managed outside of TF. But I agree that a note should be clear about it.

Do we need/want to keep `_v1` in the new resource name?

I noticed that there is also a new `dictionary info` endpoint that provides a hash and number of items in a dictionary (I assume to make life easier with the write only ones). If  we used the same hashing function in TF, we could manage the write only dictionaries in TF, too. Though, I'd like to keep it out of scope for the initial implementation. 

I'll have few weeks off in November and can continue working on this PR.

catsby:
Hey friends! I'm going to drop out of this conversation, it seems you all have a solid plan in place. I will part saying that we should probably stick with the `_v1` part of the name for now, at least for consistency. Thanks!

msuterski:
Hey @thommahoney @catsby 

I'm struggling with how to design the schema for the dictionary so that it's accessible by name (or something descriptive) so that what you suggested above would work

```hcl
resource "fastly_dictionary_items_v1" "my_dictionary_items" {
  dictionary_id = "${fastly_service_v1.demo.dictionaries["my_dictionary"]}"
}
```

It seems to me that only top level resources can be referenced by name (`my_dictionary`). The inner resources are just elements of lists and can not be assigned names. They could probably be accessed by an index in the list, but that's not very helpful in this case.

something like 

```hcl
resource "fastly_service_v1" "service" {
  name = "s"
  dictionary "my_dictionary" {
    name = "my_dictioanry_name"
  }
}
```
 is not a valid syntax.

Could you point me to an example schema definition that exposes inner resources by name, if possible?

thommahoney:
Sorry for the delayed response. I'm hoping to have an answer for you soon.

msuterski:
@thommahoney thanks.

How about we scope this PR to manage just `dictionary` elements and deal with items later once we can get better handle on them? Having at least `dictionary` objects managed by TF would solve 25% of my current requirements :)

minichate:
Hey folks! Anything I can do to help get this merged? I've got a rebased branch at
https://github.com/minichate/terraform-provider-fastly/tree/dictionaries which just resolves the current merge conflicts. Happy to help with testing -- though it seems to work fairly well in the current build as-is!

thommahoney:
@minichate Looks like there is a conflict on this branch so that would need to be resolved at a minimum. Judging by the conversation above it isn't clear what the direction of this PR should be. It seems we can more easily support creation / deletion of the dictionary itself without adding support for the (versionless) dictionary items.

msuterski:
@thommahoney I'd be happy to move forward with this PR but I need some guidance from the maintainers where we want to go.

Dictionaries is something I still would like to have supported in the provider and I was waiting for your feedback for my questions here https://github.com/terraform-providers/terraform-provider-fastly/pull/37#issuecomment-341536231

I could also scope this change to just dictionaries (a partial win), but that also is something I need a decision from the maintainers. 

Thanks!

thommahoney:
> How about we scope this PR to manage just dictionary elements and deal with items later once we can get better handle on them?

@msuterski This sounds good for now. If you bring this PR in line with that goal I will happily review.

chrismikehogan:
Is this effort still active, or has it died out?

msuterski:
It died out. I moved on to work on something else. Sorry.

In summary, the way Fastly's API is structured, how objects are referenced (i.e. by mutable names vs something like immutable IDs) and relationships between versions of the service vs versions of the dictionaries, makes it difficult to manage them without certain compromises or restrictions.

I hope the Fastly team will be able to provide better solution in the provider to manage dictionaries _and_ items, since they also manage the API.

msuterski:
Sorry. I won't be able to spend more time on this at this time.

phamann:
@msuterski @chrismikehogan @minichate We now have a working implementation in https://github.com/terraform-providers/terraform-provider-fastly/pull/178 which would be great to get your feedback on.

chrismikehogan:
Thanks @phamann! I'm not longer on a team that manages Fastly -- @taer this something y'all might be interested in? 
> We now have a working implementation in #178 which would be great to get your feedback on.

msuterski:
thanks @phamann!

I'll ping the team that manages fastly setup currently.

PR #41, opened 2017-08-25
Fix support for GCS logging, and add message_type support

spheromak:
Adds message_type to the gcs logging. As well as fixes some bugs that prevented gcs logging from actually working correctly. 

This depends on changes in sethvargo/go-fastly introduced in 9e9a7d46e796 
 I re imported that library with govendor to point at master for that repo. 

If you don't want to merge this cause of the vendoring thats fine. Let me know what the path forward should be tho.


joemiller:
TODO: add message_type to website-docs

joemiller:
ping @sethvargo or other maintainers - We are also interested in this PR as it both adds missing `message_type` attr for gcslogging but also fixes a bug with the existing gcslogging support that made it unusable for us. Any eyes on this or feedback on what is missing would be much appreciated.

udryan10:
@joemiller or @spheromak what was the behavior of the gcslogging bug you guys ran into? I'm wondering if we happen to be running into the same thing with something weird we're seeing

msuterski:
@udryan10 The behavior is that if adding/changing/removing gcs is the only change TF sees in a plan it won't actually apply the change in Fastly (make API calls), but will tell you it succeeded.

spheromak:
@udryan10 what @msuterski  said is pretty much right on. its allso in the commit comment from fcb3959 
```
Without this patch GCS logging wouldn't apply correctly or at all
depending on the starting state of the service. This adds all the
documented fields to the resource and correctly invokes set for the
gcslogging attribute.
```

When reading from the service it was incorrectly reading into a `gcs` key where `gcslogging` is used everywhere else. Also was not considering gcs at all when looking for resource that might need to be updated. As well as not having all of the fields correctly populated for gcslogging. 

Most the broken bits are in that commit diff: 
https://github.com/terraform-providers/terraform-provider-fastly/pull/41/commits/fcb39598ea4d50de63ff015eafdf3e75dd2f4dc5#diff-e11f97f0c242e77100f2f794b0ac08abR2290

spheromak:
@radeksimko tho this adds an enhancement this is really a bug fix as the feature as implemented on the released version is totally broken for managing GCS logging in fastly.

spheromak:
ping @sethvargo anything preventing this from merging at this point ?  We have been running this modification in production for while, and the master branch is totally broken for GCS without these changes.

thommahoney (approved):

PR #43, opened 2017-09-15
Added support for custom Directors

hynd:
Hello!
Custom director's aren't particularly well documented on Fastly's site yet, but essential for anything other than per-request random balancing across origins.

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -340,6 +340,68 @@ func resourceServiceV1() *schema.Resource {
 				},
 			},
 
+			"director": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "A name to refer to this director",
+						},
+						"backends": {
+							Type:        schema.TypeSet,
+							Required:    true,
+							Description: "List of backends associated with this director",
+							Elem:        &schema.Schema{Type: schema.TypeString},
+						},
+						// optional fields
+						"capacity": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Default:     100,
+							Description: "Load balancing weight for the backends",
+						},
+						"comment": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"shield": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Default:     "",
+							Description: "Selected POP to serve as a 'shield' for origin servers.",
+						},
+						"quorum": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Default:     75,
+							Description: "Percentage of capacity that needs to be up for the director itself to be considered up",
+							ValidateFunc: func(v interface{}, k string) (ws []string, es []error) {
+								if v.(int) < 0 || v.(int) > 100 {
+									es = append(es, fmt.Errorf("Fastly Director quorum should be a percentage between 0 and 100; found %d", v.(int)))
+								}
+								return
+							},
+						},
+						"type": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							Default:      1,
+							Description:  "Type of load balance group to use. Integer, 1 to 4. Values: 1 (random), 3 (hash), 4 (client)",
+							ValidateFunc: validateDirectorType,
+						},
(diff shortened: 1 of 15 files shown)


website/docs/r/service_v1.html.markdown
```diff
@@ -119,6 +117,40 @@ resource "fastly_service_v1" "demo" {
 }
 ```
 
+Basic usage with [custom Director](https://docs.fastly.com/api/config#director):
```
thommahoney, line 122: Thanks for your thorough docs. I may not hold onto this section specifically long term. I think we can document each of the API resources without providing full examples for each of them.

thommahoney (dismissed):

mgue:
Hi @thommahoney any idea when we can get this merged?

thommahoney:
Sorry for the delayed response. Hoping to have an answer for you soon.

hynd:
Hi @thommahoney - any progress on an answer for this?  Thanks!

lanej:
@hynd if you can resolve the conflicts, we can re-review this PR.  Thanks for your patience.

hynd:
Sure can! Rebased...

fastly/resource_fastly_service_v1.go
```diff
@@ -1350,7 +1405,7 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 					Name:    cf["name"].(string),
 				}
 
-				log.Printf("[DEBUG] Fastly Conditions Removal opts: %#v", opts)
+				log.Printf("[DEBUG] Condition Removal opts: %#v", opts)
```
lanej: While I do agree that adding 'Fastly' to the log line is redundant, it has no relation to the intended change and should be done in another PR with a dedicated purpose.

The same is true for the remaining changes below.

website/docs/r/service_v1.html.markdown
```diff
@@ -217,6 +250,15 @@ used in the `request_condition`, `response_condition`, or
 * `priority` - (Optional) A number used to determine the order in which multiple
 conditions execute. Lower numbers execute first. Default `10`.
 
+The `director` block supports:
```
JakeChampion, line 255: Should the director block support all the fields that the director api has available? https://docs.fastly.com/api/config#director

lanej (dismissed):

website/docs/r/service_v1.html.markdown
lanej: That would be ideal but not required (cc/ @hynd).  I would prefer to prioritize the relief of the users for features that can easily be augmented in the future.

hynd: Good spot - `comment` was missing from the doc, i *think* the only thing actually missing from the code is the `shield` field (which isn't in the underlying go-fastly lib.... yet)

hynd: Ah, and `capacity` is only half-implemented in go-fastly - i'll PR fixes for both of those first, and will update this accordingly....

lanej: Thanks @hynd

hynd: Done - merged upstream and refreshed this PR

lanej:
Got an acceptance test failure:

```
--- FAIL: TestAccFastlyServiceV1_directors_basic (18.96s)
	testing.go:434: Step 0 error: Check failed: Check 2/4 error: Bad match Director match, expected (&fastly.Director{ServiceID:"3UIdDTdTSwAh1AZ1rw20RS", Version:1, Name:"mydirector", Comment:"", Shield:"", Quorum:0x0, Type:0x3, Retries:0x0, Capacity:0x0, DeletedAt:"", CreatedAt:"", UpdatedAt:""}), got (&fastly.Director{ServiceID:"3UIdDTdTSwAh1AZ1rw20RS", Version:1, Name:"mydirector", Comment:"", Shield:"", Quorum:0x4b, Type:0x3, Retries:0x5, Capacity:0x64, DeletedAt:"", CreatedAt:"2018-09-19T17:35:52Z", UpdatedAt:"2018-09-19T17:35:52Z"})
```

Looks like the timestamps don't match.  I haven't dug into the reason but if you could investigate, I would appreciate it.

hynd:
Fixed!  Also removed the type "2" (round-robin) director, which seems to have been recently dropped from the docs (and throws an HTTP 400).

lanej (approved):

lanej:
Thanks @hynd

PR #45, opened 2017-09-21
add `activate` argument

thommahoney:
This new argument conditionally prevents the service from being activated. By default
it maintains the previous activation behavior.

### todo

- [x] add documentation

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -42,6 +42,13 @@ func resourceServiceV1() *schema.Resource {
 				Computed: true,
 			},
 
+			"activate": {
+				Type:        schema.TypeBool,
+				Description: "Conditionally prevents the Service from being activated",
+				Default:     true,
+				Optional:    true,
+			},
+
 			"domain": {
 				Type:     schema.TypeSet,
 				Required: true,
@@ -2653,18 +2660,23 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 			return fmt.Errorf("[ERR] Invalid configuration for Fastly Service (%s): %s", d.Id(), msg)
 		}
 
-		log.Printf("[DEBUG] Activating Fastly Service (%s), Version (%v)", d.Id(), latestVersion)
-		_, err = conn.ActivateVersion(&gofastly.ActivateVersionInput{
-			Service: d.Id(),
-			Version: latestVersion,
-		})
-		if err != nil {
-			return fmt.Errorf("[ERR] Error activating version (%d): %s", latestVersion, err)
-		}
+		shouldActivate := d.Get("activate").(bool)
+		if shouldActivate {
+			log.Printf("[DEBUG] Activating Fastly Service (%s), Version (%v)", d.Id(), latestVersion)
+			_, err = conn.ActivateVersion(&gofastly.ActivateVersionInput{
+				Service: d.Id(),
+				Version: latestVersion,
+			})
+			if err != nil {
+				return fmt.Errorf("[ERR] Error activating version (%d): %s", latestVersion, err)
+			}
 
-		// Only if the version is valid and activated do we set the active_version.
-		// This prevents us from getting stuck in cloning an invalid version
-		d.Set("active_version", latestVersion)
+			// Only if the version is valid and activated do we set the active_version.
+			// This prevents us from getting stuck in cloning an invalid version
+			d.Set("active_version", latestVersion)
+		} else {
+			log.Printf("[DEBUG] Skipping activation of Fastly Service (%s), Version (%v)", d.Id(), latestVersion)
+		}
 	}
 
 	return resourceServiceV1Read(d, meta)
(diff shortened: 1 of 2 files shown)


catsby (commented):
I'm not sure how this resource in it's current design will handle this. 

- Assume `activate` is false, and we make changes to domains or backends (or anything)
- Terraform applies these changes to a new version, but `activate` is false, so the new version is created and validated but not activated. 
- Future terraform plan/refresh will read the `active_version` from config and query the API for info on that version, and save the results to state
- Those results are different from what was just put into the config but not activated, so Terraform shows a diff
- Terraform never returned the version number that was created but not activated, so you can’t simply activate it from terraform

jeanpaulgalea:
What if `terraform apply` shows a unified diff between the current active VCL and the newly generated (but not yet activated) VCL? Once the diff pager is exited, a confirmation is required on whether to activate the new VCL or not.

Perhaps have a `.tf` directive to override this behaviour and activate automatically.

tenzone:
has there been movement on this.  It will be a big help to be able to run a diff on the site before activating a version, especially when making a big change. for some services, i have been "breaking" it by using an invalid logging format version, but thats hackey and not all my services have the same logging parameters.

neogenix:
Ditto, for us it'd allow us to handle activate, and deactivate in another flow which we'd far prefer.

philippschulte:
This feature definitely has some downsides but a lot of users have either requested to be able to activate the Terraform generated draft version manually or complained about the confusing Terraform diff output due to the provider's architecture. Some users are not able to use the current implantation of the provider because of their security compliances which prohibits automatic version activation. For all these reasons it’s worth supporting this feature because it will allow a user to use the Fastly web application or the Fastly API to review and compare the draft version and to activate it manually.

@catsby is completely right about the behavior if the new `activate` attribute is set to `false` but you will experience the same behavior if Fastly throws an error on Terraforms attempt to activate the newly generated draft version. This happens for instance for any `VCL` compilation errors which of course get only validated on the Fastly side. Actually, @tenzone uses this behavior especially to prevent Terraform from activating the new draft version in order to be able to activate it manually.

Here is a list of all the cons and pros for the new `activate` attribute:

Cons:

1. If you set `activate=false` and you run `$ terraform apply` you’ll lose state because the source is always the active version. So if you run `$ terraform plan`  directly after you have run `$ terraform apply` you’ll see the same diff output again because it’s comparing your Terraform definition against the active version. But you can easily sync state after you have activated the generated draft version manually by using the Terraform `import` command (`$ terraform import fastly_service_v1.resource_name service_id`). 
2. If you set `activate=false` you’ll have to activate the draft version manually either via the Fastly web application or the Fastly API. After you run  `$ terraform apply`  you’re not able to activate the draft version by running  `$ terraform apply` again. Each time you run `$ terraform apply` it will create a new draft version and there won’t be any difference between the drafts because the source of the state is the active version.
3. The new `activate` attribute won’t be included when importing state via the Terraform CLI (`$ terraform import fastly_service_v1.resource_name service_id`). The reason is that the new `activate` attribute is not a Fastly resource. The same applies to the `force_destroy` attribute. Terraform doesn’t support the creation of attributes just for behavior purposes. By design, Terraform wants to track all attributes in the state. However, the purpose of the `activate` and the `force_destroy` attributes are just to control behavior and not state. If you use the `activate` or the `force_destroy` attributes in your Terraform definition and you want to import state via the Terraform CLI then you’ll need to add the attributes to the `terraform.tfstate` manually. Otherwise  `$ terraform plan` will list differences even if you just imported the state and you haven’t made any changes to your Terraform configuration.

Pros:

1. The user can choose between automatic (`activate=true`) and manual (`activate=false`) draft version activation. By default  (`activate=true`) the provider will activate the new generated draft version automatically and there won’t be any differences to the original behavior when running `$ terraform apply`.
2. The user has the ability to compare the generated draft version against any other version before activating it. The user can either use the Fastly web application or the Fastly API for a version comparison.

fastly/resource_fastly_service_v1.go
```diff
@@ -2653,18 +2660,26 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
...
 
-		// Only if the version is valid and activated do we set the active_version.
-		// This prevents us from getting stuck in cloning an invalid version
-		d.Set("active_version", latestVersion)
+			// Only if the version is valid and activated do we set the active_version.
+			// This prevents us from getting stuck in cloning an invalid version
+			d.Set("active_version", latestVersion)
+		} else {
+			log.Printf("[INFO] Skipping activation of Fastly Service (%s), Version (%v)", d.Id(), latestVersion)
+			log.Print("[INFO] The Terraform definition is explicitly specified to not activate the changes on Fastly")
```
chenrui333, line 2679: This line should be also `log.Printf`?

philippschulte: [Printf](https://golang.org/pkg/fmt/#Printf) formats according to a format specifier and writes to standard output. The string: `The Terraform definition is explicitly specified to not activate the changes on Fastly` doesn't need to be formatted. Therefore we just need to use [Print](https://golang.org/pkg/fmt/#Print) in this case in order to write the string to standard output.

Here is an example of how the info messages are formatted in standard output:

```
[INFO] Skipping activation of Fastly Service (5XTpVB5M69L4UBQBlX70zE), Version (2)
[INFO] The Terraform definition is explicitly specified to not activate the changes on Fastly
[INFO] Version (2) has been pushed and validated
[INFO] Visit https://manage.fastly.com/configure/services/5XTpVB5M69L4UBQBlX70zE/versions/2 and activate it manually
```

PR #56, opened 2017-11-02
Sumologic update

msuterski:
This fixes a problem where changes to existing `sumologic` definition would not be pushed to Fastly.


msuterski:
@thommahoney I was wondering if this could be reviewed. We're bumping into this problem all over again.

Thanks

jrynyt:
this also prevents adding a new sumologic configuration to an existing fastly service.

thommahoney:
Thanks for fixing this!

PR #58, opened 2017-11-20
Owasp

tkuhlman:
This pull request adds initial WAF and OWASP functionality to the terraform provider.

I expect this to be the first of a series of changes for this functionality to be complete. I'm pretty disappointed with the difficultly of testing this functionality and expect that when Fastly makes it possible to enable WAF from the API the tests will need reworking and that in turn may lead to additional work with the rest of the WAF/OWASP implementation.


fastly/owasp.go
```diff
@@ -0,0 +1,250 @@
...
+			},
+			"xss_score_threshold": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
+		},
+	},
+}
+
+// expandOWASP converts a matchine schema to a gofastly OWASP.
```
jnazarioFastly: typo - matchine -> machine

josehelps:
@tkuhlman guessing there is no expectation of rule management, or having the provider start the WAF container with a specific rule set as it has not been implemented in the `go-fastly` library yet.  Also, concern that if a customer attempts to use this and does not have the feature flag on in the Fastly side it will fail, would we want to catch those failures?

tkuhlman:
Rule management can be added on top of this but is out of scope for this PR, it is being worked on by a colleague of mine. This will fail for anyone without WAF enabled on the Fastly side but if I remember correctly gives a pretty clear error message and so I don't think anything explicit needs to be done to catch it.

PR #60, opened 2017-11-23
Add tls_hostname support for syslog.

jeanpaulgalea:
Add support for `tls_hostname` in syslog block. 

Simply passing the value to the vendor'd go-fastly `TLSHostname`.


jeanpaulgalea:
BTW, if it makes any difference, I've successfully used this with production systems.

Obviously made sure that `syslog { tls_hostname = "dummyname"; }` as specified in the `.tf` file shows up correctly on Fastly's https://manage.fastly.com panel.

thommahoney (approved):
Thanks!

PR #61, opened 2017-11-23
Add support for {max,min}_tls_version.

jeanpaulgalea:
Add support for `max_tls_version` and `min_tls_version` in backend block.

Simply passing the values to the vendor'd go-fastly `MaxTLSVersion` and `MinTLSVersion`.


jeanpaulgalea:
BTW, if it makes any difference, I've successfully used `min_tls_version` with production systems.

Made sure that `backend { min_tls_version = "1.2"; }` as specified in the `.tf` file shows up correctly on Fastly's https://manage.fastly.com panel.

However, I didn't test `max_tls_version`... although following the code paths, it seems equivalent to `min_tls_version`.

thommahoney (approved):
Thanks!

jeanpaulgalea:
Awesome, thanks!

Any hint as to when the next version will be released? ;-)

Kind of sucks having to build the fastly provider from master instead of letting terraform do its magic and pull the latest release :-)

PR #62, opened 2018-01-06
Override user-agent

### LINKED ISSUE #42: Override user-agent
We should override the user-agent header from the go-fastly library to be "terraform" instead of "go-fastly". This will help us better understand who and how is using the Terraform provider in logs:

https://github.com/sethvargo/go-fastly/blob/ffa250f2871d7785b003a21a24a34e8cf005ae95/fastly/client.go#L40-L42

aerostitch:
Hi!

@sethvargo in #42 you were talking about using just "terraform" as a user agent. I saw that there's a more standard user agent inside the terraform package so I used that. Let me know if that fits what you had in mind.

Thanks
Joseph

Closes #42

### DIFF
--- a/fastly/config.go
+++ b/fastly/config.go
@@ -3,6 +3,7 @@ package fastly
 import (
 	"fmt"
 
+	"github.com/hashicorp/terraform/terraform"
 	gofastly "github.com/sethvargo/go-fastly/fastly"
 )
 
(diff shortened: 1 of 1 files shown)


aerostitch:
Note: I wanted to add a test for this using something like:
```go
if gofastly.UserAgent != terraform.UserAgentString() {
	t.Fatalf("User Agent incorrectly setup. Got: %s, expecting %s\n", gofastly.UserAgent, terraform.UserAgentString())
 }
```
But I'm not sure where to put it. It needs to be done only during acceptance tests as the config is not executed otherwise. But I can't put it in `testAccPreCheck` either as it is ran before the initialization of the provider configuration.

Putting it in a resource test seems a bit much.Maybe if we had an interface that we inplement for the fastly client it would be possible to unit test Config() but I'm not sure there's a real point for that.

Any thoughts?

Thanks for your help,
Joseph

PR #71, opened 2018-02-18
fix base_url implementation

thommahoney:
this adds base_url to the resource schema to avoid acceptance test
failures

this also makes it so that the base_url can be configured using an
environment variable `FASTLY_API_URL`


thommahoney:
@whiteley could you take a look at this? The PR that added this functionality wasn't quite finished but this should complete the implementation.

tombuildsstuff (approved):
LGTM 👍🏼

PR #80, opened 2018-05-02
Adding bigquery logging

bridgetlane:

PR #83, opened 2018-05-29
Add bq logging remove history

paultyng:
Updated PR for #80

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -793,6 +793,50 @@ func resourceServiceV1() *schema.Resource {
 				},
 			},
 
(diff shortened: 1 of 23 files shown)


paultyng:
@bridgetlane I had to set a default on `format` to get the tests to pass, otherwise they would fail with a pending diff.

It seems like [per the docs](https://docs.fastly.com/api/logging#logging_bigquery) though that format may be required as the JSON schema needs to match your table?

> Apache style log formatting. Must produce JSON that matches the schema of your BigQuery table

I'm not really familiar with this, so would appreciate your input on how to handle.

paultyng:
closing, got push to work to existing branch

PR #89, opened 2018-08-21
Handle deletion of already deleted or never created resources

lanej:
Detect a not found response with the resource deletion conditions and consider a not found response a success condition.


thommahoney (approved):
Thanks!

PR #92, opened 2018-09-03
Add explanation for importing service

sakajunquality:
Although terraform import works for fastly service, it is not documented.
I think it's nice to have explanations for importing.


lanej (commented):
Thanks @sakajunquality

lanej (approved):

PR #93, opened 2018-09-04
Add support for VCL Snippets

### LINKED ISSUE #7: [fastly][fastly_service_v1] support vcl snippets
_This issue was originally opened by @mtougeron as hashicorp/terraform#14600. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

It would be great if the `fastly_service_v1` resource supported the [VCL Snippets](https://docs.fastly.com/guides/vcl-snippets/). The way the `fastly_service_v1` resource works now we have to update the entire `VCL`, including the boilerplate, in order to modify just part of the logic (e.g., a Header). This prevents us from being able to use the `VCL Snippets` and simplifying our code-base & workflow.

hynd:
Closes #7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 0.3.1 (Unreleased)
 
+IMPROVEMENTS:
+
+* [GH-90] Add `template` to Google BigQuery logging
+
 BUG FIXES:
 
 * Handle deletion of already deleted or never created resources ([#89](https://github.com/terraform-providers/terraform-provider-fastly/pull/89))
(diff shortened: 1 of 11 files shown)


lanej (approved):
Looks solid to me :+1:

*note*: regular snippet support only.  Dynamic support would require adding an additional attribute

https://docs.fastly.com/api/config#snippet

cruzanmo:
@lanej does there need to be another version release for this? `terraform init` installs v0.3.0 and gives an error of `invalid or unknown key: snippet`.

PR #96, opened 2018-09-26
add log placement support

josehelps:
The Fastly WAF uses the placement functionality to determine if the logging endpoint is a WAF logging endpoint or not. A WAF logging endpoint is unique in the fact that has unique fields like waf.logdata and waf.rule available to log for one or many WAF event in a single request.  

You can find out more information how placement is used for WAF at: https://docs.fastly.com/guides/web-application-firewall/fastly-waf-logging#waf_debug_log

I am not 100% sure if I should update testing for other endpoints, please advice. Also updated the go-fastly library to have the underlying calls for the logging integration also understand placement. Some already supported it but not all, see: https://github.com/sethvargo/go-fastly/pull/88 for details. Thank you

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -680,10 +680,10 @@ func resourceServiceV1() *schema.Resource {
 							Description: "The S3 redundancy level.",
 						},
 						"placement": {
-						Type:        schema.TypeString,
-						Optional:    true,
-						Description: "The placement for the logging endpoint. Can be ",
-					},
+							Type:        schema.TypeString,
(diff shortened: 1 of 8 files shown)


lanej:
@d1vious The error should be more obvious now.  `Placement` support in go-fastly needs to be added in other logging providers AND the create/update logging input structs.

josehelps:
@lanej it does thank you, just created a pull to correct these at https://github.com/sethvargo/go-fastly/pull/89

philippschulte:
I'll close this PR in favor of https://github.com/terraform-providers/terraform-provider-fastly/pull/106

PR #97, opened 2018-09-29
Add argument reference for bigquerylogging

tjun:
`bigquerylogging` option is missing from "Argument Reference" in the doc.


lanej:
Thanks @tjun

PR #106, opened 2018-12-06
add support for the placement field to all logging providers

philippschulte:
Currently, none of the logging providers support the `placement` field in order to be able to place the logging call in a specific VCL subroutine.

At the moment Fastly provides two different values for the `placement` field:

1. `none` - doesn't place the logging call in any VCL subroutine
2. `waf_debug` - places the logging call in the VCL `waf_debug_log` subroutine

There is already a [PR (#96)](https://github.com/terraform-providers/terraform-provider-fastly/pull/96) for adding the `placement` field to the logging providers. However, this one covers all providers and has better validation. Also, this one updates the docs with the proper descriptions.

The build is currently failing because `go-fastly` needs to be updated in order to support the `placement` field. For the work of [PR (#96)](https://github.com/terraform-providers/terraform-provider-fastly/pull/96) there have been already opened and merged two PRs:

1. [Added log placement support for integrations that were missing it #88](https://github.com/sethvargo/go-fastly/pull/88)
2. [Add log placement #89](https://github.com/sethvargo/go-fastly/pull/89)

The problem with these two PRs is that only the `response` struct or the `request` struct has been updated in order to accept the `placement` field. The PR [add placement field to all missing providers #92)](https://github.com/sethvargo/go-fastly/pull/92) for the `go-fastly` client adds the `placement` field to the remaining structs in order to pass the build.

### DIFF
--- a/fastly/validators.go
+++ b/fastly/validators.go
@@ -32,6 +32,20 @@ func validateLoggingMessageType(v interface{}, k string) (ws []string, errors []
 	return
 }
 
+func validateLoggingPlacement(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+	validTypes := map[string]struct{}{
+		"none":      {},
+		"waf_debug": {},
+	}
+
+	if _, ok := validTypes[value]; !ok {
+		errors = append(errors, fmt.Errorf(
+			"%q must be one of ['none', 'waf_debug']", k))
+	}
+	return
+}
+
 func validateDirectorType(v interface{}, k string) (ws []string, errors []error) {
 	value := uint(v.(int))
 	validVersions := map[uint]struct{}{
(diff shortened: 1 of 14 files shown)


fastly/validators.go
```diff
@@ -32,6 +32,20 @@ func validateLoggingMessageType(v interface{}, k string) (ws []string, errors []
 	return
 }
 
+func validateLoggingPlacement(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+	validTypes := map[string]struct{}{
```
lanej, line 37: this could be pulled out into a package-local variable

lanej (approved):
One small note but otherwise looks good.  This is purely an additive change.

PR #113, opened 2019-01-04
allow terraform changes to be pushed to fastly as draft only, without activation

marcb1:
For some changes, the terraform diff is completely unreadable (see https://github.com/terraform-providers/terraform-provider-fastly/issues/104). For these changes, it is sometimes easier to make the changes in Fastly as draft only, then view the diff on Fastly and activate it there if it looks good (or re-activate it in terraform)

Could not figure out how to make `draft_only` as a command line argument to `terraform apply` so I made it part of the schema. 

The way this works, is when `draft_only` is set in the schema:
- the change is made in fastly.com, and a new draft version is created
- the change is not applied to the terraform state file
- if the user is happy with the change (after looking at the diff on fastly.com), they can re-run `terraform plan && terraform apply` with `draft_only` set to false.
- the same change, is now re-created and activated in fastly via terraform.
- terraform state file is updated

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -42,6 +42,15 @@ func resourceServiceV1() *schema.Resource {
 				Computed: true,
 			},
 
+			// Allow user to specify a draft only change, so the terraform apply does not activate the version in fastly.
+			// This is sometimes useful as fastly diffs can be easier to read than terraform diffs.
+			"draft_only": {
+				Type:        schema.TypeBool,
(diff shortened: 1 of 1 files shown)


philippschulte:
Hi there :wave:

Thanks for your contribution but there is already a PR https://github.com/terraform-providers/terraform-provider-fastly/pull/45 for this feature! Your implementation is slightly different but it does the same trick!

I'll close this one in favor of https://github.com/terraform-providers/terraform-provider-fastly/pull/45.

PR #121, opened 2019-01-25
Fix VCL snippet acceptance test

radeksimko:
This acceptance test was failing previously:

```
=== RUN   TestAccFastlyServiceV1Snippet_basic
--- FAIL: TestAccFastlyServiceV1Snippet_basic (0.01s)
	testing.go:434: Step 0 error: Configuration is invalid.
		
		Warnings: []string(nil)
		
		Errors: []string{"fastly_service_v1.foo: : invalid or unknown key: vcl_snippet"}
FAIL
```
and it appears it never actually passed (since it was added to the repo).

## After patch

```
=== RUN   TestAccFastlyServiceV1Snippet_basic
--- PASS: TestAccFastlyServiceV1Snippet_basic (94.18s)
```

### DIFF
--- a/fastly/resource_fastly_service_v1_snippet_test.go
+++ b/fastly/resource_fastly_service_v1_snippet_test.go
@@ -23,11 +23,17 @@ func TestAccFastlyServiceV1Snippet_basic(t *testing.T) {
 		Content:  "if ( req.url ) {\n set req.http.my-snippet-test-header = \"true\";\n}",
 	}
 
-	s2 := gofastly.Snippet{
+	updatedS1 := gofastly.Snippet{
+		Name:     "recv_test",
+		Type:     gofastly.SnippetTypeRecv,
+		Priority: int(110),
+		Content:  "if ( req.url ) {\n set req.http.different-header = \"true\";\n}",
+	}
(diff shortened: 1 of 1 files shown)


philippschulte:
@radeksimko the PR looks good to me. I guess you need to add me to the reviewers. Otherwise, I can't approve it. I had to wrap my head around it because of these statements:

```
// We don't know these things ahead of time, so ignore them
lr.ID = ""
lr.CreatedAt = ""
lr.UpdatedAt = ""
```

All Fastly resources have the same metadata but I couldn't find any other tests where these fields have been set to the zero values. I was wondering why all the other tests didn't fail. The reason is the inconsistency of the `go-fastly` HTTP client response structs. These metadata fields are part of the `Snippet` response struct and it makes total sense that we have to override them before we are able to compare them because we don't know these values ahead of time.

It's acutally very bad that this got merged into `master` without passing the tests! The acceptence tests should run on Travis and PRs should only be allowed to be merged into `master` if the tests pass. Maybe Travis wasn't set up that way because the tests take a long time to run?

philippschulte (approved):
Looks good to me 👍

PR #131, opened 2019-03-19
Increase default health check interval

vespa59:
Health checks are run from every one of Fastly's ~60 POPs, so an interval if 5s results in 20/s for each POP, or roughly 1200 requests per minute. A more "sane" default would be once per minute.

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -141,7 +141,7 @@ func resourceServiceV1() *schema.Resource {
 						"check_interval": {
 							Type:        schema.TypeInt,
 							Optional:    true,
-							Default:     5000,
+							Default:     60000,
 							Description: "How often to run the healthcheck in milliseconds",
 						},
 						"expected_response": {


philippschulte:
👋 @vespa59 ,

I agree with @thommahoney that we should keep the default value of `5000` milliseconds:

**Request**:
```
curl -X POST \
  https://api.fastly.com/service/<service_id>/version/<version_no>/healthcheck \
  -H 'Content-Type: application/json' \
  -H 'Fastly-Key: <your_key>' \
  -d '{
	"name": "your_healthcheck"
}'
```
**Response**:
```
{
    "name": "your_healthcheck",
    "service_id": "<service_id>",
    "version": <version_no>,
    "threshold": 3,
    "window": 5,
    "http_version": "1.1",
    "timeout": 500,
    "method": "HEAD",
    "updated_at": "2019-03-28T15:57:48Z",
    "expected_response": 200,
    "deleted_at": null,
    "path": "/",
    "host": "",
    "created_at": "2019-03-28T15:57:48Z",
    "comment": "",
    "check_interval": 5000,
    "initial": 2
}
```

Across the entire code base, we stick to the default values of the Fastly API. Therefore, I will close this PR. Thanks for your contribution.

PR #133, opened 2019-03-25
Support for ACLs and ACL entries

kamsz:
#9

### DIFF
--- a/fastly/provider.go
+++ b/fastly/provider.go
@@ -27,8 +27,9 @@ func Provider() terraform.ResourceProvider {
 			"fastly_ip_ranges": dataSourceFastlyIPRanges(),
 		},
 		ResourcesMap: map[string]*schema.Resource{
-			"fastly_service_v1": resourceServiceV1(),
-			"fastly_acl_v1":     resourceACLV1(),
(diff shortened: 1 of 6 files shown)


kamsz:
There's one issue in this PR I've noticed today - there's a race condition while creating more than one ACL due to version cloning and I'm not sure how to handle that. Right now we're doing dependencies between ACLs that are going to modify the same service.

philippschulte:
👋Hello! Thanks for the work here. The implementation of ACL Entries, Edge Dictionary Entries, and Dynamic Snippets in Terraform is quite challenging. The reason is that these resources are versionless so they can be modified without activating a new version.

I haven't reviewed the entire code yet but I have noticed that you split ACLs and its Entries into two separate resources. I recommend looking at [#37](https://github.com/terraform-providers/terraform-provider-fastly/pull/37). @thommahoney and @phamann recommended to only create a new resource for Entries and keep the container under the service resource. Please share your thoughts on this idea. Please provide also an update on the race condition you noticed recently.

Thank you for putting so much thought into this implementation.

philippschulte:
Thanks for your hard work but we decided to go with #184.

PR #141, opened 2019-04-05
Return full string content for VCL state

### LINKED ISSUE #140: VCL comparison problem after import
After importing a fastly service the VCL content are stored (rightfully so IMHO) in the terraform state.
Yet, it is impossible to compare VCL's with local resources as the output will always be a the SHA sum of the string instead of the full string.

https://github.com/terraform-providers/terraform-provider-fastly/blob/afa36965dc4a92683385f4bd5b8b13f336fec570/fastly/resource_fastly_service_v1.go#L1242

Is that a known issue? we'd like to keep the full VCL comparison possible as it help understanding the potentially destructive changes.

flovilmart:
When comparing vcl after import, only the checksum of the file would be printed. This allows for comparing the full content of the VCL. 

Closes #140.

Another resolution would have been to use the SHA on import but this prevents accurate comparison of the differences upon applying.

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -1239,8 +1239,7 @@ func resourceServiceV1() *schema.Resource {
 							StateFunc: func(v interface{}) string {
 								switch v.(type) {
 								case string:
-									hash := sha1.Sum([]byte(v.(string)))
-									return hex.EncodeToString(hash[:])
+									return v.(string)
 								default:
 									return ""
 								}


chenrui333 (commented):
Besides `vcl` section, `snippets` section also needs to be considered.

Also, do we have test for this change?

flovilmart:
@chenrui333

> Besides vcl section, snippets section also needs to be considered.

agreed, my issue was solely around VCL's as I was not using snippets either.

> Also, do we have test for this change?

no particular test but no regressions. What do you envision?

chenrui333:
yeah, I am using snippets :(, so it would be good to do for both (I do think VCL is more important though).

in terms of tests, basically test the diff process.

philippschulte:
👋@flovilmart, I totally understand your pain about the `plan` output and that it doesn't allow you to compare `VCL` changes like in the following example:

```

  ~ fastly_service_v1.test
      vcl.2780786572.content: "sub vcl_recv {\n#FASTLY recv\n\n  if (req.method != \"HEAD\" && req.method != \"GET\" && req.method != \"FASTLYPURGE\") {\n    return(pass);\n  }\n\n  return(lookup);\n}\n\nsub vcl_fetch {\n#FASTLY fetch\n\n  if ((beresp.status == 500 || beresp.status == 503) && req.restarts < 1 && (req.method == \"GET\" || req.method == \"HEAD\")) {\n    restart;\n  }\n\n  if (req.restarts > 0) {\n    set beresp.http.Fastly-Restarts = req.restarts;\n  }\n\n  if (beresp.http.Set-Cookie) {\n    set req.http.Fastly-Cachetype = \"SETCOOKIE\";\n    return(pass);\n  }\n\n  if (beresp.http.Cache-Control ~ \"private\") {\n    set req.http.Fastly-Cachetype = \"PRIVATE\";\n    return(pass);\n  }\n\n  if (beresp.status == 500 || beresp.status == 503) {\n    set req.http.Fastly-Cachetype = \"ERROR\";\n    set beresp.ttl = 1s;\n    set beresp.grace = 5s;\n    return(deliver);\n  }\n\n  if (beresp.http.Expires || beresp.http.Surrogate-Control ~ \"max-age\" || beresp.http.Cache-Control ~ \"(s-maxage|max-age)\") {\n    # keep the ttl here\n  } else {\n    # apply the default ttl\n    set beresp.ttl = 3600s;\n  }\n\n  return(deliver);\n}\n\nsub vcl_hit {\n#FASTLY hit\n\n  if (!obj.cacheable) {\n    return(pass);\n  }\n  return(deliver);\n}\n\nsub vcl_miss {\n#FASTLY miss\n  return(fetch);\n}\n\nsub vcl_deliver {\n#FASTLY deliver\n  return(deliver);\n}\n\nsub vcl_error {\n#FASTLY error\n}\n\nsub vcl_pass {\n#FASTLY pass\n}\n\nsub vcl_log {\n#FASTLY log\n}\n" => ""
      vcl.2780786572.main:    "true" => "false"
      vcl.2780786572.name:    "my_custom_main_vcl" => ""
      vcl.48442639.content:   "" => "aeae5855ee1551ceee4655ead95faea05918a4b4"
      vcl.48442639.main:      "" => "true"
      vcl.48442639.name:      "" => "my_custom_main_vcl"


``` 

However, I would like to close this PR out because even if we populate the `VCL` content for the new and the old resource it's difficult to compare changes in the Shell. Therefore, we have recently added the new [activate](https://www.terraform.io/docs/providers/fastly/r/service_v1.html#activate) attribute to the provider which allows you to use Fastly's [Diff Tool](https://docs.fastly.com/guides/basic-setup/working-with-services#comparing-different-service-versions) before the new version goes live.

If you set `activate` to `false` it will prevent Terraform from activating the new version. This allows you to compare the version in the UI which is much nicer than to compare any changes in the Shell. Another reason why I don't like your changes is that most users are running Terraform in automation. Most CI tools don't like large outputs (truncated log output). Also, the log length is limited! It wouldn't be a big deal for the above example but usually, users who use Terraform also use custom VCL and there can be easily thousands of lines of code for programming HTTP requests.

Please let me know if the `activate` attribute solves your issue and if you're ok if I close this one out. Thanks for your contribution!

flovilmart:
hi @philippschulte, I completely understand your point. Thanks!

PR #145, opened 2019-04-16
Add support for splunk logging

### LINKED ISSUE #130: Feature Request: Add Splunk logging argument
Configuring a Splunk logging enpoint is available in the web GUI, but there's no argument available in the terraform resource.

philippschulte:
Closes #130.

Acceptance tests:

```
=== RUN   TestResourceFastlyFlattenSplunk
--- PASS: TestResourceFastlyFlattenSplunk (0.00s)
=== RUN   TestAccFastlyServiceV1_splunk_basic
--- PASS: TestAccFastlyServiceV1_splunk_basic (90.82s)
=== RUN   TestAccFastlyServiceV1_splunk_default
--- PASS: TestAccFastlyServiceV1_splunk_default (46.54s)
=== RUN   TestAccFastlyServiceV1_splunk_env
--- PASS: TestAccFastlyServiceV1_splunk_env (47.41s)
```

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -1069,6 +1069,67 @@ func resourceServiceV1() *schema.Resource {
 				},
 			},
 
+			"splunk": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						// Required fields
+						"name": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "The unique name of the Splunk logging endpoint",
+						},
+						"url": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "The Splunk URL to stream logs to",
+						},
+						"token": {
+							Type:        schema.TypeString,
+							Required:    true,
+							DefaultFunc: schema.EnvDefaultFunc("FASTLY_SPLUNK_TOKEN", ""),
+							Description: "The Splunk token to be used for authentication",
+							Sensitive:   true,
+							StateFunc: func(v interface{}) string {
+								switch v.(type) {
+								case string:
+									hash := sha512.Sum512([]byte(v.(string)))
+									return hex.EncodeToString(hash[:])
+								default:
+									return ""
+								}
+							},
+						},
+						// Optional fields
+						"format": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Default:     "%h %l %u %t \"%r\" %>s %b",
+							Description: "Apache-style string or VCL variables to use for log formatting (default: `%h %l %u %t \"%r\" %>s %b`)",
+						},
+						"format_version": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							Default:      2,
+							Description:  "The version of the custom logging format used for the configured endpoint. Can be either 1 or 2. (default: 2)",
+							ValidateFunc: validateLoggingFormatVersion(),
+						},
+						"placement": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Description:  "Where in the generated VCL the logging call should be placed",
+							ValidateFunc: validateLoggingPlacement(),
+						},
+						"response_condition": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Description: "The name of the condition to apply",
+						},
+					},
+				},
+			},
+
 			"response_object": {
 				Type:     schema.TypeSet,
 				Optional: true,
@@ -1329,6 +1390,7 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 		"syslog",
 		"sumologic",
 		"logentries",
+		"splunk",
 		"response_object",
 		"condition",
 		"request_setting",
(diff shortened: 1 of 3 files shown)


fastly/resource_fastly_service_v1_splunk_test.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+}
+
+// struct to preserve the current environment
+type currentSplunkEnv struct {
+	Token string
+}
+
+func getSplunkEnv() *currentSplunkEnv {
+	// Grab the existing Fastly Azure SAS token and preserve, in the off chance
```
trentrosenbaum: This needs to be changed and not to mentioned Azure, I thought I was experiencing Déjà vu :)

fastly/resource_fastly_service_v1_splunk_test.go
```diff
@@ -0,0 +1,404 @@
...
+
+	for _, c := range cases {
+		out := flattenSplunks(c.remote)
+		if !reflect.DeepEqual(out, c.local) {
+			t.Fatalf("Error matching:\nexpected: %#v\ngot: %#v", c.local, out)
+		}
+	}
+}
+
+func TestAccFastlyServiceV1_splunk_complete(t *testing.T) {
```
trentrosenbaum: For me, it is a bit hard to understand what the acceptance tests are covering.  If I read this then I understand that the servicev1 resource is being tested with the Splunk nest block, but when I read 'complete' I am not sure about that.  The other acceptance tests end with default and env.

fastly/resource_fastly_service_v1_splunk_test.go
```diff
@@ -0,0 +1,404 @@
...
+		},
+	})
+}
+
+func TestAccFastlyServiceV1_splunk_env(t *testing.T) {
+	var service gofastly.ServiceDetail
+	serviceName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+
+	// set env variable to something we expect
+	resetEnv := setSplunkEnv("test-token", t)
```
trentrosenbaum, line 157: I am a bit confused about this line.  Why do we only set this here?  I can see in the other tests that the value 'test-token' is being used by it is only set and unset here.

philippschulte: We only invoke `setSplunkEnv` once because we have only one test to make sure that the provider applies the `FASTLY_SPLUNK_TOKEN` environment variable. At the end of the test, we delete the `FASTLY_SPLUNK_TOKEN` test variable and reset it with the existing value. This test suite uses the exact same pattern to test the `schema.EnvDefaultFunc` functionality as all other suites of this provider.

fastly/resource_fastly_service_v1_splunk_test.go
philippschulte: This particular test applies all possible fields of the API. Therefore, I didn't want to name it `basic`. One of the biggest problems, in my opinion, is that the other suites have no tests which cover all possible fields. I'll rename it to `basic`.

philippschulte:
@trentrosenbaum please review the PR again.

fastly/resource_fastly_service_v1.go
```diff
@@ -1069,6 +1069,67 @@ func resourceServiceV1() *schema.Resource {
...
+							Optional:    true,
+							Default:     "%h %l %u %t \"%r\" %>s %b",
+							Description: "Apache-style string or VCL variables to use for log formatting (default: `%h %l %u %t \"%r\" %>s %b`)",
+						},
+						"format_version": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							Default:      2,
+							Description:  "The version of the custom logging format used for the configured endpoint. Can be either 1 or 2. (default: 2)",
+							ValidateFunc: validateLoggingFormatVersion,
```
trentrosenbaum: The following line is causing the build to fail.  Not sure if there was a fix that has not been pushed to the PR, but we need to add "()"

The following error is returned by travis

```cannot use validateLoggingFormatVersion (type func() schema.SchemaValidateFunc) as type schema.SchemaValidateFunc in field value```

File: fastly/resource_fastly_service_v1.go [hunk 5]
```diff
@@ -1069,6 +1069,67 @@ func resourceServiceV1() *schema.Resource {
...
+						},
+						"placement": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Description:  "Where in the generated VCL the logging call should be placed",
+							ValidateFunc: validateLoggingPlacement,
```
trentrosenbaum: The following line is causing the build to fail. Not sure if there was a fix that has not been pushed to the PR, but we need to add "()"

The following error is returned by travis

```cannot use validateLoggingPlacement (type func() schema.SchemaValidateFunc) as type schema.SchemaValidateFunc in field value```

File: fastly/resource_fastly_service_v1.go [hunk 6]
```diff
@@ -1069,6 +1069,67 @@ func resourceServiceV1() *schema.Resource {
...
+							Required:    true,
+							Description: "The Splunk URL to stream logs to",
+						},
+						"token": {
+							Type:        schema.TypeString,
+							Required:    true,
+							DefaultFunc: schema.EnvDefaultFunc("FASTLY_SPLUNK_TOKEN", ""),
+							Description: "The Splunk token to be used for authentication",
+							Sensitive:   true,
+							StateFunc: func(v interface{}) string {
```
trentrosenbaum, line 1094: For the statefunc we will need to remove this when we merge in the 0.12.x changes/PR later.  For now other resources have these in the 0.11.x release of the provider.

trentrosenbaum (changes requested):
I have made the changes suggested locally and executed the tests.  They all pass.  We just need to make them to this PR and confirm Travis builds.

trentrosenbaum (approved):

PR #149, opened 2019-05-06
Add validation for condition type attribute

### LINKED ISSUE #148: Condition types are now uppercase
### Terraform Version
[0.11.7 Linux AMD 64](https://releases.hashicorp.com/terraform/0.11.7/terraform_0.11.7_linux_amd64.zip)

### Affected Resource(s)
- condition

Fastly provider version 0.6.0
AWS provider version 2.8.0

### Actual Behavior

`condition` blocks used to accept a lowercase type (`request`, `response`, etc). Now it appears they require uppercase types (`REQUEST`, `RESPONSE`).

The lowercase types stopped working sometime after April 15th. We never bumped the major version of Terraform or the Fastly provider so did not expect a breaking change.

### Expected Behavior

I would expect the lowercase types to continue working when using the same Terraform and provider versions.

philippschulte:
Closes #148.

Since conditions are supported the type must be either `REQUEST`, `RESPONSE`, or `CACHE`. The `go-fastly` HTTP client and the Fastly API allow lower-case values for the type attribute! However, the Fastly API always returns an upper-case value for the type attribute. Therefore, a `plan` output always detects changes for a condition if the type attribute in the Terraform definition is lower-case because it compares a lower-case with an upper-case value! This validation prevents the Terraform definition from being different from what the Fastly API returns. **This is not a breaking change!**

Unit tests:

```
go test -i $(go list ./... |grep -v 'vendor') || exit 1
echo $(go list ./... |grep -v 'vendor') | \
		xargs -t -n4 go test  -timeout=30s -parallel=4
go test -timeout=30s -parallel=4 github.com/terraform-providers/terraform-provider-fastly github.com/terraform-providers/terraform-provider-fastly/fastly
ok  	github.com/terraform-providers/terraform-provider-fastly/fastly	0.020s
```

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -94,9 +94,10 @@ func resourceServiceV1() *schema.Resource {
 							Description: "A number used to determine the order in which multiple conditions execute. Lower numbers execute first",
 						},
 						"type": {
-							Type:        schema.TypeString,
-							Required:    true,
-							Description: "Type of the condition, either `REQUEST`, `RESPONSE`, or `CACHE`",
+							Type:         schema.TypeString,
+							Required:     true,
+							Description:  "Type of the condition, either `REQUEST`, `RESPONSE`, or `CACHE`",
+							ValidateFunc: validateConditionType(),
 						},
 					},
 				},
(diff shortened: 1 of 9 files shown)


phamann (approved):
LGTM :+1:

PR #150, opened 2019-05-17
Outsource and refactor validators

philippschulte:
👋 This PR moves all remaining validators out of `resource_fastly_service_v1.go` into `validators.go`.

Unit tests:
```
go test -i $(go list ./... |grep -v 'vendor') || exit 1
echo $(go list ./... |grep -v 'vendor') | \
		xargs -t -n4 go test  -timeout=30s -parallel=4
go test -timeout=30s -parallel=4 github.com/terraform-providers/terraform-provider-fastly github.com/terraform-providers/terraform-provider-fastly/fastly
ok  	github.com/terraform-providers/terraform-provider-fastly/fastly	0.021s
```

### DIFF
--- a/fastly/validators.go
+++ b/fastly/validators.go
@@ -154,3 +154,19 @@ func validateHeaderAction(v interface{}, k string) (ws []string, errors []error)
 	}
 	return
 }
+
+func validateHeaderType(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+	validTypes := map[string]struct{}{
+		"request":  {},
+		"fetch":    {},
+		"cache":    {},
+		"response": {},
+	}
+
+	if _, ok := validTypes[value]; !ok {
+		errors = append(errors, fmt.Errorf(
+			"%q must be one of ['request', 'fetch', 'cache', 'response']", k))
+	}
+	return
+}
--- a/fastly/validators.go
+++ b/fastly/validators.go
@@ -137,3 +137,20 @@ func validateSnippetType(v interface{}, k string) (ws []string, errors []error) 
 	}
 	return
 }
+
+func validateHeaderAction(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+	validActions := map[string]struct{}{
+		"set":          {},
+		"append":       {},
+		"delete":       {},
+		"regex":        {},
+		"regex_repeat": {},
+	}
+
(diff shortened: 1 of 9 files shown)


fastly/validators_test.go
```diff
@@ -78,30 +78,170 @@ func TestValidateLoggingPlacement(t *testing.T) {
 	}
 }
 
+func TestValidateDirectorQuorum(t *testing.T) {
```
trentrosenbaum, line 81: Just to follow up on the conversations.  Suggestion to turn these into table driven tests.  Allowing increased feedback on all test scenarios instead of stopping at the failing point in the test data array.  Placing comment here, but same suggestion for all the of the test functions in the validator_test.go file.

fastly/validators.go
```diff
@@ -34,29 +34,108 @@ func validateLoggingMessageType(v interface{}, k string) (ws []string, errors []
...
+	if _, ok := validTypes[value]; !ok {
 		errors = append(errors, fmt.Errorf(
 			"%q must be one of ['1' (random), '3' (hash), '4' (client)]", k))
 	}
 	return
 }
+
+func validateConditionType(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+	validTypes := map[string]struct{}{
```
trentrosenbaum, line 121: A map kind of feels a bit out of place here and in the other validation functions.  Could a string array be used with a contains function.  Values are not being added dynamically to the map.

fastly/validators.go
```diff
@@ -34,29 +34,108 @@ func validateLoggingMessageType(v interface{}, k string) (ws []string, errors []
...
+		"REQUEST":  {},
+		"RESPONSE": {},
+		"CACHE":    {},
+	}
+
+	if _, ok := validTypes[value]; !ok {
+		errors = append(errors, fmt.Errorf(
+			"%q must be one of ['REQUEST', 'RESPONSE', 'CACHE']", k))
```
trentrosenbaum: Could this message be built using the values in the previous data structure, (or string array, see last comment)?  Just a way to keep them in sync.

trentrosenbaum (commented):
I have just added some comments to the PR covering the table driven test suggestion and a minor thought about the use of map in the validator functions.

philippschulte:
The approach of how we want to refactor the validators has changed. Therefore, it's easier to just close this one and open another one.

PR #158, opened 2019-05-29
Introduce and rename flatten function unit tests.

trentrosenbaum:
This PR introduces missing flatten function unit test for the following nested blocks within the service
cache_setting, conditionals, director, gzip, headers, healthcheck, logentries, papertrail, request_setting, response_object, snippet, syslog, vcl, S3logging.

The following make target and regx can be used to execute the unit tests.
```sh
make testacc TESTARGS="-run=TestResourceFastlyFlatten.*"
```

The test cases that featured inside the test function has been replaced with table tests.  This will allow the data to be increased without the function becoming too noisy.

### DIFF
--- a/fastly/provider_test.go
+++ b/fastly/provider_test.go
@@ -1,6 +1,10 @@
 package fastly
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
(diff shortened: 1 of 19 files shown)


matpimenta (approved):
LGTM

PR #159, opened 2019-05-31
Adding terraform provider version to user-agent header

matpimenta:


GNUmakefile
```diff
@@ -2,19 +2,20 @@ TEST?=$$(go list ./... |grep -v 'vendor')
 GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=fastly
+VERSION=$(shell git describe --tags --always)
 
 default: build
 
 build: fmtcheck
-	go install
+	go install -ldflags="-X github.com/terraform-providers/terraform-provider-fastly/version.ProviderVersion=$(VERSION)"
```
phamann: I suggest moving the repo string `github.com/terraform-providers/terraform-provider-fastly/` to a variable at the top of the file to use in both of these lines.

phamann (approved):
:+1: LGTM other than minor nit.

GNUmakefile
matpimenta: good point, let me fix that

trentrosenbaum (approved):

fastly/config.go
```diff
@@ -24,7 +27,9 @@ func (c *Config) Client() (interface{}, error) {
 		return nil, fmt.Errorf("[Err] No API key for Fastly")
 	}
 
-	gofastly.UserAgent = terraform.UserAgentString()
+	composedUserAgent := fmt.Sprintf("%s %s/%s", terraform.UserAgentString(), TerraformProviderProductUserAgent, version.ProviderVersion)
```
philippschulte: `terraform.UserAgentString()` is deprecated!

philippschulte (changes requested):
I have looked at the big public cloud providers how they set the `User-Agent` header and noticed that `terraform.UserAgentString()` is deprecated! I would prefer the same pattern used by the big providers:

```
tfUserAgent := httpclient.UserAgentString()
pv := version.ProviderVersion
providerUserAgent := fmt.Sprintf("%s terraform-provider-fastly/%s", tfUserAgent, pv)
gofastly.UserAgent = providerUserAgent
```

matpimenta:
I've removed the deprecation and changed user agent to `terraform-provider-fastly`.

philippschulte (approved):

PR #165, opened 2019-06-10
Add support for override host field

### LINKED ISSUE #163: Can't set override_host on an individual backend
Hi there,

We have a multi backend site that uses a mix of S3 buckets for static pages and EC2 instances, and need to have an EC2 instance set to our default host (i.e. when a page 404s, we need to be sure this is handled by one of our servers), so we cannot follow the [Fastly guide here](https://docs.fastly.com/guides/basic-configuration/specifying-an-override-host) for configuring an s3 bucket as this configures the override_host at a global level. We need the ability to set the override_host flag for s3 on a backend through terraform.

Currently we have to implement a local-exec workaround that makes a CURL call directly to the API and this is not reflected in our terraform plan/apply diff:

```
 provisioner "local-exec" {
    command = "curl -X PUT 
    -H \"Fastly-Key: ${var.fastly_api}\" 
    -H \"Content-Type: application/x-www-form-urlencoded\" 
    -H \"Accept: application/json\" 
    -d \"override_host=s3-bucket-url-goes-here\" 
    \"https://api.fastly.com/service/${self.id}/version/${self.active_version + 1}/backend/Backend-Name-In-Fastly\""
  }
```

### Terraform Version
```
Terraform v0.11.14
+ provider.fastly v0.6.1
```

philippschulte:
🎉this closes #163.

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -2884,7 +2884,7 @@ func resourceServiceV1Read(d *schema.ResourceData, meta interface{}) error {
 
 		// refresh BigQuery Logging
 		log.Printf("[DEBUG] Refreshing BigQuery for (%s)", d.Id())
-		BQList, err := conn.GetBigQuery(&gofastly.GetBigQueryInput{
+		BQList, err := conn.ListBigQueries(&gofastly.ListBigQueriesInput{
 			Service: d.Id(),
 			Version: s.ActiveVersion.Number,
 		})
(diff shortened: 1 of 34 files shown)


trentrosenbaum (changes requested):
From the point of view of introducing a new field, this looks ok and the field is documented in the provider and on the fastly api website.  I cannot see any documented validation and so this should be fine.

The current build is failing and so just marking it as "Request Changes" to see a green build.

philippschulte:
@trentrosenbaum the build wasn't passing because the client hasn't been updated. The PR updates the `go-fastly` client from v0.4.3 to v1.0.0 in order to support the new `override_host` field. By doing so some fixes were necessary because v1.0.0 includes some breaking changes. Please review it again.

trentrosenbaum (approved):

PR #182, opened 2019-08-01
provider: Use real Terraform version in UA header

radeksimko:
Depends on https://github.com/terraform-providers/terraform-provider-fastly/pull/187

---

The existing implementation leverages Terraform's `version` package as Terraform currently provides the SDK for interacting with core and is therefore imported by the provider, including the `version` package.

The side effect of this is that provider reports version of vendored Terraform (i.e. plugin SDK), not the version which actually talks to the provider's gRPC server and parses the configuration.

This PR is leveraging recent changes introduced in 0.12 to fix that and report accurate version.

----

This also changes the header from the following

```
Terraform/X.Y.Z terraform-provider-fastly/X.Y.Z
```
to
```
HashiCorp Terraform/X.Y.Z (+https://www.terraform.io) terraform-provider-fastly/X.Y.Z
```

to align UA string with https://github.com/hashicorp/terraform/pull/22272


philippschulte (approved):
Thanks, @radeksimko for upgrading Terraform and for fixing deprecation issues!

PR #185, opened 2019-08-07
Refactor CHANGELOG for consistency

philippschulte:
The CHANGELOG was not consistent for all releases!


phamann (approved):
The term `New Nested Resource` sounds strange to me, but I can't decide why. They're not "resources", they're just `arguments` or `attributes` of the service resource. However, I can't think of anything better :)

If we keep as is, can we use sentence case, i.e. change to  "New nested resource"

philippschulte:
@phamann you're right about the wording! I used it because we often talk about nested resources.  I updated it to `arguments`. The `FEATURES` flag should only be used for new resources!

PR #184, opened 2019-08-05
Introduction of ACL, Dictionary and Dynamic Snippet versionless resources

trentrosenbaum:
The following PR represents the introduction of new resources into the provider.  These resources have two parts them.  The first is a new nested block within the main service resources.  The second is a separate resource representing the versionless content to be linked to the nested block.

ACL and ACL Entries resources represent the Fastly ACL and ACL Entries respectively.
* https://docs.fastly.com/api/config#api-section-acl
* https://docs.fastly.com/api/config#api-section-acl_entry

Dictionary and Dictionary Items resources represent the Fastly Dictionary and Dictionary Items
* https://docs.fastly.com/api/config#api-section-dictionary
* https://docs.fastly.com/api/config#api-section-dictionary_item

Dynamic Snippet and Dynamic Snippet Content represent the Dynamic aspect of a Fastly Snippet.  This resource is very similar to a snippet, but the content is managed through the new versionless resource.
* https://docs.fastly.com/api/config#api-section-snippet

Each of the resources introduces updates to the service and new resources.  Along side these updates are new Acceptance tests for each resource type and documentation.

Each of the resources have been developed through separate branches.  The following PRs were raised and have now been closed to cover the individual pieces of work.  For reference they are listed.
* https://github.com/terraform-providers/terraform-provider-fastly/pull/178
* https://github.com/terraform-providers/terraform-provider-fastly/pull/180
* https://github.com/terraform-providers/terraform-provider-fastly/pull/181

PR #186, opened 2019-08-07
Fix documentation sidebar navigation links and markdown rendering

phamann:
### TL;DR
Fixes the sidebar navigation links for the documentation website and some rendering of the markdown.

### What:
- Update all new resource markdown to include correct `sidebar_current` metadata value
- Fix link for `fastly_service_v1` which was accidentally hidden in #184 
- Add links to new resources introduce in #184
- Fix rendering of `fastly_service_dynamic_snippet_content_v1` markdown

### DIFF
--- a/website/docs/r/service_dynamic_snippet_content_v1.html.markdown
+++ b/website/docs/r/service_dynamic_snippet_content_v1.html.markdown
@@ -4,117 +4,116 @@ page_title: "Fastly: service_dynamic_snippet_content_v1"
 sidebar_current: "docs-fastly-resource-service-dynamic-snippet-content-v1"
 description: |-
   Provides a means to define blocks of VCL logic that is inserted into your service through Fastly dynamic snippets.
-   
 ---
 
- # fastly_service_dynamic_snippet_content_v1
+# fastly_service_dynamic_snippet_content_v1
 
- Defines content that represents blocks of VCL logic that is inserted into your service.  This resource will populate the content of a dynamic snippet and allow it to be manged without the creation of a new service verison. 
+Defines content that represents blocks of VCL logic that is inserted into your service.  This resource will populate the content of a dynamic snippet and allow it to be manged without the creation of a new service verison. 
  
- ~> **Warning:** Terraform will take precedence over any changes you make through the API. Such changes are likely to be reversed if you run Terraform again.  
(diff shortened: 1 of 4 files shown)


trentrosenbaum (approved):

trentrosenbaum (changes requested):
Currently looking at the header to the dynamic snippet page.  It looks to be rendering incorrectly.

website/docs/r/service_dynamic_snippet_content_v1.html.markdown
```diff
@@ -128,34 +127,34 @@ resource "fastly_service_dynamic_snippet_content_v1" "my_dyn_content" {
...
 The resource ID is a combined value of the `service_id` and `snippet_id` separated by a forward slash.
 
- ```
+```
 $ terraform import fastly_service_dynamic_snippet_content_v1.content xxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxx
 ```
 
-If Terraform is already managing remote content against a resource being imported then the user will be asked to remove it from the existing Terraform state.  
+If Terraform is already managing remote content against a resource being imported then the user will be asked to remove it from the existing Terraform state.
 The following is an example of the Terraform state command to remove the resource named fastly_service_dynamic_snippet_content_v1.content` from the Terraform state file.
```
trentrosenbaum: The following value is not rendering like the rest of the references on the page.  The value is missing a start backtick, but it does have the closing backtick.

fastly_service_dynamic_snippet_content_v1.content

trentrosenbaum (changes requested):

website/docs/r/service_dynamic_snippet_content_v1.html.markdown
phamann: 👍 Fixed

trentrosenbaum (approved):
Looked over the branch in Goland so I could see the render of markdown.  The dynamic snippet page looks correct with the new spacing, (covering the heading) and the inline formatting, (covering the backtick).  The sidebar values match the Fastly anchor markup and the references in the anchors were checked to make sure they link to real pages in the live site.

This looks to be rendering correctly and so approving.

philippschulte (approved):
+1

PR #189, opened 2019-09-27
add support for write-only dictionaries

sbfaulkner:
This PR updates the service resource so that private (write-only) dictionaries are supported.
(see https://docs.fastly.com/en/guides/private-dictionaries)

By default dictionaries are not write-only (ie. `write_only` == `false`).

Changing the write-only attribute is not supported, so the provider will recreate the dictionary if the value of the write_only attribute is changed.

#### acceptance tests
```
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccFastlyServiceV1_dictionary.* -timeout 120m -ldflags="-X=github.com/terraform-providers/terraform-provider-fastly/version.ProviderVersion=acc"
=== RUN   TestAccFastlyServiceV1_dictionary
--- PASS: TestAccFastlyServiceV1_dictionary (41.78s)
=== RUN   TestAccFastlyServiceV1_dictionary_write_only
--- PASS: TestAccFastlyServiceV1_dictionary_write_only (44.44s)
=== RUN   TestAccFastlyServiceV1_dictionary_update_name
--- PASS: TestAccFastlyServiceV1_dictionary_update_name (75.85s)
PASS
ok  	github.com/terraform-providers/terraform-provider-fastly/fastly	162.097s
?   	github.com/terraform-providers/terraform-provider-fastly/version	[no test files]
```

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -1452,6 +1452,13 @@ func resourceServiceV1() *schema.Resource {
 							Computed:    true,
 							Description: "Generated dictionary ID",
 						},
+						"write_only": {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Default:     false,
+							ForceNew:    true,
+							Description: "Determines if items in the dictionary are readable or not",
+						},
 					},
 				},
 			},
@@ -4563,7 +4570,8 @@ func flattenDynamicSnippets(dynamicSnippetList []*gofastly.Snippet) []map[string
 func buildDictionary(dictMap interface{}) (*gofastly.CreateDictionaryInput, error) {
 	df := dictMap.(map[string]interface{})
 	opts := gofastly.CreateDictionaryInput{
-		Name: df["name"].(string),
+		Name:      df["name"].(string),
+		WriteOnly: gofastly.CBool(df["write_only"].(bool)),
 	}
 
 	return &opts, nil
@@ -4576,6 +4584,7 @@ func flattenDictionaries(dictList []*gofastly.Dictionary) []map[string]interface
 		dictMapString := map[string]interface{}{
 			"dictionary_id": currentDict.ID,
 			"name":          currentDict.Name,
+			"write_only":    currentDict.WriteOnly,
 		}
 
 		// prune any empty values that come from the default string value in structs
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -1456,7 +1456,6 @@ func resourceServiceV1() *schema.Resource {
 							Type:        schema.TypeBool,
 							Optional:    true,
 							Default:     false,
-							ForceNew:    true,
 							Description: "Determines if items in the dictionary are readable or not",
 						},
 					},
--- a/fastly/resource_fastly_service_v1_dictionary_test.go
+++ b/fastly/resource_fastly_service_v1_dictionary_test.go
@@ -19,14 +19,16 @@ func TestResourceFastlyFlattenDictionary(t *testing.T) {
 		{
 			remote: []*gofastly.Dictionary{
 				{
-					ID:   "1234567890",
-					Name: "dictionary-example",
+					ID:        "1234567890",
+					Name:      "dictionary-example",
+					WriteOnly: false,
 				},
 			},
 			local: []map[string]interface{}{
 				{
 					"dictionary_id": "1234567890",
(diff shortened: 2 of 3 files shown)


sbfaulkner:
not sure who to ping, or what the process for this provider/repo is so looked at recent contributors
@philippschulte

philippschulte:
@sbfaulkner thank you for opening a PR to support `write-only` dictionaries! We recently added support for dictionaries and if I remember correctly there was a reason why we didn't support it in the first place. I will discuss this PR with my team and get back to you shortly.

fastly/resource_fastly_service_v1.go
```diff
@@ -1452,6 +1452,13 @@ func resourceServiceV1() *schema.Resource {
 							Computed:    true,
 							Description: "Generated dictionary ID",
 						},
+						"write_only": {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Default:     false,
+							ForceNew:    true,
```
philippschulte: `ForceNew` indicates that any change in this field requires the resource to be destroyed and recreated. It makes totally sense to me to apply it here but it doesn't work because of the implementation of the Fastly provider. The reason is that a dictionary is not a standalone resource! Changing this field results to:

```

  # fastly_service_v1.test must be replaced
-/+ resource "fastly_service_v1" "test" {
        activate       = true
      ~ active_version = 1 -> (known after apply)
        comment        = "Managed by Terraform"
      + default_host   = (known after apply)
        default_ttl    = 3600
      + force_destroy  = true
      ~ id             = "609pjtHtRLoZkvK5IuGnI1" -> (known after apply)
        name           = "demofastly"

        backend {
            address               = "1.2.3.4"
            auto_loadbalance      = true
            between_bytes_timeout = 10000
            connect_timeout       = 1000
            error_threshold       = 0
            first_byte_timeout    = 15000
            max_conn              = 200
            name                  = "localhost"
            port                  = 80
            ssl_check_cert        = true
            use_ssl               = false
            weight                = 100
        }

      - dictionary { # forces replacement
          - dictionary_id = "1gaW92CKUj3mYKn6iY9bvF" -> null
          - name          = "my_dictionary" -> null
          - write_only    = true -> null
        }
      + dictionary { # forces replacement
          + dictionary_id = (known after apply)
          + name          = "my_dictionary"
          + write_only    = false
        }

        domain {
            name = "www.test.com"
        }
    }

Plan: 1 to add, 0 to change, 1 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

fastly_service_v1.myservice: Destroying... [id=609pjtHtRLoZkvK5IuGnI1]

Error: 400 - Bad Request:

    Title:  Bad request
    Detail: Cannot delete service with active version
```

philippschulte:
@sbfaulkner I did review the PR and the code looks good besides the `ForceNew` field. It just needs to be deleted. The `write_only` field also needs to be documented [here](https://www.terraform.io/docs/providers/fastly/r/service_v1.html#name-24). It is very important to note that the `write_only` field should only be used if items are managed outside of Terraform! The documentation changes need to be made in `terraform-provider-fastly/website/docs/r/service_v1.html.markdown`.

fastly/resource_fastly_service_v1.go
sbfaulkner: bummer... I'd hoped that it was just the sub-resource that would be destroyed

sbfaulkner:
I've removed `ForceNew`...
and have confirmed it works as expected...

```
      - dictionary {
          - dictionary_id = "6PooubYTc15fgQXreURx22" -> null
          - name          = "test" -> null
          - write_only    = false -> null
        }
      + dictionary {
          + dictionary_id = (known after apply)
          + name          = "test"
          + write_only    = true
        }
```

I've also added an acceptance test for that case.

sbfaulkner:
@philippschulte documentation has now been added in addition to the code (and test) changes

philippschulte:
@sbfaulkner the changes look great! Thanks again for all the work. I guess we won't be able to release it this week but next week for sure!

sbfaulkner:
@philippschulte any chance this will be rolled out soon?

philippschulte:
@sbfaulkner [v0.11.0](https://github.com/terraform-providers/terraform-provider-fastly/blob/master/CHANGELOG.md#0110-october-15-2019) has been released today!

PR #194, opened 2019-10-28
Fix document

skaji:
This PR fixes:

* 72f4526: Header -> header
* 40ce705: I think `name - (Required) The domain for this request setting.` is incorrect. It's just an identifier in request_settings.


philippschulte (approved):
@skaji thank you for your contribution!

PR #196, opened 2019-10-30
Sensitive vcl snippet

### LINKED ISSUE #195: Feature Request: support sensitive VCL snippet which is hidden in plan result
Fastly provides [a guide](https://docs.fastly.com/en/guides/amazon-s3) that injects secret keys into VCL snippet to use Amazon S3 private bucket as origin.
When managing this kind of VCL snippet by terraform, I wouldn't like at least `awsSecretKey` in the guide to be visible in `terraform plan` result.
Therefore, I'd like `snippet` block in `fastly_service_v1` to support an argument like [`sensitive_content` in `local_file` resource](https://www.terraform.io/docs/providers/local/r/file.html#sensitive_content).

### Terraform Version
```
$ terraform -v
Terraform v0.12.8
```

### Affected Resource(s)
- fastly_service_v1

```hcl
resource "fastly_service_v1" "myservice" {
  name = "myawesometestservice"

  snippet {
    name = "myawesomesnippet"
    type = "recv"
    sensitive_content = "some sensitive vcl snippet"
  }
}
```

### Expected Behavior
```
$ terraform plan
...
      + snippet {
          + sensitive_content    = (sensitive value)
          + name     = "myawesomesnippet"
          + priority = 100
          + type     = "recv"
        }
...
```

cookie-s:
Resolves #195 .
See the above issue for the detail.

I wanted to use `ConflictsWith` or `ValidateFunc` in `schema.Schema`, but both don't work for this case (arguments for resource with `TypeSet` type) now.
So instead, this PR validates the exclusiveness during apply.

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -1384,8 +1384,14 @@ func resourceServiceV1() *schema.Resource {
 						},
 						"content": {
 							Type:        schema.TypeString,
-							Required:    true,
+							Optional:    true,
 							Description: "The contents of the VCL snippet",
+						},
+						"sensitive_content": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Sensitive:   true,
+							Description: "The contents of the VCL snippet (not displayed in diffs)",
 						},
 						"priority": {
 							Type:        schema.TypeInt,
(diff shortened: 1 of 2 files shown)


cookie-s:
@philippschulte 
I'm not sure but I think you are the right person to ask this review...?
What do you think about the issue #195 as well...?
Thank you.

philippschulte:
@cookie-s I am sorry for the delay in getting back to you! Unfortunately, I'll go ahead and close this one out because of the following reasons:

a) The Fastly API doesn't have a field for snippets called `sensitive_content`
b) The `content` field is required by the Fastly API and if you set it to `optional` then `$ terraform validate` won't capture it. Therefore, `$ terraform apply` will do a lot of things but in the end, it will fail if the `content` and the `sensitive_content` field hasn't been set.

The solution for #195 is to use a [write-only dictionary](https://www.terraform.io/docs/providers/fastly/r/service_v1.html#write_only) for your secrets which you can then reference in your snippet.

cookie-s:
Thank you for reply, I'll try that !

PR #201, opened 2019-12-12
Use gofastly library in ip ranges data source

0x91:
This PR changes the IP addresses data source to consume from the gofastly library instead of calling the endpoint directly.

This change is needed because the current implementation does not handle HTTP Errors received from the Fastly API. This caused a production outage for us a week ago when the Fastly API returned 503, and the data source did not return an error.

This change brings this part of the provider in line with the rest of the codebase, which already uses the gofastly library.

### DIFF
--- a/fastly/data_source_ip_ranges.go
+++ b/fastly/data_source_ip_ranges.go
@@ -1,14 +1,10 @@
 package fastly
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"sort"
-	"strconv"
 
-	"github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/hashcode"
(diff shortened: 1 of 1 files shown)


rymir (approved):

phamann (approved):
Thanks for submitting this @0x91, sorry to hear the provider caused an outage. 

LGTM :+1: - we'll get this merged and released and will let you know once done.

philippschulte (approved):
@0x91 thanks for your contribution! I hope that Hashicorp will cut a new release on Monday. We'll keep you updated as @phamann already mentioned.

philippschulte:
@0x91, @rymir just wanted to let you know that Hashicorp cut a new release of the Fastly provider today: https://github.com/terraform-providers/terraform-provider-fastly/blob/master/CHANGELOG.md#0111-december-16-2019. Thanks again for your contribution!

PR #205, opened 2020-01-17
Fix setting default_ttl to 0 on create

spiralman:
This change allows a service to be created with a  `default_ttl` of `0`, fixing #204.

The problem was that, for integer values, the "old" value in the `ResourceData` is `0`, meaning that, if the "new" value is `0`, `HasChange(key)` returns `false`.

It didn't seem possible to override this behavior, either by modifying the "old" value in the `ResourceData` or by specifying a `CustomizeDiff` function (which appears to only be able to *remove* differences, not add ones that wouldn't, otherwise, be detected). Note: if there *is* a way to accomplish this, and that is the preferred method to fix these issues, let us know and we can update the PR.

Instead, we have added an explicit check for the `default_ttl` being set to `0` on service creation and we force the update call to be made, in that case.

We aren't very familiar with the tools available inside the Terraform Plugin SDK, so we may have missed a better solution. Please let us know if there's a better way to fix this issue.

### DIFF
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -1621,6 +1621,11 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 		}
 
 		// update general settings
+
+		// If the requested default_ttl is 0, and this is the first
+		// version being created, HasChange will return false, but we need
+		// to set it anyway, so ensure we update the settings in that
+		// case.
 		if d.HasChange("default_host") || d.HasChange("default_ttl") || (d.Get("default_ttl") == 0 && initialVersion) {
 			opts := gofastly.UpdateSettingsInput{
 				Service: d.Id(),
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -1574,9 +1574,12 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	initialVersion := false
+
 	if needsChange {
 		latestVersion := d.Get("active_version").(int)
 		if latestVersion == 0 {
+			initialVersion = true
 			// If the service was just created, there is an empty Version 1 available
 			// that is unlocked and can be updated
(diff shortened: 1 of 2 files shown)


spiralman:
@philippschulte Not sure what the procedure is for requesting code review in this project, but it seems like you are the primary maintainer. Do you think we could get this into the next release, if you're OK with the solution?

fastly/resource_fastly_service_v1.go
```diff
@@ -1618,7 +1621,7 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 		}
 
 		// update general settings
-		if d.HasChange("default_host") || d.HasChange("default_ttl") {
+		if d.HasChange("default_host") || d.HasChange("default_ttl") || (d.Get("default_ttl") == 0 && initialVersion) {
```
phamann, line 1629: Can you please add a comment above this explaining why the third conditional exists.

phamann (changes requested):
:+1: Thank you for your contribution @spiralman I'm happy with the implementation. Only one minor request around adding a comment to explain why the logic exists.

spiralman:
@phamann Thanks for the review! Updated the PR with a code comment explaining why the check is necessary. Thanks!

phamann (approved):
👍 LGTM

spiralman:
@phamann Thanks for the review! Is there anything else necessary to get this merged and into a release? 

Currently, this bug is hindering our ability to roll out some new services in Fastly. We'd rather not have to build some sort of automation on top of Terraform (i.e. run `apply` twice) to work around the bug, if we can get a fix out relatively soon.

Please let me know if there's anything else we can do to help.

phamann:
@spiralman apologies, I also wanted to confirm @philippschulte is happy with this before merging.

philippschulte (approved):
@spiralman the changes look good to me! Thanks for also including the tests! I'll reach out to Hashicorp to request a new release.

spiralman:
Great, thanks very much @philippschulte and @phamann!

PR #232, opened 2020-05-15
Update Fastly: service_acl_entries_v1 markdown

alpagot:
Help with setup steps involved to provision an ACL. You cannot add the var/locals, ACL and resource all in one go, so making this clearer to the reader.

### DIFF
--- a/website/docs/r/service_acl_entries_v1.html.markdown
+++ b/website/docs/r/service_acl_entries_v1.html.markdown
@@ -10,6 +10,8 @@ description: |-
 
 Defines a set of Fastly ACL entries that can be used to populate a service ACL.  This resource will populate an ACL with the entries and will track their state.
 
+Note: You must plan / apply the acl definition along with the variable/locals first. Once provisioned you can then add in the fastly_service_acl_entries_v1 resource to populate the ACL container with entries. 
+
 ~> **Warning:** Terraform will take precedence over any changes you make in the UI or API. Such changes are likely to be reversed if you run Terraform again.  
 
 If Terraform is being used to populate the initial content of an ACL which you intend to manage via API or UI, then the lifecycle `ignore_changes` field can be used with the resource.  An example of this configuration is provided below.    
--- a/website/docs/r/service_acl_entries_v1.html.markdown
+++ b/website/docs/r/service_acl_entries_v1.html.markdown
@@ -10,7 +10,7 @@ description: |-
 
 Defines a set of Fastly ACL entries that can be used to populate a service ACL.  This resource will populate an ACL with the entries and will track their state.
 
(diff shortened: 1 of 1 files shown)


website/docs/r/service_acl_entries_v1.html.markdown
```diff
@@ -10,6 +10,8 @@ description: |-
 
 Defines a set of Fastly ACL entries that can be used to populate a service ACL.  This resource will populate an ACL with the entries and will track their state.
 
+Note: You must plan / apply the ACL definition along with the variable/locals first. Once provisioned you can then add in the fastly_service_acl_entries_v1 resource to populate the ACL container with entries.
```
Integralist, line 13:
```suggestion
You must plan/apply the [ACL definition](https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_v1#nestedblock--acl) (along any variables/locals first) in your service configuration. Once the ACL has been provisioned you can then add in the fastly_service_acl_entries_v1 resource to populate the ACL container with entries. 
```

Integralist: Also can we validate if we even need this addition or if we could update the examples to use [`depends_on`](https://www.terraform.io/docs/language/meta-arguments/depends_on.html) in the `fastly_service_acl_entries_v1` block example below?

e.g.

```terraform
resource "fastly_service_acl_entries_v1" "entries" {
  service_id = fastly_service_v1.myservice.id
  acl_id = {for d in fastly_service_v1.myservice.acl : d.name => d.acl_id}[var.myacl_name]
  entry {
    ip = "127.0.0.1"
    subnet = "24"
    negated = false
    comment = "ALC Entry 1"
  }
  depends_on = [fastly_service_v1.myservice]
}
```

alpagot: I've not used depends_on before. Looks like it came into play in TF 0.13 which was GA Aug 2020. There's been a few versions since and now TF 1.0 so it might still be worth referencing for customers using older versions.

smaeda-ks: Hey, @alpagot

> Looks like it came into play in TF 0.13 which was GA Aug 2020

What you're referring to is:
https://www.terraform.io/docs/language/meta-arguments/depends_on.html

> Version note: Module support for `depends_on` was added in Terraform 0.13, and previous versions can only use it with resources.

In this case, we're talking about `resource`, not `module`, and `depends_on` was added quite some time ago. So I'm +1 with @Integralist 's suggestion.

alpagot:
Closing as per comments above. This PR is not needed.

PR #234, opened 2020-05-19
Add support for Elasticsearch logging

ezkl:
Adds support for configuring [Elasticsearch log streaming](https://docs.fastly.com/en/guides/log-streaming-elasticsearch) ([API docs](https://developer.fastly.com/reference/api/logging/elasticsearch/)) to provider.

cc: @phamann


fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,284 @@
...
+				Optional:    true,
+				Description: "BasicAuth password.",
+				Sensitive:   true,
+			},
+
+			"request_max_entries": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The maximum number of logs sent in one request.",
+			},
```
mccurdyc, line 59: Should we include a validation function since we cast this to uint later?

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,284 @@
...
+
+			"request_max_bytes": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The maximum number of bytes sent in one request.",
```
mccurdyc, line 64: Should we include a validation function since we cast this to uint later?

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,284 @@
...
+		esl = append(esl, nel)
+	}
+
+	return esl
+}
+
+func buildCreateElasticsearch(elasticsearchMap interface{}, serviceID string, serviceVersion int) *fst.CreateElasticsearchInput {
+	df := elasticsearchMap.(map[string]interface{})
+
+	opts := fst.CreateElasticsearchInput{
```
mccurdyc: nitpick: 
```suggestion
	return &fst.CreateElasticsearchInput{
```

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,284 @@
...
+		ResponseCondition: fst.NullString(df["response_condition"].(string)),
+	}
+
+	return &opts
+}
+
+func buildDeleteElasticsearch(elasticsearchMap interface{}, serviceID string, serviceVersion int) *fst.DeleteElasticsearchInput {
+	df := elasticsearchMap.(map[string]interface{})
+
+	opts := fst.DeleteElasticsearchInput{
```
mccurdyc: nitpick: 
```suggestion
	return &fst.DeleteElasticsearchInput{
```

fastly/resource_fastly_service_v1.go
```diff
@@ -1252,6 +1252,8 @@ func resourceServiceV1() *schema.Resource {
 			},
 
 			"httpslogging": httpsloggingSchema,
+
+			"logging_elasticsearch": elasticsearchSchema,
```
mccurdyc, line 1256: :heart: the builder pattern

mccurdyc (commented):
A few nitpicks. Looking good.

Will do a full pass tomorrow.

fastly/block_fastly_service_v1_logging_elasticsearch.go
mccurdyc: @ezkl and I talked offline. This is validated by the API and the failure mode is handled nicely. Therefore, to avoid the possibility of validation drift, omitting the validation here.

fastly/block_fastly_service_v1_logging_elasticsearch.go
mccurdyc: We decided that we forgo this nitpick now, but will start applying in all new changes.

This wasn't picked up by `gofmt -s fastly/block_fastly_service_v1_logging_elasticsearch.go` or `staticcheck`. (See list of staticcheck checks [here](https://staticcheck.io/docs/checks))

mccurdyc (approved):
LGTM!

fastly/resource_fastly_service_v1.go
mccurdyc, line 1256: @ezkl should we stick with the pattern of the `logging` suffix (i.e., `elasticsearchlogging`) rather than changing to use `logging_` as a prefix, just for consistency with what already exists?

mccurdyc (commented):
@ezkl should we stick with the pattern of the `logging` suffix (i.e., `elasticsearchlogging`) rather than changing to use `logging_` as a prefix, just for consistency with what already exists?

This would be the only necessary change that I see worth going back and applying.

cc @phamann

fastly/resource_fastly_service_v1.go
mccurdyc: We discussed offline. This is the pattern that we would like to use going forward and will be the most prevalent with the staged changes.

mccurdyc:
Force-pushed rebasing master.

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,277 @@
+package fastly
+
+import (
+	"fmt"
+	"log"
+
+	fst "github.com/fastly/go-fastly/fastly"
```
phamann: Convention in the rest of the project is to call this `gofastly`.

fastly/block_fastly_service_v1_logging_elasticsearch.go
phamann: What would the validation function do? As I believe you get that for free when defining the schema type.

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,277 @@
...
+		ne = new(schema.Set)
+	}
+
+	oes := oe.(*schema.Set)
+	nes := ne.(*schema.Set)
+
+	removeElasticsearchLogging := oes.Difference(nes).List()
+	addElasticsearchLogging := nes.Difference(oes).List()
+
+	// DELETE old Elasticsearch logging endpoints
```
phamann:
```suggestion
	// DELETE old Elasticsearch logging endpoints.
```

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,277 @@
...
+		opts := buildDeleteElasticsearch(of, serviceID, latestVersion)
+
+		log.Printf("[DEBUG] Fastly Elasticsearch logging endpoint removal opts: %#v", opts)
+
+		if err := deleteElasticsearch(conn, opts); err != nil {
+			return err
+		}
+	}
+
+	// POST new/updated Elasticsearch logging endpoints
```
phamann:
```suggestion
	// POST new/updated Elasticsearch logging endpoints.
```

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,277 @@
...
+		if err := createElasticsearch(conn, opts); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func readElasticsearch(conn *fst.Client, d *schema.ResourceData, s *fst.ServiceDetail) error {
+	// refresh Elasticsearch
```
phamann:
```suggestion
	// Refresh Elasticsearch.
```

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,277 @@
...
+			"request_max_bytes":   el.RequestMaxBytes,
+			"placement":           el.Placement,
+			"tls_ca_cert":         el.TLSCACert,
+			"tls_client_cert":     el.TLSClientCert,
+			"tls_client_key":      el.TLSClientKey,
+			"tls_hostname":        el.TLSHostname,
+			"format_version":      el.FormatVersion,
+		}
+
+		// prune any empty values that come from the default string value in structs
```
phamann:
```suggestion
		// Prune any empty values that come from the default string value in structs.
```

fastly/block_fastly_service_v1_logging_elasticsearch_test.go
```diff
@@ -0,0 +1,336 @@
...
+			return fmt.Errorf("Elasticsearch List count mismatch, expected (%d), got (%d)", len(elasticsearch), len(elasticsearchList))
+		}
+
+		log.Printf("[DEBUG] elasticsearchList = %#v\n", elasticsearchList)
+
+		var found int
+		for _, e := range elasticsearch {
+			for _, el := range elasticsearchList {
+				if e.Name == el.Name {
+					// we don't know these things ahead of time, so populate them now
```
phamann:
```suggestion
					// We don't know these things ahead of time, so populate them now.
```

fastly/block_fastly_service_v1_logging_elasticsearch_test.go
```diff
@@ -0,0 +1,336 @@
...
+					e.ServiceID = service.ID
+					e.Version = service.ActiveVersion.Number
+					// We don't track these, so clear them out because we also wont know
+					// these ahead of time
```
phamann:
```suggestion
					// We don't track these, so clear them out because we also won't know
					// these ahead of time.
```

fastly/block_fastly_service_v1_logging_elasticsearch.go
```diff
@@ -0,0 +1,277 @@
...
+			}
+		}
+
+		esl = append(esl, nel)
+	}
+
+	return esl
+}
+
+func buildCreateElasticsearch(elasticsearchMap interface{}, serviceID string, serviceVersion int) *fst.CreateElasticsearchInput {
```
phamann: You mentioned in another PR that by abstracting these methods it helps with unit test dependency injection. However, I'm not seeing that in this PR. Can you help me understand the pattern more?

phamann (approved):
👍 LGTM other than some minor style comments and a question.

fastly/block_fastly_service_v1_logging_elasticsearch.go
mccurdyc: @phamann this PR has it.

The other PR was missing functions like this - https://github.com/terraform-providers/terraform-provider-fastly/pull/234/files#diff-481864a021c0c6dbb09378b5a70165d9R189-R192

The pattern is basically, 

`opts := build...()`
`someCRUDAction(opts)`

You can see a concrete implementation here - https://github.com/terraform-providers/terraform-provider-fastly/pull/234/files#diff-481864a021c0c6dbb09378b5a70165d9R156-R161

fastly/block_fastly_service_v1_logging_elasticsearch.go
mccurdyc: Should have also included this here -

@ezkl and I talked offline. This is validated by the API and the failure mode is handled nicely. Therefore, to avoid the possibility of validation drift, omitting the validation here.

PR #235, opened 2020-05-19
Add support for FTP logging

ezkl:
Adds support for configuring [FTP log streaming](https://docs.fastly.com/en/guides/log-streaming-ftp) ([API docs](https://developer.fastly.com/reference/api/logging/ftp/)) to provider.

cc: @phamann


fastly/block_fastly_service_v1_logging_ftp.go
```diff
@@ -0,0 +1,268 @@
...
+	Elem: &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			// Required fields
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The unique name of the FTP logging endpoint",
+			},
+
+			"address": {
```
mccurdyc, line 23: Just making note that for FTP we don't explicitly expose IPV4 or Hostname because Address includes these values.

https://developer.fastly.com/reference/api/logging/ftp/

fastly/block_fastly_service_v1_logging_ftp.go
```diff
@@ -0,0 +1,268 @@
...
+func deleteFTP(conn *fst.Client, i *fst.DeleteFTPInput) error {
+	err := conn.DeleteFTP(i)
+
+	if errRes, ok := err.(*fst.HTTPError); ok {
+		if errRes.StatusCode != 404 {
+			return err
+		}
+	} else if err != nil {
+		return err
+	}
```
mccurdyc: Wouldn't a 404 error returned from `DeleteFTP` result in a nil error being returned from this function? Is that expected?

mccurdyc: Why can't we simply have the following here:

```golang
return conn.DeleteFTP(i)
```

fastly/resource_fastly_service_v1.go
```diff
@@ -1574,6 +1575,7 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 		"splunk",
 		"blobstoragelogging",
 		"httpslogging",
+		"logging_ftp",
```
mccurdyc, line 1581: @ezkl should we stick with the pattern of the `logging` suffix (i.e., `ftplogging`) rather than changing to use `logging_` as a prefix, just for consistency with what already exists?

This would be the only necessary change that I see worth going back and applying.

cc @phamann

mccurdyc (commented):
@ezkl should we stick with the pattern of the `logging` suffix (i.e., `ftplogging`) rather than changing to use `logging_` as a prefix, just for consistency with what already exists?

This would be the only necessary change that I see worth going back and applying.

cc @phamann

mccurdyc:
Force-pushed rebasing master.

mccurdyc:
Force-pushed rebasing master.

fastly/block_fastly_service_v1_logging_ftp_test.go
```diff
@@ -0,0 +1,286 @@
...
+    address  = "ftp.example.com"
+    user     = "user"
+		public_key = <<EOF
+`+pgpPublicKey()+`
+EOF
+    password = "p@ssw0rd"
+    path     = "/path"
+    port     = 27
+    format   = "%%h %%l %%u %%t \"%%r\" %%>s %%b"
+		gzip_level = 3
```
mccurdyc: This really bothers me. This must be my tab width or something. It looks fine locally.

It would be nice if we could leverage Terraform's `terraform fmt`. Maybe if we pulled these out into `*.tf` test fixtures?

But I also kind of don't like pulling the test data that far from the test itself. Thoughts? Other ideas?

fastly/block_fastly_service_v1_logging_ftp_test.go
```diff
@@ -0,0 +1,286 @@
...
+		Placement:       "none",
+	}
+
+	log1_after_update := fst.FTP{
+		Version:         1,
+		Name:            "ftp-endpoint",
+		Address:         "ftp2.example.com",
+		Username:        "user",
+		Password:        "p@ssw0rd2",
+		PublicKey:       pgpPublicKey() + "\n", // The '\n' is necessary becaue of the heredocs (i.e., EOF) in the config below.
```
mccurdyc, line 95: Note this ugly `+ "\n"` and accompanying comment.

This is b/c of [this](https://github.com/terraform-providers/terraform-provider-fastly/pull/235/files#diff-16cf93bc3363904e6d40ed935a91db27R274).

fastly/block_fastly_service_v1_logging_ftp_test.go
```diff
@@ -0,0 +1,286 @@
+package fastly
+
+import (
+	"fmt"
+	"log"
+	"testing"
+
+	fst "github.com/fastly/go-fastly/fastly"
```
mccurdyc: Fix. `fst` -> `gofastly` in test.

fastly/resource_fastly_service_v1.go
phamann: I agree with this decision of starting to introduce consistency with `logging_foo`

fastly/block_fastly_service_v1_logging_ftp_test.go
phamann: I think the idea of .tf test fixtures to solve this issue, though does add some overhead to adding tests.

phamann (approved):
👍 LGTM

PR #249, opened 2020-06-08
S3 Logging PublicKey field support

mccurdyc:
## Proposed Changes

* Adds support for configuring the PGP `public_key` field for `s3logging`.
    * https://github.com/terraform-providers/terraform-provider-fastly/pull/249/commits/73753b43c26ed5af927bc55281127ecc9f674d3a
* Updates `s3logging` to use new block pattern.
    * https://github.com/terraform-providers/terraform-provider-fastly/pull/249/commits/0f3fa7a853aea6824da506d52b48dc9faaa231b1

## Relevant Link(s) / Documentation

* [API Docs](https://developer.fastly.com/reference/api/logging/s3/)

## Validation

```bash
env FASTLY_API_KEY="foo" TF_ACC=1 make testacc TESTARGS='-run=[Ss]3'
```

cc @ezkl @phamann

### DIFF
--- a/fastly/block_fastly_service_v1_s3logging.go
+++ b/fastly/block_fastly_service_v1_s3logging.go
@@ -121,7 +121,8 @@ var s3loggingSchema = &schema.Schema{
 	},
 }
 
-func processS3Logging(d *schema.ResourceData, conn *gofastly.Client, latestVersion int) error {
+func processS3(d *schema.ResourceData, conn *gofastly.Client, latestVersion int) error {
+	serviceID := d.Id()
 	os, ns := d.GetChange("s3logging")
 	if os == nil {
 		os = new(schema.Set)
@@ -132,25 +133,18 @@ func processS3Logging(d *schema.ResourceData, conn *gofastly.Client, latestVersi
 
 	oss := os.(*schema.Set)
 	nss := ns.(*schema.Set)
+
 	removeS3Logging := oss.Difference(nss).List()
 	addS3Logging := nss.Difference(oss).List()
 
 	// DELETE old S3 Log configurations
 	for _, sRaw := range removeS3Logging {
 		sf := sRaw.(map[string]interface{})
-		opts := gofastly.DeleteS3Input{
-			Service: d.Id(),
-			Version: latestVersion,
-			Name:    sf["name"].(string),
-		}
+		opts := buildDeleteS3(sf, serviceID, latestVersion)
 
 		log.Printf("[DEBUG] Fastly S3 Logging removal opts: %#v", opts)
-		err := conn.DeleteS3(&opts)
-		if errRes, ok := err.(*gofastly.HTTPError); ok {
-			if errRes.StatusCode != 404 {
-				return err
-			}
-		} else if err != nil {
+
+		if err := deleteS3(conn, opts); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -0,0 +1,313 @@
...
+				Optional:    true,
+				Default:     "%Y-%m-%dT%H:%M:%S.000",
+				Description: "specified timestamp formatting (default `%Y-%m-%dT%H:%M:%S.000`)",
+			},
+			"redundancy": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The S3 redundancy level.",
+			},
+			"public_key": {
```
mccurdyc, line 102: This is the new field. Otherwise, this schema was just copy-pasted.

mccurdyc:
Force-pushed a squashed commit to fix doc update.

fastly/block_fastly_service_v1_s3logging_test.go
```diff
@@ -471,7 +479,7 @@ type currentEnv struct {
 
 func getEnv() *currentEnv {
 	// Grab any existing Fastly AWS S3 keys and preserve, in the off chance
-	// they're actually set in the enviornment
+	// they're actually set in the environment
```
mccurdyc, line 543: Fixed a typo.

fastly/resource_fastly_service_v1.go
```diff
@@ -1252,6 +1140,7 @@ func resourceServiceV1() *schema.Resource {
 			},
 
 			"httpslogging": httpsloggingSchema,
+			"s3logging":    s3loggingSchema,
```
mccurdyc: We can't switch to `logging_s3` as this would be a breaking change.

fastly/test-fixtures/gpg-pubkey.asc
```diff
@@ -0,0 +1,53 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
mccurdyc: TODO: use the test helper instead.

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -0,0 +1,313 @@
...
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("FASTLY_S3_SECRET_KEY", ""),
+				Description: "AWS Secret Key",
+				Sensitive:   true,
+			},
+			// Optional fields
+			"path": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Path to store the files. Must end with a trailing slash",
```
mccurdyc: We should probably add a validation function to this. (Not sure how the API responds without it.)

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -0,0 +1,331 @@
...
+				DefaultFunc: schema.EnvDefaultFunc("FASTLY_S3_SECRET_KEY", ""),
+				Description: "AWS Secret Key.",
+				Sensitive:   true,
+			},
+
+			// Optional fields
+			"path": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Path to store the files. Must end with a trailing slash.",
```
mccurdyc, line 50: Wondering if we should have a validation function for this. I guess it would depend on how the API handles this --- i.e., does it validate this?

mccurdyc: The backend does validate and ensure that a trailing slash is appended if it doesn't exist.

phamann: So the API appends a trailing slash if one is present? This would result in a diff always showing in the TF plan. As the statefile will have without and the API will have with?

fastly/block_fastly_service_v1_s3logging_test.go
```diff
@@ -3,10 +3,10 @@ package fastly
 import (
 	"fmt"
 	"os"
-	"reflect"
```
phamann, line 6: 🎉

phamann (approved):
:+1: LGTM - Other than the comment regarding trailing slash. You'll also need to rebase/fix merge conflicts.

fastly/block_fastly_service_v1_s3logging.go
mccurdyc: The API does append a trailing slash, but it wouldn't result in a diff in the TF plan because the infra state that gets stored is what is returned from the API.

In other words, a CREATE also calls a READ, which is what sets the TF state.

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -18,110 +19,133 @@ var s3loggingSchema = &schema.Schema{
...
 				Optional:    true,
-				Description: "Optional server-side KMS Key Id. Must be set if server_side_encryption is set to `aws:kms`",
+				Description: "Optional server-side KMS Key Id. Must be set if server_side_encryption is set to `aws:kms`.",
 			},
 		},
 	},
 }
 
-func processS3Logging(d *schema.ResourceData, conn *gofastly.Client, latestVersion int) error {
+func processS3(d *schema.ResourceData, conn *gofastly.Client, latestVersion int) error {
```
mccurdyc, line 146: Other endpoints use `processFoo`.

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -135,84 +159,103 @@ func processS3Logging(d *schema.ResourceData, conn *gofastly.Client, latestVersi
...
 
-		redundancy := strings.ToLower(sf["redundancy"].(string))
-		switch redundancy {
-		case "standard":
-			opts.Redundancy = gofastly.S3RedundancyStandard
-		case "reduced_redundancy":
-			opts.Redundancy = gofastly.S3RedundancyReduced
+	sl := flattenS3s(s3List)
+
+	if err := d.Set("s3logging", sl); err != nil {
```
mccurdyc: See this is where state gets set. Therefore, as long as the API always returns a value with a trailing slash, it shouldn't result in a diff. We don't currently have TF validation around this and I would imagine we would have heard about this if it were an issue.

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -135,84 +159,103 @@ func processS3Logging(d *schema.ResourceData, conn *gofastly.Client, latestVersi
...
+	s3List, err := conn.ListS3s(&gofastly.ListS3sInput{
+		Service: d.Id(),
+		Version: s.ActiveVersion.Number,
+	})
+
+	if err != nil {
+		return fmt.Errorf("[ERR] Error looking up S3 Logging for (%s), version (%v): %s", d.Id(), s.ActiveVersion.Number, err)
+	}
 
-		redundancy := strings.ToLower(sf["redundancy"].(string))
```
mccurdyc, line 190: This should show up in the `buildCreateS3` and it does.

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -135,84 +159,103 @@ func processS3Logging(d *schema.ResourceData, conn *gofastly.Client, latestVersi
...
+			"timestamp_format":                  s.TimestampFormat,
+			"redundancy":                        s.Redundancy,
+			"response_condition":                s.ResponseCondition,
+			"message_type":                      s.MessageType,
+			"placement":                         s.Placement,
+			"server_side_encryption":            s.ServerSideEncryption,
+			"server_side_encryption_kms_key_id": s.ServerSideEncryptionKMSKeyID,
 		}
 
-		encryption := sf["server_side_encryption"].(string)
```
mccurdyc, line 198: This should show up in the `buildCreateS3` and it does.

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -135,84 +159,103 @@ func processS3Logging(d *schema.ResourceData, conn *gofastly.Client, latestVersi
...
+			"domain":                            s.Domain,
+			"gzip_level":                        s.GzipLevel,
+			"format":                            s.Format,
+			"format_version":                    s.FormatVersion,
+			"timestamp_format":                  s.TimestampFormat,
+			"redundancy":                        s.Redundancy,
+			"response_condition":                s.ResponseCondition,
+			"message_type":                      s.MessageType,
+			"placement":                         s.Placement,
+			"server_side_encryption":            s.ServerSideEncryption,
```
mccurdyc: These values are API response values, hence why we don't have to do the `switch` statement.

PR #255, opened 2020-06-10
Make backend auto_loadbalance default to false

acme:
The Fastly web interface defaults "Auto load balance" to "No". The most common reason for having multiple backends in a single service is to route different paths to different backends, rather than load balance between different origins. Currently, this provider defaults `auto_loadbalance` to `true`, which is surprising and confusing. This commit makes `auto_loadbalance` default to `false` to reduce confusion.


phamann:
@acme As discussed internally. While this is a needed change, it's also considered a breaking change. Therefore we are going to hold of releasing until the future v1.0.0 release.

smaeda-ks:
Closing this PR as we opened up a new one for this: https://github.com/fastly/terraform-provider-fastly/pull/452 Thank you

PR #263, opened 2020-06-16
helpers: add test helper for escaping '%'

mccurdyc:
## Proposed Change(s):

* Add a test helper to escape '%'.
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/263/commits/b72e5050c754943895ddf341c9030ab49716d7e1
* Updates New Relic tests to use helper function.
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/263/commits/386626c91fd48a8f7ef150d22ad952ba334794f7
* Updates Datadog tests to use helper function.
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/263/commits/c9f6aa232ce14b5312c0b67bc17334554982768a

## Note(s)

* We have discussed offline that we are not ready to support a generic helper function for escaping special characters for a couple reasons:
   1. It is out of scope for now due to the possible edge cases and because we would want to update the codebase to use the helper if it existed.
   2. Terraform doesn't expose it.

## Validation

```bash
FASTLY_API_KEY="foo" TF_ACC=1 make testacc
```

### DIFF
--- a/fastly/fastly_test.go
+++ b/fastly/fastly_test.go
@@ -1,5 +1,7 @@
 package fastly
 
+import "strings"
+
 // pgpPublicKey returns a PEM encoded PGP public key suitable for testing.
 func pgpPublicKey() string {
 	return `-----BEGIN PGP PUBLIC KEY BLOCK-----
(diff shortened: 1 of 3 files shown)


fastly/helpers.go
```diff
@@ -0,0 +1,27 @@
...
+//
+// In case #2, we only have to single escape.
+//
+// Refer to the Terraform documentation on string literals for more details:
+// https://www.terraform.io/docs/configuration/expressions.html#string-literals
+func escapeSpecialChars(s string) string {
+	escapePercent := strings.ReplaceAll(s, "%", "%%")
+	escapePercentTemplate := strings.ReplaceAll(escapePercent, "%%{", "%%%%{")
+
+	escapeDollarSign := strings.ReplaceAll(escapePercentTemplate, "$", "$$")
```
mccurdyc: It can't be this naive. Diff when trying to use the helper in DataDog tests.

```
        -               `  "is_cacheable": %{if(fastly_info.state~"^(HIT|MISS)$", "true","false")}V,`,
        +               `  "is_cacheable": %{if(fastly_info.state~"^(HIT|MISS)$$", "true","false")}V,`,
```

mccurdyc: @phamann Maybe for now just escape '%'? Or omit the test helper?

fastly/fastly_test.go
```diff
@@ -164,3 +166,24 @@ BQADQQA/ugzBrjjK9jcWnDVfGHlk3icNRq0oV7Ri32z/+HQX67aRfgZu7KWdI+Ju
...
+// In case #2, we only have to single escape.
+//
+// Refer to the Terraform documentation on string literals for more details:
+// https://www.terraform.io/docs/configuration/expressions.html#string-literals
+func escapePercentSign(s string) string {
+	escapeSign := strings.ReplaceAll(s, "%", "%%")
+	escapeTemplateSequence := strings.ReplaceAll(escapeSign, "%%{", "%%%%{")
+
+	return escapeTemplateSequence
+}
```
phamann, line 192: Any reason why you didn't include the table tests you had previously in this PR? I would say they are still useful for a fiddly helper like this.

mccurdyc: I can add it back!

mccurdyc: https://github.com/terraform-providers/terraform-provider-fastly/pull/263/commits/fa9200d59c0107ba137a8f52e52b7e035b9a9cef

phamann (approved):
:+1: LGTM

PR #264, opened 2020-06-17
Handle TLS Whitespace

mccurdyc:
## Proposed Change(s)

* Uses [the `StateFunc` field](https://github.com/hashicorp/terraform-plugin-sdk/blob/v1.14.0/helper/schema/schema.go#L151-L165) for trimming leading and trailing whitespace characters in TLS-related fields.
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/264/commits/412688e5ec5dcdf1e47fa1e1009d5adfab3e50f8
   * Note that while this function does what we want, it does bring with it a bug. [Here is a relevant issue documenting the bug](https://github.com/hashicorp/terraform-plugin-sdk/issues/160#issuecomment-522935697).
   * [Example use in the google provider](https://github.com/terraform-providers/terraform-provider-google/blob/d0f40640e54121c3deec7f41be3431505b903b9c/google/resource_google_project.go#L71-L75)
* Fixes an issue in the Elasticsearch tests where default values should have been expected and were not
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/264/commits/6dfae5a2666f95d418bc7953d046490e51a92346
* Use `cmp.Diff` instead of `reflect.DeepEqual` in Elasticsearch flatten test.
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/264/commits/6a2c1d60c0847f4276acc3127810e209131f6ac3
* Moves `appendNewLine` helper function from Elasticsearch test to general test helpers file.
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/264/commits/0b650efee29839bca30baf6d0ddaca1fd6fed297
* Updates Splunk to use `StateFunc`.
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/264/commits/6f37592ea97032b51e4cb00289c1fa7f3c257824
* Updates HTTPS to use `StateFunc`.
   * https://github.com/terraform-providers/terraform-provider-fastly/pull/264/commits/f689b34542daf49414c4ceff95c216774e263083

## Validation

```
export FASTLY_API_KEY="foo"; export TF_ACC=1; make testacc
```


phamann (commented):
Couple of comments/questions:
- I can see `go-cmp` and `terraform-plugin-sdk` being updated in the vendor files, however am confused as to why `aws-sdk-go` has changed - is it a transient dep of one of those? Not an issue now, but its easier if vendoring is self contained to a single commit to make it easier for reviewers.
- I like the `StateFunc` approach to this and am now left thinking we should apply this to all logging endpoints that accept certs and our backend TLS fields too. Thoughts?
-

mccurdyc:
>I like the StateFunc approach to this and am now left thinking we should apply this to all logging endpoints that accept certs and our backend TLS fields too. Thoughts?

Absolutely! I will work on a commits that do this.

* Splunk - https://github.com/terraform-providers/terraform-provider-fastly/pull/264/commits/6f37592ea97032b51e4cb00289c1fa7f3c257824
* HTTPS - https://github.com/terraform-providers/terraform-provider-fastly/pull/264/commits/f689b34542daf49414c4ceff95c216774e263083

phamann (approved):
👍 LGTM

PR #267, opened 2020-06-19
OC Migrate Blocks to using AttributeHandlers

ghost:


fastly/resource_fastly_service_v1.go
```diff
@@ -12,6 +12,10 @@ import (
 
 var fastlyNoServiceFoundErr = errors.New("No matching Fastly Service found")
 
+var acl 			= NewServiceACL()
+var backend 		= NewServiceBackend()
+var bigquerylogging = NewServiceBigQueryLogging()
```
mccurdyc: Nitpick and arguable.

We could use the following instead. I don't think it's worth going back through and updating everywhere honestly.

```
var (
    acl ...
    backend ...
    bigquerylogging ...
)
```

fastly/block_fastly_service_v1_bigquerylogging.go
```diff
@@ -141,7 +154,7 @@ func processBigQueryLogging(d *schema.ResourceData, conn *gofastly.Client, lates
 	return nil
 }
 
-func readBigQueryLogging(conn *gofastly.Client, d *schema.ResourceData, s *gofastly.ServiceDetail) error {
+func (h *BigQueryLoggingServiceAttributeHandler) Read(d *schema.ResourceData, s *gofastly.ServiceDetail, conn *gofastly.Client) error {
```
mccurdyc, line 87: I do like that this is explicitly passed in. It's clear from the method signature that this makes a call to the Fastly API.

fastly/resource_fastly_service_v1.go
mccurdyc: It does look like it may need a `make fmt` though.

fastly/resource_fastly_service_v1.go
```diff
@@ -100,7 +104,7 @@ func resourceServiceV1() *schema.Resource {
 			"papertrail":         papertrailSchema,
 			"sumologic":          sumologicSchema,
 			"gcslogging":         gcsloggingSchema,
-			"bigquerylogging":    bigqueryloggingSchema,
+			bigquerylogging.GetKey():    bigquerylogging.GetSchema(),
```
mccurdyc: I love this pattern :)

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+	Read(d *schema.ResourceData, s *gofastly.ServiceDetail, conn *gofastly.Client) error
+
+	// Process creates or updates the attribute against the Fastly API
+	Process(d *schema.ResourceData, latestVersion int, conn *gofastly.Client) error
+
+	// HasChange returns whether the state of the attribute has changed against Terraform stored state
+	HasChange(d *schema.ResourceData) bool
+
+	// MustProcess returns whether we must Process the resource (usually HasChange==true but allowing exceptions)
+	MustProcess(d *schema.ResourceData, initialVersion bool) bool
```
mccurdyc, line 32: Not really blocking, but I would like others' thoughts on this.

I think the `initialVersion` param in this `MustProcess` might be a little restrictive to the potential uses of this function in the future.

I think I know why it exists now, but maybe we could use and interface or struct instead?

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+	return h.key
+}
+
+func (h *DefaultServiceAttributeHandler) HasChange(d *schema.ResourceData) bool {
+	return d.HasChange(h.key)
+}
+
+func (h *DefaultServiceAttributeHandler) MustProcess(d *schema.ResourceData, initialVersion bool) bool {
+	return h.HasChange(d)
+}
```
mccurdyc, line 52: Not blocking.

Comments on these two functions would be really appreciated because they are very similar and it might not be clear when to use one or the other.

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+
+// ServiceDefinition defines the data model for service definitions
+// There are two types of service: VCL and WASM. This interface specifies the data object from which service resources
+// are constructed.
+type ServiceDefinition interface {
+	// GetType returns whether this is a VCL or WASM service?
```
mccurdyc, line 58:
```suggestion
	// GetType returns whether this is a VCL or WASM service
```

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+		a.Register(s) // Mutates s, adding handler-specific schema items to the list
+	}
+
+	return s
+}
+
+// resourceCreate satisfies the Terraform resource schema Create "interface"
+// while injecting the ServiceDefinition into the true Create functionality
+func resourceCreate(serviceDef ServiceDefinition) schema.CreateFunc {
+	return func(data *schema.ResourceData, i interface{}) error {
```
mccurdyc, line 154: Suggestion:

It might be a little clearer if we propagate the `meta` param name to this level.
```suggestion
	return func(data *schema.ResourceData, meta interface{}) error {
```

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+		return resourceServiceCreate(data, i, serviceDef)
```
mccurdyc, line 155: (See below for an explanation).

```suggestion
                 m := i.(Meta) // assert the interface type or panic
		return resourceServiceCreate(data, m, serviceDef)
```

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+	}
+}
+
+// resourceServiceCreate provides service resource Create functionality
+func resourceServiceCreate(d *schema.ResourceData, meta interface{}, serviceDef ServiceDefinition) error {
+	if err := validateVCLs(d); err != nil {
+		return err
+	}
+
+	conn := meta.(*FastlyClient).conn
```
mccurdyc, line 189: Isn't `meta.(*FastlyClient)` suggesting that `meta` is a `*FastlyClient`?

Ah I see. Okay this makes sense.

Note: I know that the `interface{}` is to satisfy the Terraform interface.

**Something to think about and that might not work.**

I wonder if at this level, we could have a function signature like the following:

```go
type Meta struct {
    Conn *FastlyClient
}

func resourceServiceCreate(d *schema.ResourceData, meta Meta, serviceDef ServiceDefinition) error { ... }
```

This way the function signature wouldn't change if we needed to pass additional meta data downstream. We could do the interface assertion above too.

mccurdyc: Actually, maybe `meta` is just `conn FastlyClient`.

I would rather have the function signature change when we need to add additional dependencies rather than a `Meta` struct that hides a lot of information.

Even if we did go the `Meta` route, I would still like to see `FastlyClient` separate to indicate that it makes external calls.

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+
+				log.Printf("[DEBUG] Update Version opts: %#v", opts)
+				_, err := conn.UpdateVersion(&opts)
+				if err != nil {
+					return err
+				}
+			}
+		}
+
+		for _, a := range serviceDef.GetAttributeHandler() {
```
mccurdyc, line 304: Just noting that we had the following before:

```golang
		if d.HasChange("default_host") || d.HasChange("default_ttl") || (d.Get("default_ttl") == 0 && initialVersion) {
```

Could `GetAttributeHandler` return a different order of the attributes that introduces a bug? (Thinking out loud).

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+			if a.MustProcess(d, initialVersion) {
```
mccurdyc, line 305: I know we introduce `MustProcess` to handle `d.HasChange("default_ttl") || (d.Get("default_ttl") == 0 && initialVersion)`.

fastly/base_fastly_service_v1.go
mccurdyc: Generally speaking, I don't think order matters, but before, we did check `default_host` and `default_ttl` first and I think it could be possible that these aren't evaludated until later, which could mean that [Fastly Settings aren't updated](https://github.com/terraform-providers/terraform-provider-fastly/blob/master/fastly/resource_fastly_service_v1.go#L306-L310) until after we process something like a logging endpoint, for example. I'm not sure if this matters.

mccurdyc: Otherwise, this looks like logic parallels what is [currently in `master`](https://github.com/terraform-providers/terraform-provider-fastly/blob/5b471c351ba16454ad3905e4d8f36fce3c29263b/fastly/resource_fastly_user_v1.go).

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+			log.Printf("[INFO] Version (%v) has been pushed and validated", latestVersion)
+			log.Printf("[INFO] Visit https://manage.fastly.com/configure/services/%s/versions/%v and activate it manually", d.Id(), latestVersion)
+		}
+	}
+
+	return resourceServiceRead(d, meta, serviceDef)
+}
+
+// resourceServiceRead provides service resource Read functionality
+func resourceServiceRead(d *schema.ResourceData, meta interface{}, serviceDef ServiceDefinition) error {
```
mccurdyc, line 353: Same as above, at this level, we could possibly use `conn FastlyClient` instead of `meta interface{}`.

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+	d.Set("comment", s.Comment)
+	d.Set("version_comment", s.Version.Comment)
+	d.Set("active_version", s.ActiveVersion.Number)
+
+	// If CreateService succeeds, but initial updates to the Service fail, we'll
+	// have an empty ActiveService version (no version is active, so we can't
+	// query for information on it)
+	if s.ActiveVersion.Number != 0 {
+
+		for _, a := range serviceDef.GetAttributeHandler() {
```
mccurdyc, line 388: Looks like it [parallels `master`](https://github.com/terraform-providers/terraform-provider-fastly/blob/5b471c351ba16454ad3905e4d8f36fce3c29263b/fastly/resource_fastly_service_v1.go#L566-L571) other than settings could possibly be updated in a different order.

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+
+	} else {
+		log.Printf("[DEBUG] Active Version for Service (%s) is empty, no state to refresh", d.Id())
+	}
+
+	return nil
+}
+
+// resourceServiceDelete provides service resource Delete functionality
+func resourceServiceDelete(d *schema.ResourceData, meta interface{}, serviceDef ServiceDefinition) error {
```
mccurdyc, line 402: Perfectly parallel with master branch.

fastly/base_fastly_service_v1.go
```diff
@@ -0,0 +1,482 @@
...
+// that a Service has been deleted. GET requests to a deleted Service will
+// return 200 OK and have the full output of the Service for an unknown time
+// (days, in my testing). In order to determine if a Service is deleted, we
+// need to hit /service and loop the returned Services, searching for the one
+// in question. This endpoint only returns active or "alive" services. If the
+// Service is not included, then it's "gone"
+//
+// Returns a fastlyNoServiceFoundErr error if the Service is not found in the
+// ListServices response.
+func findService(id string, meta interface{}) (*gofastly.Service, error) {
```
mccurdyc, line 466: Perfectly parallel with master branch.

fastly/block_fastly_service_v1_backend.go
```diff
@@ -9,169 +9,19 @@ import (
...
-			"weight": {
-				Type:        schema.TypeInt,
-				Optional:    true,
-				Default:     100,
-				Description: "The portion of traffic to send to a specific origins. Each origin receives weight/total of the traffic.",
-			},
+func NewServiceBackend() ServiceAttributeDefinition {
+	return &BackendServiceAttributeHandler{
+		&DefaultServiceAttributeHandler{
+			key: "backend",
```
mccurdyc, line 19: ~I expected to see `schema` set here.~

This is done in the `Register` method.

mccurdyc: We have to set `key` here to be used below. I agree with setting it here and how it is used below as this will lend itself well for templating the `Register` methods for future additions. Rather than having to update the string in every `Register` function.

fastly/block_fastly_service_v1_dynamicsnippet.go
```diff
@@ -97,7 +78,7 @@ func processDynamicSnippet(d *schema.ResourceData, conn *gofastly.Client, latest
 	return nil
 }
 
-func readDynamicSnippet(conn *gofastly.Client, d *schema.ResourceData, s *gofastly.ServiceDetail) error {
+func (h *DynamicSnippetServiceAttributeHandler) Read(d *schema.ResourceData, s *gofastly.ServiceDetail, conn *gofastly.Client) error {
```
mccurdyc, line 81: I also want to point out how much I appreciate the constant use of `h` as the method receiver param name. This will make templating future additions easier. Thank you!

fastly/block_fastly_service_v1_logentries.go
```diff
@@ -127,7 +81,7 @@ func processLogentries(d *schema.ResourceData, conn *gofastly.Client, latestVers
 	return nil
 }
 
-func readLogentries(conn *gofastly.Client, d *schema.ResourceData, s *gofastly.ServiceDetail) error {
+func (h *LogentriesServiceAttributeHandler) Read(d *schema.ResourceData, s *gofastly.ServiceDetail, conn *gofastly.Client) error {
```
mccurdyc, line 84: Can we make sure that we use the same approach that you [used for Datadog](https://github.com/terraform-providers/terraform-provider-fastly/blob/53f1b973ce7b15e5baf5119211b9fe1b55aa25ab/fastly/block_fastly_service_v1_logging_datadog.go) throughout (i.e., [this](https://github.com/terraform-providers/terraform-provider-fastly/pull/267/files#diff-e7c883d632aa25f0ce157e91d526c92fR97) becomes `d.Set(h.GetKey(), ...)`.

mccurdyc: Similarly, here - https://github.com/terraform-providers/terraform-provider-fastly/pull/267/files#diff-e7c883d632aa25f0ce157e91d526c92fR24

fastly/block_fastly_service_v1_logging_elasticsearch_test.go
```diff
@@ -243,15 +243,9 @@ resource "fastly_service_v1" "foo" {
...
-		tls_ca_cert        = <<EOF
-`+caCert()+`
-EOF
-		tls_client_cert    = <<EOF
-`+certificate()+`
-EOF
-		tls_client_key     = <<EOF
-`+privateKey()+`
-EOF
+		tls_ca_cert       = file("fastly_test_cacert")
```
mccurdyc: :smile:

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -155,41 +34,30 @@ func processS3(d *schema.ResourceData, conn *gofastly.Client, latestVersion int)
...
-		sf := sRaw.(map[string]interface{})
-
-		opts, err := buildCreateS3(sf, serviceID, latestVersion)
-		if err != nil {
-			return err
-		}
-
-		log.Printf("[DEBUG] Create S3 Logging Opts: %#v", opts)
-		_, err = conn.CreateS3(opts)
+		opts, _ := buildCreateS3(sRaw, d.Id(), latestVersion)
```
mccurdyc, line 51: I didn't expect to see a diff here, but this looks fine.

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -155,41 +34,30 @@ func processS3(d *schema.ResourceData, conn *gofastly.Client, latestVersion int)
...
 
 	// DELETE old S3 Log configurations.
 	for _, sRaw := range removeS3Logging {
-		sf := sRaw.(map[string]interface{})
-		opts := buildDeleteS3(sf, serviceID, latestVersion)
-
-		log.Printf("[DEBUG] Fastly S3 Logging removal opts: %#v", opts)
-
-		if err := deleteS3(conn, opts); err != nil {
+		opts := buildDeleteS3(sRaw, serviceID, latestVersion)
```
mccurdyc, line 42: I didn't expect to see a diff here, but this looks fine.

fastly/block_fastly_service_v1_s3logging.go
```diff
@@ -216,6 +204,8 @@ func createS3(conn *gofastly.Client, i *gofastly.CreateS3Input) error {
 }
 
 func deleteS3(conn *gofastly.Client, i *gofastly.DeleteS3Input) error {
+	log.Printf("[DEBUG] Fastly S3 Logging removal opts: %#v", i)
```
mccurdyc, line 207: :+1:

fastly/fastly_test.go
```diff
@@ -1,176 +1,40 @@
...
-MGhHeHMbNR2bb6JxGja8BJqbDcNwYaV0PRiMV2t9jFraRuYclMj+5gYFq4UvvzK3
-IbJL9LQ/uoFkKrsx/lUTRnDlMF+FDuZ5lLOEPcWWPvF1NBiH85H1b9bw83vBKinD
-8UwYnCiq6SLHj4aGajKvMeq/JzyywRh0htUiajlkG2nowShJIhySd1s+40ekxQFG
-/2T0w0Fv79DMzaWR4LmWoSDSfFCQi2874XZI+WOVHxkCcr9MUD2hDCaW8os/2Zav
-26wENaFBpuLwGv9ppcXV1YHhxB2aVNOgLWP7mPdUCMP9dfJofuYbwY49GtPcrcWy
-zny1Rn8/BNa/GvYQzfcMxDUIk5xnv6/olmEDWswUL0OHXOCOf+FnWHXso0gcnrAu
-1/YjhVeSYa8UN7gwO6+rOGlpY/XL60Fozg9vzTYDMrk=
-=FSt9
------END PGP PUBLIC KEY BLOCK-----`
+	return readTestFile("./fastly_test_publickey")
```
mccurdyc: I think we should put these files in a `fixtures` sub-directory.
```suggestion
	return readTestFile("./fixtures/fastly_test_publickey")
```

mccurdyc: Or google uses `test-fixtures` - https://github.com/terraform-providers/terraform-provider-google/tree/master/google.

fastly/fastly_test.go
```diff
@@ -230,3 +94,7 @@ func TestEscapePercentSign(t *testing.T) {
 		})
 	}
 }
+
+func appendNewLine(s string) string {
```
mccurdyc, line 98: If this isn't used anymore (I don't think it should be), we can delete it.

fastly/fastly_test.go
```diff
@@ -1,176 +1,40 @@
...
-BgNVBAMTC0hlcm9uZyBZYW5nggEAMAwGA1UdEwQFMAMBAf8wDQYJKoZIhvcNAQEE
-BQADQQA/ugzBrjjK9jcWnDVfGHlk3icNRq0oV7Ri32z/+HQX67aRfgZu7KWdI+Ju
-Wm7DCfrPNGVwFWUQOmsPue9rZBgO
------END CERTIFICATE-----`
+	return readTestFile("./fastly_test_cacert")
 }
 
-func appendNewLine(s string) string {
-	return s + "\n"
+func readTestFile(filename string) string {
```
mccurdyc: What if instead, we made it look more obvious that this was a test helper by having a `t *testing.T` parameter and calling `t.Fatal` instead of `panic`?

fastly/resource_fastly_service_v1.go
```diff
@@ -1,764 +1,52 @@
...
-	// findService above returned something and nil error, but shouldn't have
-	return fmt.Errorf("[WARN] Tried deleting Service (%s), but was still found", d.Id())
-
+// Ordering is important - stored is processing order
+// Conditions need to be updated first, as they can be referenced by other
+// configuration objects (Backends, Request Headers, etc)
+var vclService = &BaseServiceDefinition{
+	Type: "vcl",
+	Attributes: []ServiceAttributeDefinition{
+		NewServiceSettings(),
```
mccurdyc, line 13: Looks like `Settings` are first in an ordered slice, so disregard my comments above.

mccurdyc (approved):
LGTM! This is amazing! This pattern will certainly improve maintainability and adding new resources.

The only action items:

1. Consider making `MustProcess` method more dynamic by having a more general param than `initialVersion bool`
2. Use `GetKey` pattern ubiquitously in logging endpoints versus the hardcoded string.
3. Move test fixtures to a `test-fixtures` sub-directory.
4. Update test helper for reading test fixtures to include a `*testing.T` param and call `t.Fatal` instead of `panic`.

fastly/resource_fastly_service_v1.go
ghost: Superceded

ghost:
Resolves: 
2. Use GetKey pattern ubiquitously in logging endpoints versus the hardcoded string.

ghost:
Resolves:
3. Move test fixtures to a test-fixtures sub-directory.

ghost:
Resolves:
4. Update test helper for reading test fixtures to include a *testing.T param and call t.Fatal instead of panic.

fastly/block_fastly_service_v1_logging_googlepubsub.go
```diff
@@ -9,74 +9,89 @@ import (
...
 		},
-	},
+	}
+	return nil
 }
 
-func processGooglePubSub(d *schema.ResourceData, conn *gofastly.Client, latestVersion int) error {
+func (h *GooglePubSubServiceAttributeHandler) Process(d *schema.ResourceData, latestVersion int, conn *gofastly.Client) error {
 	serviceID := d.Id()
 	oldLogCfg, newLogCfg := d.GetChange("logging_googlepubsub")
```
mccurdyc, line 96:
```suggestion
	oldLogCfg, newLogCfg := d.GetChange(h.GetKey())
```

fastly/block_fastly_service_v1_logging_scalyr.go
```diff
@@ -9,63 +9,78 @@ import (
...
 		},
-	},
+	}
+	return nil
 }
 
-func processScalyr(d *schema.ResourceData, conn *gofastly.Client, latestVersion int) error {
+func (h *ScalyrServiceAttributeHandler) Process(d *schema.ResourceData, latestVersion int, conn *gofastly.Client) error {
 	serviceID := d.Id()
 	oldLogCfg, newLogCfg := d.GetChange("logging_scalyr")
```
mccurdyc, line 85:
```suggestion
	oldLogCfg, newLogCfg := d.GetChange(h.GetKey())
```

fastly/test_fixtures/fastly_test_publickey
```diff
@@ -0,0 +1,53 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
mccurdyc, line 1: You may have to update this file once you rebase master.

mccurdyc:
- [ ] ~1. Consider making `MustProcess` method more dynamic by having a more general param than `initialVersion bool`~
   * We discussed offline. We will make this change when it is necessary as changing the API should be safe.
- [x] 2. Use `GetKey` pattern ubiquitously in logging endpoints versus the hardcoded string.
- [x] 3. Move test fixtures to a `test-fixtures` sub-directory.
- [x] 4. Update test helper for reading test fixtures to include a `*testing.T` param and call `t.Fatal` instead of `panic`.

Everything looks good to me! I saw that googlepubsub and scalyr did still need to be updated in one place to use the `GetKey` pattern.

PR #272, opened 2020-06-23
Adds Honeycomb logging support

mccurdyc:
## Proposed Change(s)

* Adds support for provisioning Honeycomb logging endpoints.

## Relevant Link(s) / Doc(s)

* [API Docs](https://developer.fastly.com/reference/api/logging/honeycomb/)
* [Relevant Fastly API Go client library file](https://github.com/fastly/go-fastly/blob/master/fastly/honeycomb.go)

## Validation

```bash
export FASTLY_API_KEY="foo"; export TF_ACC=1; make testacc TESTARGS='-run=[hH]oneycomb'
```

### DIFF
--- a/fastly/block_fastly_service_v1_logging_honeycomb.go
+++ b/fastly/block_fastly_service_v1_logging_honeycomb.go
@@ -1,0 +1,220 @@
+package fastly
+
+import (
+	"fmt"
+	"log"
+
+	gofastly "github.com/fastly/go-fastly/fastly"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+type HoneycombServiceAttributeHandler struct {
+	*DefaultServiceAttributeHandler
+}
+
+func NewServiceLoggingHoneycomb() ServiceAttributeDefinition {
+	return &HoneycombServiceAttributeHandler{
+		&DefaultServiceAttributeHandler{
+			key: "logging_honeycomb",
+		},
+	}
+}
+
+func (h *HoneycombServiceAttributeHandler) Process(d *schema.ResourceData, latestVersion int, conn *gofastly.Client) error {
+	serviceID := d.Id()
+	ol, nl := d.GetChange(h.GetKey())
+
+	if ol == nil {
+		ol = new(schema.Set)
+	}
+	if nl == nil {
+		nl = new(schema.Set)
+	}
+
+	ols := ol.(*schema.Set)
+	nls := nl.(*schema.Set)
+
+	removeHoneycombLogging := ols.Difference(nls).List()
+	addHoneycombLogging := nls.Difference(ols).List()
+
+	// DELETE old Honeycomb logging endpoints.
+	for _, oRaw := range removeHoneycombLogging {
+		of := oRaw.(map[string]interface{})
+		opts := buildDeleteHoneycomb(of, serviceID, latestVersion)
+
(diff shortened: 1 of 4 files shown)


fastly/block_fastly_service_v1_logging_honeycomb.go
```diff
@@ -0,0 +1,220 @@
...
+					Required:    true,
+					Description: "The Honeycomb Dataset you want to log to.",
+				},
+
+				// Optional fields
+				"format": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "Apache style log formatting. Your log must produce valid JSON that Honeycomb can ingest.",
+				},
```
mccurdyc, line 194: On `manage.fastly.com`, there is a suggested value --- I don't think it is actually a default.

```json
{
  "time":"%{begin:%Y-%m-%dT%H:%M:%SZ}t",
  "data":  {
    "service_id":"%{req.service_id}V",
    "time_elapsed":%D,
    "request":"%m",
    "host":"%{Fastly-Orig-Host}i",
    "url":"%{cstr_escape(req.url)}V",
    "protocol":"%H",
    "is_ipv6":%{if(req.is_ipv6, "true", "false")}V,
    "is_tls":%{if(req.is_ssl, "true", "false")}V,
    "is_h2":%{if(fastly_info.is_h2, "true", "false")}V,
    "client_ip":"%h",
    "geo_city":"%{client.geo.city.utf8}V",
    "geo_country_code":"%{client.geo.country_code}V",
    "server_datacenter":"%{server.datacenter}V",
    "request_referer":"%{Referer}i",
    "request_user_agent":"%{User-Agent}i",
    "request_accept_content":"%{Accept}i",
    "request_accept_language":"%{Accept-Language}i",
    "request_accept_charset":"%{Accept-Charset}i",
    "cache_status":"%{regsub(fastly_info.state, "^(HIT-(SYNTH)|(HITPASS|HIT|MISS|PASS|ERROR|PIPE)).*", "\\2\\3") }V",
    "status":"%s",
    "content_type":"%{Content-Type}o",
    "req_header_size":%{req.header_bytes_read}V,
    "req_body_size":%{req.body_bytes_read}V,
    "resp_header_size":%{resp.header_bytes_written}V,
    "resp_body_size":%{resp.body_bytes_written}V
  }
}
```

I confirmed that this is not actually a defined default value according to the backend API.

fastly/block_fastly_service_v1_logging_honeycomb_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+		if err != nil {
+			return fmt.Errorf("[ERR] Error looking up Honeycomb Logging for (%s), version (%d): %s", service.Name, service.ActiveVersion.Number, err)
+		}
+
+		if len(honeycombList) != len(honeycomb) {
+			return fmt.Errorf("Honeycomb List count mismatch, expected (%d), got (%d)", len(honeycomb), len(honeycombList))
+		}
+
+		log.Printf("[DEBUG] honeycombList = %#v\n", honeycombList)
```
phamann, line 171: Is this still needed?

phamann (approved):
LGTM - other than non-blocking question

fastly/block_fastly_service_v1_logging_honeycomb_test.go
mccurdyc: It could be helpful for inspecting global state in debug logs.

PR #274, opened 2020-06-24
Adds OpenStack logging support

mccurdyc:
## Proposed Change(s)

* Adds support for provisioning OpenStack logging endpoints.

## Relevant Link(s) / Doc(s)

* [API Docs](https://developer.fastly.com/reference/api/logging/openstack/)
* [Relevant Fastly API Go client library file](https://github.com/fastly/go-fastly/blob/master/fastly/openstack.go)

## Validation

```bash
export FASTLY_API_KEY="foo"; export TF_ACC=1; make testacc TESTARGS='-run=[oO]penstack'
```

### DIFF
--- a/fastly/block_fastly_service_v1_logging_openstack.go
+++ b/fastly/block_fastly_service_v1_logging_openstack.go
@@ -1,0 +1,306 @@
+package fastly
+
+import (
+	"fmt"
+	"log"
+
+	gofastly "github.com/fastly/go-fastly/fastly"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+type OpenstackServiceAttributeHandler struct {
+	*DefaultServiceAttributeHandler
+}
+
+func NewServiceLoggingOpenstack() ServiceAttributeDefinition {
+	return &OpenstackServiceAttributeHandler{
+		&DefaultServiceAttributeHandler{
+			key: "logging_openstack",
+		},
+	}
+}
+
+func (h *OpenstackServiceAttributeHandler) Process(d *schema.ResourceData, latestVersion int, conn *gofastly.Client) error {
+	serviceID := d.Id()
+	ol, nl := d.GetChange(h.GetKey())
+
+	if ol == nil {
+		ol = new(schema.Set)
+	}
+	if nl == nil {
+		nl = new(schema.Set)
+	}
+
+	ols := ol.(*schema.Set)
+	nls := nl.(*schema.Set)
+
+	removeOpenstackLogging := ols.Difference(nls).List()
+	addOpenstackLogging := nls.Difference(ols).List()
+
+	// DELETE old OpenStack logging endpoints.
+	for _, oRaw := range removeOpenstackLogging {
+		of := oRaw.(map[string]interface{})
+		opts := buildDeleteOpenstack(of, serviceID, latestVersion)
+
+		log.Printf("[DEBUG] Fastly OpenStack logging endpoint removal opts: %#v", opts)
+
+		if err := deleteOpenstack(conn, opts); err != nil {
+			return err
+		}
+	}
+
+	// POST new/updated OpenStack logging endpoints.
+	for _, nRaw := range addOpenstackLogging {
+		lf := nRaw.(map[string]interface{})
+
+		// @HACK for a TF SDK Issue.
+		//
+		// This ensures that the required, `name`, field is present.
+		//
+		// If we have made it this far and `name` is not present, it is most-likely due
+		// to a defunct diff as noted here - https://github.com/hashicorp/terraform-plugin-sdk/issues/160#issuecomment-522935697.
+		//
+		// This is caused by using a StateFunc in a nested TypeSet. While the StateFunc
+		// properly handles setting state with the StateFunc, it returns extra entries
+		// during state Gets, specifically `GetChange("logging_openstack")` in this case.
+		if v, ok := lf["name"]; !ok || v.(string) == "" {
+			continue
+		}
+
+		opts := buildCreateOpenstack(lf, serviceID, latestVersion)
+
+		log.Printf("[DEBUG] Fastly OpenStack logging addition opts: %#v", opts)
+
+		if err := createOpenstack(conn, opts); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (h *OpenstackServiceAttributeHandler) Read(d *schema.ResourceData, s *gofastly.ServiceDetail, conn *gofastly.Client) error {
+	// Refresh OpenStack.
+	log.Printf("[DEBUG] Refreshing OpenStack logging endpoints for (%s)", d.Id())
+	openstackList, err := conn.ListOpenstack(&gofastly.ListOpenstackInput{
+		Service: d.Id(),
+		Version: s.ActiveVersion.Number,
+	})
+
+	if err != nil {
+		return fmt.Errorf("[ERR] Error looking up OpenStack logging endpoints for (%s), version (%v): %s", d.Id(), s.ActiveVersion.Number, err)
+	}
+
+	ell := flattenOpenstack(openstackList)
+
+	if err := d.Set(h.GetKey(), ell); err != nil {
+		log.Printf("[WARN] Error setting OpenStack logging endpoints for (%s): %s", d.Id(), err)
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


fastly/block_fastly_service_v1_logging_openstack.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func buildCreateOpenstack(openstackMap interface{}, serviceID string, serviceVersion int) *gofastly.CreateOpenstackInput {
+	df := openstackMap.(map[string]interface{})
+
+	return &gofastly.CreateOpenstackInput{
+		Service:           serviceID,
+		Version:           serviceVersion,
+		Name:              gofastly.NullString(df["name"].(string)),
+		URL:               gofastly.NullString(df["url"].(string)),
```
mccurdyc, line 167: Just noting that this URL is used for authentication. The name `url` aligns with the API and client library.

fastly/block_fastly_service_v1_logging_openstack.go
```diff
@@ -0,0 +1,289 @@
...
+				"period": {
+					Type:        schema.TypeInt,
+					Optional:    true,
+					Default:     3600,
+					Description: "How frequently log files are finalized so they can be available for reading (in seconds, default 3600).",
+				},
+
+				"path": {
+					Type:        schema.TypeString,
+					Optional:    true,
```
mccurdyc, line 257: Maybe we should set a default value of `"/"`?

We don't set a default for `S3` - https://github.com/terraform-providers/terraform-provider-fastly/blob/master/fastly/block_fastly_service_v1_s3logging.go#L112

`manage.fastly.com` sets the default to `"/"`, however, this is not set by the backend API.

fastly/block_fastly_service_v1_logging_openstack.go
```diff
@@ -0,0 +1,289 @@
...
+
+				"access_key": {
+					Type:        schema.TypeString,
+					Required:    true,
+					Sensitive:   true,
+					Description: "Your OpenStack account access key.",
+				},
+
+				// Optional fields
+				"public_key": {
```
mccurdyc, line 218: Not documented [here](https://developer.fastly.com/reference/api/logging/cloudfiles/), but supported by the API.

fastly/block_fastly_service_v1_logging_openstack.go
```diff
@@ -0,0 +1,289 @@
...
+					Type:        schema.TypeString,
+					Required:    true,
+					Description: "The name of your OpenStack container.",
+				},
+
+				"access_key": {
+					Type:        schema.TypeString,
+					Required:    true,
+					Sensitive:   true,
+					Description: "Your OpenStack account access key.",
```
mccurdyc, line 229: I don't think this is PEM format.

fastly/block_fastly_service_v1_logging_openstack.go
```diff
@@ -0,0 +1,289 @@
...
+				},
+
+				// Optional fields
+				"public_key": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "A PGP public key that Fastly will use to encrypt your log files before writing them to disk.",
```
mccurdyc, line 236: Need to add the `StateFunc` 
```golang
					// Related issue for weird behavior - https://github.com/hashicorp/terraform-plugin-sdk/issues/160
					StateFunc: trimSpaceStateFunc,
```

fastly/block_fastly_service_v1_logging_openstack.go
```diff
@@ -0,0 +1,289 @@
...
+		log.Printf("[DEBUG] Fastly OpenStack logging endpoint removal opts: %#v", opts)
+
+		if err := deleteOpenstack(conn, opts); err != nil {
+			return err
+		}
+	}
+
+	// POST new/updated OpenStack logging endpoints.
+	for _, nRaw := range addOpenstackLogging {
+		lf := nRaw.(map[string]interface{})
```
mccurdyc, line 54: To use the custom `StateFunc` and to ensure that it works when there is a trailing new line, add the following:

```suggestion
		lf := nRaw.(map[string]interface{})
		
		// @HACK for a TF SDK Issue.
		//
		// This ensures that the required, `name`, field is present.
		//
		// If we have made it this far and `name` is not present, it is most-likely due
		// to a defunct diff as noted here - https://github.com/hashicorp/terraform-plugin-sdk/issues/160#issuecomment-522935697.
		//
		// This is caused by using a StateFunc in a nested TypeSet. While the StateFunc
		// properly handles setting state with the StateFunc, it returns extra entries
		// during state Gets, specifically `GetChange("logging_kafka")` in this case.
		if v, ok := cfg["name"]; !ok || v.(string) == "" {
			continue
		}
```

fastly/block_fastly_service_v1_logging_openstack.go
phamann: Interesting. I'd personally say we don't set a default, as that means it is serialised and sent. It would be good to let OpenStack itself use its default, or can we check if Bulliet may set `/` if not present.

fastly/block_fastly_service_v1_openstack_test.go
```diff
@@ -0,0 +1,307 @@
...
+
+		if err != nil {
+			return fmt.Errorf("[ERR] Error looking up OpenStack Logging for (%s), version (%d): %s", service.Name, service.ActiveVersion.Number, err)
+		}
+
+		if len(openstackList) != len(openstack) {
+			return fmt.Errorf("OpenStack List count mismatch, expected (%d), got (%d)", len(openstack), len(openstackList))
+		}
+
+		log.Printf("[DEBUG] openstackList = %#v\n", openstackList)
```
phamann, line 181: Is this still needed?

website/docs/r/service_v1.html.markdown
```diff
@@ -656,6 +658,27 @@ The `logging_kafka` block supports:
...
+* `bucket_name` - (Required) The name of your OpenStack container.
+* `url` - (Required) Your OpenStack auth url.
+* `user` - (Required) The username for your OpenStack account.
+* `access_key` - (Required) Your OpenStack account access key.
+* `public_key` - (Optional) A PGP public key that Fastly will use to encrypt your log files before writing them to disk.
+* `path` - (Optional) Path to store the files. Must end with a trailing slash.
+If this field is left empty, the files will be saved in the bucket's root path.
+* `period` - (Optional) How frequently the logs should be transferred, in
+seconds. Default `3600`.
+* `gzip_level` - (Optional) What level of GZIP encoding to have when dumping logs (default 0, no compression).
```
phamann, line 709: I know this is taken verbatim from our docs, but this _should_ be Gzip not GZIP.

website/docs/r/service_v1.html.markdown
```diff
@@ -656,6 +658,27 @@ The `logging_kafka` block supports:
...
+* `message_type` - (Optional) How the message should be formatted; one of: `classic`, `loggly`, `logplex` or `blank`. Default `classic`. [Fastly Documentation](https://developer.fastly.com/reference/api/logging/gcs/)
+* `format` - (Optional) Apache style log formatting.
+* `format_version` - (Optional) The version of the custom logging format used for the configured endpoint. Can be either `1` or `2`. The logging call gets placed by default in `vcl_log` if `format_version` is set to `2` and in `vcl_deliver` if `format_version` is set to `1`. Default `2`.
+* `timestamp_format` - (Optional) The strftime specified timestamp formatting (default `%Y-%m-%dT%H:%M:%S.000`).
+* `response_condition` - (Optional) The name of an existing condition in the configured endpoint, or leave blank to always execute.
+* `placement` - (Optional) Where in the generated VCL the logging call should be placed; one of: `none` or `waf_debug`.
+
+
```
phamann: Is this second newline intentional?
```suggestion
```

phamann (approved):
LGTM - other than minor non-blocking comments

fastly/block_fastly_service_v1_openstack_test.go
mccurdyc: It does still help in debugging global state.

fastly/block_fastly_service_v1_logging_openstack.go
mccurdyc: From a quick grep, the backend doesn't set a default for path for any logging endpoint. We do have it noted that we should have the backend set `/` as the default when none is set by the client for logging endpoints where it would make sense to do so.

PR #275, opened 2020-06-24
Adds Rackspace Cloud Files logging support

mccurdyc:
## Proposed Change(s)

* Adds support for provisioning Rackspace Cloud Files logging endpoints.

## Relevant Link(s) / Doc(s)

* [API Docs](https://developer.fastly.com/reference/api/logging/cloudfiles/)
* [Relevant Fastly API Go client library file](https://github.com/fastly/go-fastly/blob/master/fastly/cloudfiles.go)

## Validation

```bash
export FASTLY_API_KEY="foo"; export TF_ACC=1; make testacc TESTARGS='-run=[cC]loudfiles'
```

### DIFF
--- a/fastly/block_fastly_service_v1_logging_cloudfiles.go
+++ b/fastly/block_fastly_service_v1_logging_cloudfiles.go
@@ -1,0 +1,305 @@
+package fastly
+
+import (
+	"fmt"
+	"log"
+
+	gofastly "github.com/fastly/go-fastly/fastly"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+type CloudfilesServiceAttributeHandler struct {
+	*DefaultServiceAttributeHandler
+}
+
+func NewServiceLoggingCloudfiles() ServiceAttributeDefinition {
+	return &CloudfilesServiceAttributeHandler{
+		&DefaultServiceAttributeHandler{
+			key: "logging_cloudfiles",
+		},
+	}
+}
+
+func (h *CloudfilesServiceAttributeHandler) Process(d *schema.ResourceData, latestVersion int, conn *gofastly.Client) error {
+	serviceID := d.Id()
+	ol, nl := d.GetChange(h.GetKey())
+
+	if ol == nil {
+		ol = new(schema.Set)
+	}
+	if nl == nil {
+		nl = new(schema.Set)
+	}
+
+	ols := ol.(*schema.Set)
+	nls := nl.(*schema.Set)
+
+	removeCloudfilesLogging := ols.Difference(nls).List()
+	addCloudfilesLogging := nls.Difference(ols).List()
+
+	// DELETE old Cloud Files logging endpoints.
+	for _, oRaw := range removeCloudfilesLogging {
+		of := oRaw.(map[string]interface{})
+		opts := buildDeleteCloudfiles(of, serviceID, latestVersion)
+
+		log.Printf("[DEBUG] Fastly Cloud Files logging endpoint removal opts: %#v", opts)
+
+		if err := deleteCloudfiles(conn, opts); err != nil {
+			return err
+		}
+	}
+
+	// POST new/updated Cloud Files logging endpoints.
+	for _, nRaw := range addCloudfilesLogging {
+		lf := nRaw.(map[string]interface{})
+
+		// @HACK for a TF SDK Issue.
+		//
+		// This ensures that the required, `name`, field is present.
+		//
+		// If we have made it this far and `name` is not present, it is most-likely due
+		// to a defunct diff as noted here - https://github.com/hashicorp/terraform-plugin-sdk/issues/160#issuecomment-522935697.
+		//
+		// This is caused by using a StateFunc in a nested TypeSet. While the StateFunc
+		// properly handles setting state with the StateFunc, it returns extra entries
+		// during state Gets, specifically `GetChange("logging_cloudfiles")` in this case.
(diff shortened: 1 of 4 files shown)


fastly/block_fastly_service_v1_logging_cloudfiles.go
```diff
@@ -0,0 +1,291 @@
...
+
+				"access_key": {
+					Type:        schema.TypeString,
+					Required:    true,
+					Sensitive:   true,
+					Description: "Your Cloudfile account access key.",
+				},
+
+				// Optional fields
+				"public_key": {
```
mccurdyc, line 226: Not documented [here](https://developer.fastly.com/reference/api/logging/cloudfiles/), but supported by the API.

fastly/block_fastly_service_v1_logging_cloudfiles.go
```diff
@@ -0,0 +1,291 @@
...
+
+	removeCloudfilesLogging := ols.Difference(nls).List()
+	addCloudfilesLogging := nls.Difference(ols).List()
+
+	// DELETE old Cloudfiles logging endpoints.
+	for _, oRaw := range removeCloudfilesLogging {
+		of := oRaw.(map[string]interface{})
+		opts := buildDeleteCloudfiles(of, serviceID, latestVersion)
+
+		log.Printf("[DEBUG] Fastly Cloudfiles logging endpoint removal opts: %#v", opts)
```
mccurdyc: Not sure if I should be using `Rackspace Cloud Files`, `Cloud Files` or simply `Cloudfiles` as the docs and `manage.fastly.com` use all of these.

fastly/block_fastly_service_v1_logging_cloudfiles.go
```diff
@@ -0,0 +1,291 @@
...
+					Type:        schema.TypeString,
+					Required:    true,
+					Description: "The username for your Cloudfile account.",
+				},
+
+				"access_key": {
+					Type:        schema.TypeString,
+					Required:    true,
+					Sensitive:   true,
+					Description: "Your Cloudfile account access key.",
```
mccurdyc, line 222: I don't think this is PEM format since it's just an API key.

fastly/block_fastly_service_v1_logging_cloudfiles.go
```diff
@@ -0,0 +1,291 @@
...
+		log.Printf("[DEBUG] Fastly Cloudfiles logging endpoint removal opts: %#v", opts)
+
+		if err := deleteCloudfiles(conn, opts); err != nil {
+			return err
+		}
+	}
+
+	// POST new/updated Cloudfiles logging endpoints.
+	for _, nRaw := range addCloudfilesLogging {
+		lf := nRaw.(map[string]interface{})
```
mccurdyc, line 54: To use the custom `StateFunc` and to ensure that it works when there is a trailing new line, add the following:

```suggestion
		lf := nRaw.(map[string]interface{})
		
		// @HACK for a TF SDK Issue.
		//
		// This ensures that the required, `name`, field is present.
		//
		// If we have made it this far and `name` is not present, it is most-likely due
		// to a defunct diff as noted here - https://github.com/hashicorp/terraform-plugin-sdk/issues/160#issuecomment-522935697.
		//
		// This is caused by using a StateFunc in a nested TypeSet. While the StateFunc
		// properly handles setting state with the StateFunc, it returns extra entries
		// during state Gets, specifically `GetChange("logging_cloudfiles")` in this case.
		if v, ok := cfg["name"]; !ok || v.(string) == "" {
			continue
		}
```

fastly/block_fastly_service_v1_logging_cloudfiles.go
phamann: Per their own product page, I'd go with "Cloud Files".
```suggestion
		log.Printf("[DEBUG] Fastly Cloud Files logging endpoint removal opts: %#v", opts)
```

phamann: Similar comments elided for rest of review.

fastly/block_fastly_service_v1_logging_cloudfiles.go
phamann: :+1: Good catch, can you chase this internally to make sure it gets documented.

fastly/block_fastly_service_v1_logging_cloudfiles_test.go
```diff
@@ -0,0 +1,317 @@
...
+
+		if err != nil {
+			return fmt.Errorf("[ERR] Error looking up Cloudfiles Logging for (%s), version (%d): %s", service.Name, service.ActiveVersion.Number, err)
+		}
+
+		if len(cloudfilesList) != len(cloudfiles) {
+			return fmt.Errorf("Cloudfiles List count mismatch, expected (%d), got (%d)", len(cloudfiles), len(cloudfilesList))
+		}
+
+		log.Printf("[DEBUG] cloudfilesList = %#v\n", cloudfilesList)
```
phamann, line 182: Is this still needed?

phamann (changes requested):
Looking good, lets get the naming changed.

fastly/block_fastly_service_v1_logging_cloudfiles_test.go
mccurdyc: This could still help with debugging global state.

phamann (approved):
:+1: LGTM

PR #277, opened 2020-06-24
Ensure PublicKey field is using trimSpaceStateFunc

mccurdyc:
## Proposed Change(s)

* Fixes a few places where the `trimSpaceStateFunc` was ommitted for uses of the `PublicKey` field.
  * `blobstoragelogging`
  * `s3logging`
  * `logging_ftp`

## Validation

```bash
$ export FASTLY_API_KEY="foo"; export TF_ACC=1; make testacc TESTARGS="-run=[sS]3"
$ export FASTLY_API_KEY="foo"; export TF_ACC=1; make testacc TESTARGS="-run=[bB]lob[sS]torage"
$ export FASTLY_API_KEY="foo"; export TF_ACC=1; make testacc TESTARGS="-run=[fF][tT][pP]"
```

## Note(s)

* This shouldn't introduce breaking changes for those using the `PublicKey` field already. However, it could introduce a state diff where a new line is removed.

## Thoughts

* I really dislike that this is something that you have to remember to put for these fields. This feeling is made even worse by having to remember to also include the hack.

### DIFF
--- a/fastly/block_fastly_service_v1_blobstoragelogging.go
+++ b/fastly/block_fastly_service_v1_blobstoragelogging.go
@@ -58,6 +58,21 @@ func (h *BlobStorageLoggingServiceAttributeHandler) Process(d *schema.ResourceDa
 	// POST new/updated Blob Storage logging configurations
 	for _, bslRaw := range add {
 		bslf := bslRaw.(map[string]interface{})
+
+		// @HACK for a TF SDK Issue.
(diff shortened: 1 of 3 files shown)


phamann (approved):
LGTM 

> I really dislike that this is something that you have to remember to put for these fields. This feeling is made even worse by having to remember to also include the hack.

😭

PR #279, opened 2020-06-29
GZIP -> Gzip

mccurdyc:
## Proposed Change(s)

* Standardize on `Gzip` ubiquitously.

## Validation:

```bash
find . -path ./vendor -prune -type f -exec sed -i 's/GZIP/Gzip/g' {} \;
```

### DIFF
--- a/fastly/block_fastly_service_v1_gzip.go
+++ b/fastly/block_fastly_service_v1_gzip.go
@@ -9,19 +9,19 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
-type GZIPServiceAttributeHandler struct {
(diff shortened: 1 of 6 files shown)


fastly/block_fastly_service_v1_gzip.go
```diff
@@ -9,19 +9,19 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
-type GZIPServiceAttributeHandler struct {
+type GzipServiceAttributeHandler struct {
```
mccurdyc, line 12: No one should really be consuming this API, so this should be safe.

phamann (approved):
:+1: LGTM

PR #283, opened 2020-07-07
Documentation Refactoring

ghost:
This PR introduces the website-src directory which contains go templates and a script to build and compose these templates into the full markdown documentation in website.

It also includes new documentation for Compute services.

PR #298, opened 2020-07-29
Migrate CI to GitHub Action workflows for pull requests and releasing

phamann:
### TL;DR
As part of the provider migration to our own organisation we also want to migrate the CI testing to GitHub Actions away from Travis. This also introduces a new CI workflow for creating GitHub releases on any tag  matching `v*` i.e. a semver release.

### Notes for reviewers
- The release workflow and goreleaser.yml are using the recommended templates from HashiCorp and are thus tried and tested.
- We have created our own fork of the GPG import action due to the sensitive nature of its key handling.
- As we now have to pre-compile the website src (introduced in #283) I have added a step to the `website` CI job which ensures there are no new changes in the git working tree. I.e. this will hopefully ensure the docs are always up-to-date.
- I've refrained from adding any improvements to the CI jobs, such as statichcek, import checks etc and will follow up in a subsequent PR once this workflow has proven to work.


.goreleaser.yml
```diff
@@ -0,0 +1,47 @@
...
+- env:
+    - CGO_ENABLED=0
+  mod_timestamp: '{{ .CommitTimestamp }}'
+  flags:
+    - -trimpath
+  ldflags:
+    - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
+  goos:
+    - freebsd
+    - windows
```
mccurdyc, line 14: Was this goreleaser config a HashiCorp-recommended config?

Windows isn't included in the CI test matrix, should we still include it as a release target?

mccurdyc (approved):
> I've refrained from adding any improvements to the CI jobs, such as statichcek, import checks etc and will follow up in a subsequent PR once this workflow has proven to work.

:+1: Great idea!

LGTM!

I'm less comfortable with the website stuff. Looks like on PR, it runs a make target to generate the website and ensure that those changes have been pushed up (i.e., no diff in CI), right?

---

Looks like the website check CI job is failing

```
0s
##[error]Process completed with exit code 2.
Run make website-test
echo "github.com/hashicorp/terraform-website not found in your GOPATH (necessary for layouts and assets), get-ting..."
github.com/hashicorp/terraform-website not found in your GOPATH (necessary for layouts and assets), get-ting...
git clone https://github.com/hashicorp/terraform-website /src/github.com/hashicorp/terraform-website
fatal: could not create leading directories of '/src/github.com/hashicorp/terraform-website': Permission denied
make: *** [website-test] Error 128
GNUmakefile:59: recipe for target 'website-test' failed
##[error]Process completed with exit code 2.```

.goreleaser.yml
phamann: Yeah this is the HashiCorp approved config, good call RE windows testing, will include in test matrix :+1:

phamann:
> I'm less comfortable with the website stuff. Looks like on PR, it runs a make target to generate the website and ensure that those changes have been pushed up (i.e., no diff in CI), right?

Correct! It will catch if someone has changed the documentation source but not ran the generation script and committed any changes to their PR.

However as you've pointed out, the test itself is not working 😂

.goreleaser.yml
phamann: Have added Windows and MacOS to the test matrix.

phamann:
@mccurdyc I've now fixed the website-test job and added Linux and MacOS to the matrix per your suggestion. Therefore, this is ready for one last review before I merge.

mccurdyc (approved):
LGTM!

ezkl (approved):

Issue #85, opened 2018-06-19
default_host does not get removed


siwyd:
When a service had a default_host variable defined and it's removed, the default host is not removed from the Fastly service.

joshsouza:
I am also impacted by this, and found that the `default_host` field inside `terraform.tfstate` remains whatever value was assigned to it last if removed. Setting the value explicitly to `""` does nothing as well, so at this point turning off the setting must be manually done.

Integralist:
Thank you for taking the time to open this issue.

I appreciate this was reported a long time ago (and I appreciate your patience awaiting our feedback) but I believe we may now have a fix in place to resolve the issue you're describing. We merged it this morning as the new minor version `v0.21.3`, here are some details about the changes:

- [Migrate Terraform provider to use latest Go-Fastly 2.0.0 release](https://github.com/fastly/terraform-provider-fastly/releases/tag/v0.21.3).
- [Key changes in Go-Fastly 2.0.0](https://github.com/fastly/terraform-provider-fastly/issues/85) that I think relate to this issue.

If possible, could you try upgrading to the latest version (`v0.21.3`) and let me know if this indeed resolves the issue.

Thanks again.

siwyd:
@Integralist Thanks for the feedback. Unfortunately I'm not working for a Fastly customer any longer, maybe @joshsouza can help out.

smaeda-ks:
This has been fixed in v0.33.0:
https://github.com/fastly/terraform-provider-fastly/blob/main/CHANGELOG.md

Sorry for taking such a long time! We appreciate your feedback.

PR #302, opened 2020-08-07
have docs recommend override_host for s3 buckets

brwsh:
override_host is a better way to do this to avoid the workaround for shielding where it is necessary to add the backend to the configured domains on the service. 

this would also help users to avoid the known bug where default_host is not removed properly: https://github.com/fastly/terraform-provider-fastly/issues/85


phamann (approved):
:+1: LGTM

PR #310, opened 2020-09-09
scalyr token should be hidden from plan output

rayzorinc:
The scalyr log token shouldn't be display from terraform plan output and should be marked as sensitive.


mccurdyc (approved):
LGTM! Thank you!

Issue #94, opened 2018-09-13
Use GetService instead of ListService


lanej:
Current implementation of `findService` uses `ListServices` where `GetService` would be much faster, especially for customers with larger service lists.

The `deleted_at` parameter is now returned from the api. https://docs.fastly.com/api/config#service_a884a9abd5af9723f6fcbb1ed13ae4cc

https://github.com/terraform-providers/terraform-provider-fastly/blob/f86a461b8dac294b8d919ea86431b6abd282477b/fastly/resource_fastly_service_v1.go#L2927-L2956

PR #311, opened 2020-09-10
Improve service read and delete logic

### LINKED ISSUE #132: Delay in Fastly API causes error while destroying service
We occasionally run into the following error:

1 error(s) occurred:
* module.draft.fastly_service_v1.\<name> (destroy): 1 error(s) occurred:
* fastly_service_v1.\<name>: [WARN] Tried deleting Service (\<serviceID>), but was still found

However when we check Fastly UI, the service is no longer there. 

Therefor it looks like there sometimes is a slight delay in deleting the service, which causes that the following request meant to ensure that the service was indeed deleted still retrieves the service. 

The problem is that this error stops the whole destroy process, which might lead to discrepancy between TF state and infrastructure and therefor rerun of destroy fails again with something else and a manual intervention is required to fix this.

For the record, we are using TF to manage temporary environments for testing and therefor we run TF destroy on about 1500+ Fastly services a week and this happens cca 3-10 times a week and in these cases there is about 50% chance in our case to corrupt the TF state so that rerun is not even possible. Therefor reproducing the issue might prove difficult and if you need more information, we are willing to cooperate. 

Please either find a solution to write a warning without breaking the destroy process, or ideally put a sensible retry logic there.

TF version: 0.11.13
Fastly provider version: 0.6

phamann:
Closes #132 #94

### TL;DR
Currently the `resourceServiceRead` and `resourceServiceDelete` logic for the service resource uses a very old and very hacky implementation of `listServices`, due to the fact the API didn't previously return a `deleted_at` property. Fast forward many years we can clean up this implementation to use `fastly.GetServiceDetails` directly and infer deleted and not found resources via the go-fastly interface. 

The currently logic is also prone to race conditions in the API and therefore leads to false positives/negatives around deletion which is painful if you use the provider in CI environments to spin up and tear down services quickly.

### DIFF
--- a/fastly/base_fastly_service_v1.go
+++ b/fastly/base_fastly_service_v1.go
@@ -421,25 +421,23 @@ func resourceServiceUpdate(d *schema.ResourceData, meta interface{}, serviceDef 
 func resourceServiceRead(d *schema.ResourceData, meta interface{}, serviceDef ServiceDefinition, isImport bool) error {
 	conn := meta.(*FastlyClient).conn
 
-	// Find the Service. Discard the service because we need the ServiceDetails,
-	// not just a Service record.
-	_, err := findService(d.Id(), meta)
-	if err != nil {
-		switch err {
-		case fastlyNoServiceFoundErr:
-			log.Printf("[WARN] %s for ID (%s)", err, d.Id())
-			d.SetId("")
-			return nil
-		default:
-			return err
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


thommahoney (approved):
This implementation looks correct to me. I confirmed once more than `GET /service/:id/details` returns a non-null `deleted_at` value for deleted services. It's great that we've cut down on the number of requests overall and the removal of `GET /service` is an obvious win. 👏

phamann:
Acceptance tests have passed locally.

PR #324, opened 2020-09-28
Changing ACL, Dictionary, Dynamic Snippets and WAF documentation

matpimenta:
Changing ACL, Dictionary and Dynamic Snippets documentation example to use [`for_each`](https://www.terraform.io/docs/configuration/resources.html#for_each-multiple-resource-instances-defined-by-a-map-or-set-of-strings) attributes instead of [_for expressions_](https://github.com/hashicorp/hcl/blob/hcl2/hclsyntax/spec.md#for-expressions).

In order to use _versionless_ resources (i.e. ACL Entries, Dictionary Items, Dynamic Snippets Content), it's required to link the ID generated from the Fastly Service (`fastly_service_v1`) to the respective field in the _versionless_ resource (i.e `fastly_service_acl_entries_v1`, `fastly_service_dictionary_items_v1`, `fastly_service_dynamic_snippet_content_v1`).

The original documentation instructed to the use of _for expressions_ to create this link. For example:

``` 
resource "fastly_service_v1" "myservice" {
  ...
  dictionary {
    name = "dict1"
  }

}
resource "fastly_service_dictionary_items_v1" "items" {
  service_id = fastly_service_v1.myservice.id
  dictionary_id = {for d in fastly_service_v1.myservice.dictionary : d.name => d.dictionary_id}["dict1"]
  ...
}
```

This seems to work for many scenarios, however, not for all. 
It fails mainly when the resource `fastly_service_v1` already exists and we add the `acl` block at the same time, we add `fastly_service_acl_entries_v1` resources (The same occur for the other _versionless_ resources).

The generated error is: `Error: Invalid index`

This seems to be due to the way Terraform evaluates *Computed* attributes inside `Sets`. When the resource doesn't exist, Terraform postpones the evaluation of the expression as the values are [not yet known](https://www.terraform.io/docs/configuration/expressions.html%23values-not-yet-known).
However, when the resource exists and the  Set attribute is being modified, Terraform tries to evaluate the expression before the all values are known causing the error. This is likely to be an issue in Terraform.

When replacing the _for expression_ with a `for_each` attribute, Terraform is able to resolve the values correctly. That is, Terraform detects that a value the computed id is still not available and postpones the expression evaluation. 

Interestingly enough, if using the _for expression_ with an `if` and `lookup`, the issue seems to also disappear. For example

```
lookup​({​for​ ​d​ in ​fastly_service_v1​.myservice.dictionary : ​d​.​name​ ​=>​ ​d​.dictionary_id ​if​ ​d​.​name == "dict1"​}, "dict1", ​""​)
```
This approach however is not as clean as the `for_each` one.

For all those reasons, the documentation was updated to recommend the `for_each` approach.

A similar issue also happens to `List` types and current affects the WAF resource. (This issue was previously raised against Terraform - see https://github.com/hashicorp/terraform/issues/23488)

For example:
```
resource "fastly_service_v1" "demo" {
  ...
  waf {
    response_object    = "WAF_Response"
  }

}

resource "fastly_service_waf_configuration" "waf" {
  waf_id = fastly_service_v1.demo.waf[0].waf_id
}
```

This will cause an error, when the `waf` block is added to an existing at the same at time that we add the `fastly_service_waf_configuration`

Unfortunately for case, there's currently no alternative approach to the index lookup. The `for_each` approach doesn't work correctly.

Therefore, WAF documentation was updated to include a warning about adding WAF to an existing service and a recommendation to do a two-phase approach in case this change is required.

Note:
For this PR, I've also tried to update the acceptance tests' HCL to reflect the `for_each` recommendation.
However, there’s apparently another issue in the Terraform Test framework that fails to destroy resources at the end of the test step if the HCL contains `for_each`. See related issue https://github.com/hashicorp/terraform-plugin-sdk/issues/264.

### DIFF
--- a/website_src/docs/r/service_acl_entries_v1.html.markdown.tmpl
+++ b/website_src/docs/r/service_acl_entries_v1.html.markdown.tmpl
@@ -15,7 +15,7 @@ Defines a set of Fastly ACL entries that can be used to populate a service ACL. 
 If Terraform is being used to populate the initial content of an ACL which you intend to manage via API or UI, then the lifecycle `ignore_changes` field can be used with the resource.  An example of this configuration is provided below.    
 
 
-## Example Usage
+## Example Usage (Terraform >= 0.12.6)
 
 Basic usage:
 
@@ -47,8 +47,11 @@ resource "fastly_service_v1" "myservice" {
 }
 
 resource "fastly_service_acl_entries_v1" "entries" {
+  for_each = {
+    for d in fastly_service_v1.myservice.acl : d.name => d if d.name == var.myacl_name
+  }
   service_id = fastly_service_v1.myservice.id
-  acl_id = {for d in fastly_service_v1.myservice.acl : d.name => d.acl_id}[var.myacl_name]
+  acl_id = each.value.acl_id
   entry {
     ip = "127.0.0.1"
     subnet = "24"
@@ -103,8 +106,11 @@ resource "fastly_service_v1" "myservice" {
 }
 
 resource "fastly_service_acl_entries_v1" "entries" {
+  for_each = {
+    for d in fastly_service_v1.myservice.acl : d.name => d if d.name == local.acl_name
+  }
   service_id = fastly_service_v1.myservice.id
-  acl_id     = { for d in fastly_service_v1.myservice.acl : d.name => d.acl_id }[local.acl_name]
+  acl_id = each.value.acl_id
   dynamic "entry" {
     for_each = [for e in local.acl_entries : {
       ip      = e.ip
@@ -121,6 +127,50 @@ resource "fastly_service_acl_entries_v1" "entries" {
 }
 ```
 
+## Example Usage (Terraform >= 0.12.0 && < 0.12.6)
+
+`for_each` attributes were not available in Terraform before 0.12.6, however, users can still use `for` expressions to achieve
+similar behaviour as seen in the example below.
+
+~> **Warning:** Terraform might not properly calculate implicit dependencies on computed attributes when using `for` expressions
+
+For scenarios such as adding an ACL to a service and at the same time, creating the ACL entries (`fastly_service_acl_entries_v1`)
+resource, Terraform will not calculate implicit dependencies correctly on `for` expressions. This will result in index lookup
+problems and the execution will fail.
+
+For those scenarios, it's recommended to split the changes into two distinct steps:
+
+1. Add the `acl` block to the `fastly_service_v1` and apply the changes
+2. Add the `fastly_service_acl_entries_v1` resource with the `for` expressions to the HCL and apply the changes
+
+Usage:
+
+```hcl
+variable "myacl_name" {
+	type = string
+	default = "My ACL"
+}
+
+resource "fastly_service_v1" "myservice" {
+  ...
+  acl {
+	name = var.myacl_name
+  }
+  ...
+}
+
+resource "fastly_service_acl_entries_v1" "entries" {
+  service_id = fastly_service_v1.myservice.id
+  acl_id = {for d in fastly_service_v1.myservice.acl : d.name => d.acl_id}[var.myacl_name]
+  entry {
+    ip = "127.0.0.1"
+    subnet = "24"
+    negated = false
+    comment = "ALC Entry 1"
+  }
+}
+```
+
 ### Supporting API and UI ACL updates with ignore_changes
 
 The following example demonstrates how the lifecycle `ignore_changes` field can be used to suppress updates against the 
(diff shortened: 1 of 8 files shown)


website_src/docs/r/service_acl_entries_v1.html.markdown.tmpl
```diff
@@ -121,6 +127,50 @@ resource "fastly_service_acl_entries_v1" "entries" {
...
 ```
 
+## Example Usage (Terraform >= 0.12.0 && < 0.12.6)
+
+`for_each` attributes were not available in Terraform before 0.12.6, however, users can still use `for` expressions to achieve
+similar behaviour as seen in the example below.
+
+~> **Warning:** Terraform might not properly calculate implicit dependencies on computed attributes when using `for` expressions
+
+For scenarios such as adding a ACL to a service and at the same time, creating the ACL entries (`fastly_service_acl_entries_v1`)
```
phamann:
```suggestion
For scenarios such as adding an ACL to a service and at the same time, creating the ACL entries (`fastly_service_acl_entries_v1`)
```

website_src/docs/r/service_acl_entries_v1.html.markdown.tmpl
```diff
@@ -121,6 +127,50 @@ resource "fastly_service_acl_entries_v1" "entries" {
...
+resource, Terraform will not calculate compute implicit dependencies correctly on `for` expressions. This will result in index lookup
```
phamann: I think this was a typo as "calculate compute" sounds weird, but it may have been intentional? I went with "calculate" to be consistent with the warning sentence above. 

```suggestion
resource, Terraform will not calculate implicit dependencies correctly on `for` expressions. This will result in index lookup
```

website_src/docs/r/service_waf_configuration.html.markdown.tmpl
```diff
@@ -472,6 +472,20 @@ output "rules" {
...
+When adding a `waf` to an existing `fastly_service_v1` and at the same time adding a `fastly_service_waf_configuration`
+resource with `waf_id = fastly_service_v1.demo.waf[0].waf_id` might result with the in the following error:
+
+> fastly_service_v1.demo.waf is empty list of object
+
+For this scenario, it's recommended to split the changes into two distinct steps:
+
+1. Add the `waf` block to the `fastly_service_v1` and apply the changes
+2. Add the `fastly_service_waf_configuration` to the HCL and apply the changes
+
```
phamann, line 488: 👌

website_src/docs/r/service_dictionary_items_v1.html.markdown.tmpl
```diff
@@ -133,15 +140,61 @@ resource "fastly_service_v1" "myservice" {
...
 
+## Example Usage (Terraform >= 0.12.0 && < 0.12.6)
+
+`for_each` attributes were not available in Terraform before 0.12.6, however, users can still use `for` expressions to achieve
+similar behaviour as seen in the example below.
+
+~> **Warning:** Terraform might not properly calculate implicit dependencies on computed attributes when using `for` expressions
+
+For scenarios such as adding a Dictionary to a service and at the same time, creating the Dictionary entries (`fastly_service_dictionary_items_v1`)
+resource, Terraform will not calculate compute implicit dependencies correctly on `for` expressions. This will result in index lookup
```
phamann: I think this was a typo as "calculate compute" sounds weird, but it may have been intentional? I went with "calculate" to be consistent with the warning sentence above. 

```suggestion
resource, Terraform will not calculate implicit dependencies correctly on `for` expressions. This will result in index lookup
```

website_src/docs/r/service_dynamic_snippet_content_v1.html.markdown.tmpl
```diff
@@ -59,51 +61,95 @@ Multiple dynamic snippets:
...
+
+## Example Usage (Terraform >= 0.12.0 && < 0.12.6)
+
+`for_each` attributes were not available in Terraform before 0.12.6, however, users can still use `for` expressions to achieve
+similar behaviour as seen in the example below.
+
+~> **Warning:** Terraform might not properly calculate implicit dependencies on computed attributes when using `for` expressions
+
+For scenarios such as adding a Dynamic Snippet to a service and at the same time, creating the Dynamic Snippets (`fastly_service_dynamic_snippet_content_v1`)
+resource, Terraform will not calculate compute implicit dependencies correctly on `for` expressions. This will result in index lookup
```
phamann: I think this was a typo as "calculate compute" sounds weird, but it may have been intentional? I went with "calculate" to be consistent with the warning sentence above. 

```suggestion
resource, Terraform will not calculate implicit dependencies correctly on `for` expressions. This will result in index lookup
```

phamann (changes requested):
Thank you for your research into this and contribution, @matpimenta. Couple of small things:

- Some suggested edits, mainly a typo across all three `s/compute/calculate/`
- Can you update the PR description to include more detail regarding the underlying issue we have found and why the `for_each` fixes it (i.e. your analysis). As I intend to link to this PR from some of the related issues and the detail will help those who find it in the future. 
- It would be good to see a test in at least one of the resources with some HCL containing two objects and the `for_each` pattern. Thoughts?

matpimenta:
@phamann No problem. I'll update the PR description with more details and change the acc tests. Makes sense to have it consistent with the documentation.

matpimenta:
@phamann I've fixed the changes and updated the PR description. I've added a note around `for_each` acc tests to the PR description.

PR #330, opened 2020-10-15
Guard WAF rule_exclusion read 

phamann:
### TL;DR
As part of the [v0.21.0](https://github.com/fastly/terraform-provider-fastly/releases/tag/v0.21.0) release we introduced the new `rule_exclusion` block to the [`fastly_service_waf_configuration`](https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_waf_configuration) resource which is causing errors for existing users of the resource - this PR attempts to fix the error with a temporary guard.

### Why?
The Fastly API backing this feature is still in "beta" and therefore behind a feature flag, users without the flag enabled on their account will see a 403 error when trying to plan/apply their `fastly_service_waf_configuration` due to the read phase always calling the UI. Until the feature moves to "limited availability" we have decided to roll-forwards and guard the read with a `d.GetOK()` check to validate the configuration includes the block before attempting to read. We have also updated the documentation to note the feature is still in beta.

### DIFF
--- a/fastly/resource_fastly_service_waf_configuration.go
+++ b/fastly/resource_fastly_service_waf_configuration.go
@@ -299,8 +299,14 @@ func resourceServiceWAFConfigurationV1Read(d *schema.ResourceData, meta interfac
 		return err
 	}
 
-	if err := readWAFRuleExclusions(meta, d, latestVersion.Number); err != nil {
-		return err
+	// As the rule exclusion API is still behind a beta feature
+	// flag, ensure we only read if the Set is non-empty.
+	//
+	// TODO(phamann): Remove d.GetOk() guard once in limited availability.
+	if _, ok := d.GetOk("rule_exclusion"); ok {
+		if err := readWAFRuleExclusions(meta, d, latestVersion.Number); err != nil {
+			return err
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


mccurdyc (approved):
LGTM!

You mentioned offline that you tested manually with a token that had and did not have access to this API.

Is there an automated way we can test this?

`d.GetOK()` makes sense per [the docs.](https://pkg.go.dev/github.com/hashicorp/terraform/helper/schema#ResourceData.GetOk)

> GetOk returns the data for the given key and whether or not the key has been set to a non-zero value at some point.
>
> The first result will not necessarilly be nil if the value doesn't exist. The second result should be checked to determine this information.

mccurdyc:
Are there any other places --- i.e., other than just READs --- where this issue would be encountered?

phamann:
> Are there any other places --- i.e., other than just READs --- where this issue would be encountered?

The only other place is the update path which is already guarded by a `d.HasChange("rule_exclusion")` which  like this fix ensures we only call the API if the configuration contains the block.

https://github.com/fastly/terraform-provider-fastly/blob/master/fastly/resource_fastly_service_waf_configuration.go#L256

PR #335, opened 2020-11-02
Correct IP range example

eheydrick:
Correct the example for `fastly_ip_ranges` datasource. `cidr_blocks` is expecting a list of strings but the example passes a list of a list of strings and that leads to this error:

`Inappropriate value for attribute "cidr_blocks": element 0: string required`

It works once I removed the extra list.


phamann (approved):
:+1: LGTM - thank you for your contribution.

PR #336, opened 2020-11-06
Fix link rendering on documentation index

jlav1n:
- cannot have a line break within Markdown link code; it will prevent rendering:
image

- tested correction using https://registry.terraform.io/tools/doc-preview 
image


phamann (approved):
👍 LGTM - Thank you for your contribution.

PR #338, opened 2020-11-14
Add headers for blocks definitions

chuhlomin:
This PR adds headers to blocks definitions, making navigation and referencing easier.


Integralist (approved):
LGTM. 

Thanks @chuhlomin for these updates!

I'll aim to merge this as soon as possible. We've some other changes that we're focused on, but once they're all done with I'll circle back to this PR.

PR #339, opened 2020-11-15
Remove go-version

chenrui333:


Integralist:
Hi @chenrui333 we were looking to deprecate this file (as we do not use it internally, as we already define the relevant supported go version within the CI configuration). 

Can you let me know what its/your use case is.

Thanks!

chenrui333:
No use case for me, just saw this file is out of sync.

chenrui333:
Ideally, it would be nice to have such file in the repo though (kind of metadata), but yeah, it does require some sort of maintenance work (which I am happy to help with)

Integralist:
We actually already have this information in the `go.mod` which is our preferred method of documenting it. Feel free to update your PR to remove the `.go-version` file, otherwise we'll go ahead and close this PR and handle the deletion ourselves. 

Thanks @chenrui333

chenrui333:
Sounds good, let's remove it.

Integralist (approved):

PR #341, opened 2020-11-18
Migrate to Go-Fastly 2.0.0

Integralist:
**Problem**: multiple lines of tagged releases for Go-Fastly is making adding new features via our various API clients much harder.
**Solution**: migrate terraform provider to using Go-Fastly 2.0.0


Integralist:
@phamann this is experimental and I've only run a single test but it's failing (see below for error) which might be simple to resolve but the solution is eluding me at the moment. If you or maybe @matpimenta (👋🏻) have a moment to review that would be appreciated...

```
$ make testacc TESTARGS='-run=TestAccFastlyServiceV1WAFUpdateResponse'

TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccFastlyServiceV1WAFUpdateResponse -timeout 360m -ldflags="-X=github.com/fastly/terraform-provider-fastly/version.ProviderVersion=acc"

=== RUN   TestAccFastlyServiceV1WAFUpdateResponse
=== PAUSE TestAccFastlyServiceV1WAFUpdateResponse
=== CONT  TestAccFastlyServiceV1WAFUpdateResponse
    testing.go:569: Step 1 error: errors during apply:
        
        Error: 400 - Bad Request:
        
            Title:  Bad request
            Detail: '' is not a valid hostname
        
          on /var/folders/zf/zlsh3t3n21ngmbzl_wr6ymm80000gn/T/tf-test059269954/main.tf line 2:
          (source code not available)
        
        
--- FAIL: TestAccFastlyServiceV1WAFUpdateResponse (85.09s)
FAIL
FAIL	github.com/fastly/terraform-provider-fastly/fastly	85.660s
?   	github.com/fastly/terraform-provider-fastly/scripts/website	[no test files]
FAIL
make: *** [testacc] Error 1
```

Unfortunately narrowing down where this error is coming from isn't very obvious to me at the moment 😬

fastly/block_fastly_service_v1_logging_cloudfiles.go
```diff
@@ -162,33 +162,33 @@ func (h *CloudfilesServiceAttributeHandler) buildCreate(cloudfilesMap interface{
...
+		AccessKey:         df["access_key"].(string),
+		PublicKey:         df["public_key"].(string),
+		GzipLevel:         uint(df["gzip_level"].(int)),
+		MessageType:       df["message_type"].(string),
+		Path:              df["path"].(string),
+		Region:            df["region"].(string),
+		Period:            uint(df["period"].(int)),
+		TimestampFormat:   df["timestamp_format"].(string),
+		Format:            vla.format,
+		FormatVersion:     uintOrDefault(vla.formatVersion),
```
Integralist, line 179: @phamann this function is called by multiple tests `TestAccFastlyServiceV1_logging_cloudfiles_basic` and `TestAccFastlyServiceV1_logging_cloudfiles_basic_compute` and to get the latter C@E test to pass I had to change the code to use `uintOrDefault` (as the fake terraform hcl code had omitted `format_version = 2` unlike the VCL equivalent which did set it, and also because there is a function `getVCLLoggingAttributes` which only sets `vla.formatVersion` if a VCL service type). 

I stumbled around on that failing test for a while and so I'd like you to give some extra attention to this change just so we can be sure it's correct. 

My concern is that if `vla.formatVersion` is not set when a C@E service, then it'll be `nil` and in the case of the pre-exising `uintOrDefault()` function, the 'default value' it returns is `0` but `FormatVersion` _should_ have a default value of `2` as its default (as per the api docs that state it refers to the apache log format version 2, where the actual acceptable values are 1 or 2, not zero!). 

I was writing my own code to check for nil and to set it to a default of `2` but then I noticed some other code also using `uintOrDefault()` when dealing with `FormatVersion` 🤔

Integralist: examples using `uintOrDefault()` with `FormatVersion`: https://github.com/fastly/terraform-provider-fastly/commit/7844ddb0161ff5518a5e0bc67a51b8261e903793 

Maybe there are no 'compute' tests that validate those particular logging blocks 🤔

Integralist: Patrick's response (for posterity):

> The `format_version` field is completely irrelevant for C@E services (as there is no VCL to format) so it doesn’t matter if the value is 0, 2 or nil for those services and hence why the C@E code paths don’t set or test for its existence.

Integralist:
@phamann this is ready for review, and I can confirm that the full test was run (3hrs) and it passed successfully.

GNUmakefile
```diff
@@ -5,6 +5,7 @@ PKG_NAME=fastly
 FULL_PKG_NAME=github.com/fastly/terraform-provider-fastly
 VERSION_PLACEHOLDER=version.ProviderVersion
 VERSION=$(shell git describe --tags --always)
+TF_LOG ?= "ERROR"
```
phamann: Do we really want `TF_LOG` in our tests to default to `ERROR`?

fastly/block_fastly_service_v1_backend.go
```diff
@@ -74,8 +74,15 @@ func (h *BackendServiceAttributeHandler) Process(d *schema.ResourceData, latestV
...
-		opts := h.buildUpdateBackendInput(d.Id(), latestVersion, df)
+		backend := dRaw.(map[string]interface{})
+
+		// Each element in the .Modified slice is a backend with nested fields.
+		// Although we know the backend has been modified, we have to manually
+		// filter out any unmodified nested fields by comparing them against the
+		// original backend set data structure using .Filter.
+		//
+		modified := setDiff.Filter(backend, obs)
+		opts := h.buildUpdateBackendInput(d.Id(), latestVersion, backend, modified)
```
phamann, line 85: Nice. Much cleaner that how we left it in the pairing session 👌🏻

fastly/block_fastly_service_v1_waf.go
```diff
@@ -44,6 +44,12 @@ func (h *WAFServiceAttributeHandler) Register(s *schema.Resource) error {
 					Computed:    true,
 					Description: "The Web Application Firewall (WAF) ID",
 				},
+				"disabled": {
+					Type:        schema.TypeBool,
+					Optional:    true,
+					Default:     false,
+					Description: "A flag used to completely disable a Web Application Firewall. This is intended to be used as an emergency.",
```
phamann: I know not originally your message, but we may as well clean it up. I think this reads better.

```suggestion
					Description: "A flag used to completely disable a Web Application Firewall. This is intended to only be used in an emergency.",
```

fastly/block_fastly_service_v1_waf.go
```diff
@@ -56,12 +62,12 @@ func (h *WAFServiceAttributeHandler) Process(d *schema.ResourceData, latestVersi
 	serviceVersion := strconv.Itoa(latestVersion)
```
phamann: The `serviceVersion` is cast to a string here and then cast back to an int in all of the CRUD methods below. Maybe we should change this to avoid having to cast back?

fastly/block_fastly_service_v1_waf.go
```diff
@@ -148,9 +155,12 @@ func flattenWAFs(wafList []*gofastly.WAF) []map[string]interface{} {
 
 func buildCreateWAF(WAFMap interface{}, serviceID string, ServiceVersion string) *gofastly.CreateWAFInput {
 	df := WAFMap.(map[string]interface{})
+
+	version, _ := strconv.Atoi(ServiceVersion)
```
phamann: See previous comment about the source of this variable and avoid having to cast. Similar comments in this file elided.

fastly/block_fastly_service_v1_waf.go
```diff
@@ -160,21 +170,40 @@ func buildCreateWAF(WAFMap interface{}, serviceID string, ServiceVersion string)
...
+	input := gofastly.UpdateWAFInput{
+		ServiceID:      gofastly.String(serviceID),
+		ServiceVersion: gofastly.Int(version),
+		ID:             df["waf_id"].(string),
 	}
-	return &opts
+
+	// NOTE: to access the WAF data we need to link to specific list index. This
+	// isn't ideal and if we already know that a WAF can only have MaxItems: 1,
+	// then we should probably move away from it being a schema.TypeList.
```
phamann: Nice. Maybe we should also add a `TODO` comment above the schema definition at the top of this file. https://github.com/fastly/terraform-provider-fastly/pull/341/files?w=1#diff-d24aa2ccdd9266a29f76375a50d26d747382113c3fcbd6738dcf8bd68e3bd835R27 so it can be found in the future?

fastly/block_fastly_service_v1_waf_test.go
```diff
@@ -251,6 +252,21 @@ resource "fastly_service_v1" "foo" {
...
   %s
 
+	# WAF was updated to insert an ALWAYS_FALSE default condition. If we didn't
+	# define the condition, then when terraform refreshes the state after the test
+	# it would error because the remote state would have this extra condition.
+	#
+	# If the WAF isn't in place and without that ALWAYS_FALSE condition, the WAF 
+	# response object (403) will be served for all inbound traffic. This 
+	# condition was added by the WAF team to prevent Fastly from returning a 403 
+	# on all of the customer traffic before WAF is provisioned to the service.
```
phamann: Maybe consider re-wording this to mention the related bug in the WAF API and a `TODO` comment or similar to change the `ALWAYS_FALSE` value to another token when fixed.

fastly/convert.go
```diff
@@ -1,20 +1,5 @@
...
-// intToPtr returns a pointer to the passed int.
-func intToPtr(i int) *int {
-	return &i
-}
-
-// boolToPtr returns a pointer to the passed bool.
-func boolToPtr(i bool) *bool {
-	return &i
-}
-
```
phamann, line 17: WOOHOO 🎉

fastly/diff.go
```diff
@@ -105,6 +106,27 @@ func (h *SetDiff) computeKey(elem interface{}) (interface{}, error) {
 	return key, nil
 }
 
+// Filter filters out unmodified fields of a map data structure by ranging over
```
phamann:
```suggestion
// Filter filters out unmodified fields of a Set elements map data structure by ranging over
```

fastly/diff.go
```diff
@@ -105,6 +106,27 @@ func (h *SetDiff) computeKey(elem interface{}) (interface{}, error) {
...
+			for k, v := range m {
+				if v != modified[k] {
+					filtered[k] = modified[k]
+				}
+			}
+		}
+	}
+
+	return filtered
+}
```
phamann, line 128: Beautiful 👨🏻‍🍳

fastly/resource_fastly_service_dictionary_items_v1_test.go
```diff
@@ -74,6 +74,16 @@ func TestAccFastlyServiceDictionaryItemV1_create(t *testing.T) {
...
+// TestAccFastlyServiceDictionaryItemV1_create_inactive_service validates that
+// when creating a new inactive service consisting of a dictionary along with a
+// predefined list of items to populate it with, are applied successfully
+// instead of generating an error suggesting the dictionary ID was missing.
+//
+// NOTE: This error stemmed from a bug in our code (#345) where we discovered
+// that if a configuration has the activate field set to false AND it has no
+// previous active version, then the state wasn't being read. This manifested
+// itself as a runtime error in certain situations, such as another resource
+// referencing the state in its configuration.
```
phamann, line 86: Thank you 🙌🏻

fastly/resource_fastly_service_waf_configuration_test.go
```diff
@@ -406,6 +406,21 @@ resource "fastly_service_v1" "foo" {
...
   }
 
+	# WAF was updated to insert an ALWAYS_FALSE default condition. If we didn't
+	# define the condition, then when terraform refreshes the state after the test
+	# it would error because the remote state would have this extra condition.
+	#
+	# If the WAF isn't in place and without that ALWAYS_FALSE condition, the WAF 
+	# response object (403) will be served for all inbound traffic. This 
+	# condition was added by the WAF team to prevent Fastly from returning a 403 
+	# on all of the customer traffic before WAF is provisioned to the service.
```
phamann: See my previous comment regarding updating this comment.

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/ajg/form v0.0.0-20160822230020-523a5da1a92f // indirect
-	github.com/fastly/go-fastly/v2 v2.0.0-alpha.3
+	github.com/fastly/go-fastly/v2 v2.0.0
```
phamann, line 7: WOOHOO!

phamann (changes requested):
This is looking great; some minor comments but nothing really blocking or major. Many are comments of joy. 

I know first-hand how painful this was to get over the line and that its especially not much of a nice task to do as your first at Fastly, so I want personally say THANK YOU for your dedication and patience. This is going to make developing and maintaining the provider much easier in the future. ❤️

phamann:
Not for this PR, however it became very apparent while reviewing that we're not really benefiting from the pointer pattern in v2 that much in the provider (and thus still exposing a potential class of bug), as most of the top-level blocks need to be refactored to use the `schema.TypeSet` diff pattern we use in [`backend`](https://github.com/fastly/terraform-provider-fastly/blob/4ef7a34b34c6a056602e8a20981359e7d237eb81/fastly/block_fastly_service_v1_backend.go), allowing us to use `gofastly`UpdateFoo` -which uses pointers - instead of deleting and creating resources every-time.

GNUmakefile
Integralist: I thought it was the implicit default value but seems I was mistaken, so I'll remove it.

fastly/block_fastly_service_v1_waf.go
Integralist: Great spot! I'll fix this.

fastly/block_fastly_service_v1_waf.go
Integralist: Cool. Will add 👍🏻

Integralist: hmm, interesting thought: what would the schema be set to? 

I had presumed we'd move from `TypeList` to `TypeMap`, but according to the docs that would be `map[string]interface{}` and also, where the current TypeList has the `Elem` field set to a `schema.Resource` (whose own Schema field is `map[string]*schema.Schema`), the documentation suggests...

> For a TypeMap, Elem must be a *Schema with a Type that is one of the primitives: TypeString, TypeBool, TypeInt, or TypeFloat.

Which sounds like you can then only have one 'type' as the value for each key? Which is also a bit confusing because the `TypeMap` should be `map[string]interface{}` from what I can tell 🤔

Integralist: I might just reword the comment instead so it's clearer why we use the weird dot notation syntax, rather than state we should move away from the TypeList (because right now it might be 'by design')

phamann (approved):
Looks good to me! 🎉 

ship-that-ships-ships

PR #342, opened 2020-11-19
Add support for kafka endpoints with sasl options

kellymclaughlin:
Add support for the following fields in order to support SASL for Kafka logging endpoints:

- parse_log_keyvals
- auth_method
- request_max_bytes
- user
- password

## Testing

```
$ env TF_ACC=1 make testacc TESTARGS='-v -run=[Kk]afka'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -v -run=[Kk]afka -timeout 360m -ldflags="-X=github.com/fastly/terraform-provider-fastly/version.ProviderVersion=acc"
=== RUN   TestResourceFastlyFlattenKafka
--- PASS: TestResourceFastlyFlattenKafka (0.00s)
=== RUN   TestAccFastlyServiceV1_kafkalogging_basic
--- PASS: TestAccFastlyServiceV1_kafkalogging_basic (105.85s)
=== RUN   TestAccFastlyServiceV1_kafkalogging_basic_compute
--- PASS: TestAccFastlyServiceV1_kafkalogging_basic_compute (46.12s)
PASS
ok  	github.com/fastly/terraform-provider-fastly/fastly	151.993s
?   	github.com/fastly/terraform-provider-fastly/scripts/website	[no test files]
```

### DIFF
--- a/fastly/block_fastly_service_v1_logging_kafka.go
+++ b/fastly/block_fastly_service_v1_logging_kafka.go
@@ -53,7 +53,7 @@ func (h *KafkaServiceAttributeHandler) Register(s *schema.Resource) error {
 		"required_acks": {
 			Type:     schema.TypeString,
 			Optional: true,
-			Description: "The Number of acknowledgements blockAttributes leader must receive before blockAttributes write is considered successful. One of: 1 (default) One server needs to respond. 0 No servers need to respond. -1	Wait for all in-sync replicas to respond.",
+			Description: "The Number of acknowledgements a leader must receive before a write is considered successful. One of: 1 (default) One server needs to respond. 0 No servers need to respond. -1	Wait for all in-sync replicas to respond.",
 		},
 
 		"use_tls": {
@@ -93,38 +93,38 @@ func (h *KafkaServiceAttributeHandler) Register(s *schema.Resource) error {
 		"tls_hostname": {
 			Type:        schema.TypeString,
 			Optional:    true,
-			Description: "The hostname used to verify the server's certificate. It can either be the Common Name or blockAttributes Subject Alternative Name (SAN).",
+			Description: "The hostname used to verify the server's certificate. It can either be the Common Name or a Subject Alternative Name (SAN).",
 		},
 
 		"parse_log_keyvals": {
 			Type:        schema.TypeBool,
 			Optional:    true,
 			Default:     false,
-			Description: "Parse key-value pairs within the log formatWhether to use TLS for secure logging.",
+			Description: "Enables parsing of key=value tuples from the beginning of a logline, turning them into record headers.",
 		},
 
-		"max_batch_size": {
+		"request_max_bytes": {
 			Type:        schema.TypeInt,
 			Optional:    true,
-			Description: "The maximum size of the log batch in bytes.",
+			Description: "Maximum size of log batch, if non-zero. Defaults to 0 for unbounded.",
 		},
 
 		"auth_method": {
 			Type:        schema.TypeString,
 			Optional:    true,
-			Description: "SASL authentication method. Valid values are: plain, scram-sha-256, scram-sha-512.",
+			Description: "SASL authentication method. One of: plain, scram-sha-256, scram-sha-512.",
 		},
 
 		"username": {
 			Type:        schema.TypeString,
 			Optional:    true,
-			Description: "SASL authentication username",
+			Description: "SASL User.",
 		},
 
 		"password": {
 			Type:        schema.TypeString,
 			Optional:    true,
-			Description: "SASL authentication password",
+			Description: "SASL Pass.",
 		},
 	}
 
@@ -283,7 +283,7 @@ func flattenKafka(kafkaList []*gofastly.Kafka) []map[string]interface{} {
 			"placement":          s.Placement,
 			"response_condition": s.ResponseCondition,
 			"parse_log_keyvals":  s.ParseLogKeyvals,
-			"max_batch_size":     s.RequestMaxBytes,
+			"request_max_bytes":  s.RequestMaxBytes,
 			"auth_method":        s.AuthMethod,
 			"username":           s.User,
 			"password":           s.Password,
@@ -324,7 +324,7 @@ func (h *KafkaServiceAttributeHandler) buildCreate(kafkaMap interface{}, service
 		Placement:         gofastly.NullString(vla.placement),
 		ResponseCondition: gofastly.NullString(vla.responseCondition),
 		ParseLogKeyvals:   fastly.CBool(df["parse_log_keyvals"].(bool)),
-		RequestMaxBytes:   fastly.Uint(uint(df["max_batch_size"].(int))),
+		RequestMaxBytes:   fastly.Uint(uint(df["request_max_bytes"].(int))),
 		AuthMethod:        fastly.NullString(df["auth_method"].(string)),
 		User:              fastly.NullString(df["username"].(string)),
 		Password:          fastly.NullString(df["password"].(string)),
--- a/fastly/block_fastly_service_v1_logging_kafka.go
+++ b/fastly/block_fastly_service_v1_logging_kafka.go
@@ -115,7 +115,7 @@ func (h *KafkaServiceAttributeHandler) Register(s *schema.Resource) error {
 			Description: "SASL authentication method. One of: plain, scram-sha-256, scram-sha-512.",
 		},
 
-		"username": {
+		"user": {
 			Type:        schema.TypeString,
 			Optional:    true,
 			Description: "SASL User.",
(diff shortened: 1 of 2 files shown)


fastly/block_fastly_service_v1_logging_kafka.go
```diff
@@ -95,6 +95,37 @@ func (h *KafkaServiceAttributeHandler) Register(s *schema.Resource) error {
 			Optional:    true,
 			Description: "The hostname used to verify the server's certificate. It can either be the Common Name or blockAttributes Subject Alternative Name (SAN).",
```
Integralist: While checking the convention of whether the descriptions match exactly the api docs or if they're re-written to be more concise/custom to TF usage I then noticed what I imagine is just a weird copy/paste irregularity.

Example...

```
blockAttributes
```

I'm not sure why `blockAttributes` is in the description when comparing the description to what's written in the [api documentation](https://developer.fastly.com/reference/api/logging/kafka/).

I appreciate it's not part of your change set, but if we're going to be publicly announcing Kafka as GA, then it would be good to ensure all docs are up to par.

Hope that's OK.

fastly/block_fastly_service_v1_logging_kafka.go
```diff
@@ -95,6 +95,37 @@ func (h *KafkaServiceAttributeHandler) Register(s *schema.Resource) error {
...
 		},
+
+		"parse_log_keyvals": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
+			Description: "Parse key-value pairs within the log formatWhether to use TLS for secure logging.",
```
Integralist: The other pre-existing descriptions appear to be aligning with the api documentation, is there a reason to break from that convention here? 

> **UPDATE**: I can see it makes sense to rewrite the descriptions when I look at `auth_method` as that would be much more verbose otherwise 👍🏻 

Although there does appear to be a typo in this description still (missing punctuation)...

```
formatWhether 
```

Also the end of the description (e.g. `Whether to use TLS for secure logging`) appears to be copied from a completely different field (i.e. the `use_tls` field).

I think it would be best to keep with the description as per the [api documentation](https://developer.fastly.com/reference/api/logging/kafka/) in this case.

fastly/block_fastly_service_v1_logging_kafka.go
```diff
@@ -95,6 +95,37 @@ func (h *KafkaServiceAttributeHandler) Register(s *schema.Resource) error {
...
+		},
+
+		"max_batch_size": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Description: "The maximum size of the log batch in bytes.",
```
Integralist: Do we need to mention `Defaults to 0 for unbounded` (like it does in the api documentation) or configure a `Default` field of zero?

fastly/block_fastly_service_v1_logging_kafka_test.go
```diff
@@ -53,6 +58,11 @@ func TestResourceFastlyFlattenKafka(t *testing.T) {
 					"format":             `%a %l %u %t %m %U%q %H %>s %b %T`,
 					"placement":          "none",
 					"format_version":     uint(2),
+					"parse_log_keyvals":  true,
+					"max_batch_size":     uint(12345),
```
Integralist: When constructing the `gofastly.Kafka` above we use the untyped values `2` and `12345` while here we then wrap them in a `uint` is that necessary?

Integralist (changes requested):
Hi @kellymclaughlin thanks for the PR! 

I have a couple of minor comments (mostly documentation related) and one comment that hopefully just requires a little clarity (for my benefit).

Also, it would be good in future to have a confirmation that the _full_ test suite passed successfully and not just the new feature added. I appreciate that in practice it's unlikely this isolated change would cause a break in behaviour elsewhere, but I'm a bit OCD like that 😬 

With that in mind I've pulled your repo/branch locally and am currently running the full test suite. I'll report back with the results once it has completed.

fastly/block_fastly_service_v1_logging_kafka.go
kellymclaughlin: :+1: I hadn't even noticed that, but will definitely fix it up.

fastly/block_fastly_service_v1_logging_kafka.go
kellymclaughlin: Nope, no reason. When I wrote the descriptions the API docs weren't done yet so I'll update them all to match the docs.

kellymclaughlin:
> Also, it would be good in future to have a confirmation that the full test suite passed successfully and not just the new feature added. I appreciate that in practice it's unlikely this isolated change would cause a break in behaviour elsewhere, but I'm a bit OCD like that :grimacing:

A question for you on that, in the repo README it mentions that running the full acceptance suite can take many hours and cost money so I was hesitant to run the entire thing. Do you have a feel for how much a full run costs and is the amount such that it's justifiable to run as part of a development workflow or should we offload that to a CI step to control costs? This is my first time working with the provider so I'm learning as I go. Thanks!

fastly/block_fastly_service_v1_logging_kafka_test.go
kellymclaughlin: Good question. This is necessary because the `FormatVersion` and `RequestMaxBytes` fields in the`Kafka` type in `go-fastly` are `uint` types and they are implicitly converted to `uint` when the struct is created. The casting is needed at the point of this comment because there is no type expectation and the values will default to `int` types and this causes the test comparison between `remote` and `local` to fail.

fastly/block_fastly_service_v1_logging_kafka.go
kellymclaughlin: Done

fastly/block_fastly_service_v1_logging_kafka.go
kellymclaughlin: Done

fastly/block_fastly_service_v1_logging_kafka.go
kellymclaughlin: Done

Integralist:
👋🏻 @kellymclaughlin I've been on paternity leave so I'm going to try and get round to re-reviewing this today.

fastly/block_fastly_service_v1_logging_kafka.go
```diff
@@ -93,7 +93,38 @@ func (h *KafkaServiceAttributeHandler) Register(s *schema.Resource) error {
...
+			Description: "Maximum size of log batch, if non-zero. Defaults to 0 for unbounded.",
+		},
+
+		"auth_method": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "SASL authentication method. One of: plain, scram-sha-256, scram-sha-512.",
+		},
+
+		"username": {
```
Integralist: This doesn't align with either the API documentation nor the Go-Fastly response struct tag, which has it defined as `user`.

Integralist (commented):
Thanks Kelly. Just one minor comment.

Integralist:
@kellymclaughlin re:

> in the repo README it mentions that running the full acceptance suite can take many hours and cost money so I was hesitant to run the entire thing. Do you have a feel for how much a full run costs and is the amount such that it's justifiable to run as part of a development workflow or should we offload that to a CI step to control costs?

So it has been suggested to me in the past (by Patrick) that we should run the full test suite once before merging the PR just so that we can be sure the changes haven't broken anything else.

I don't think we could push this to a CI check because each commit would require it to run and that wouldn't be practical due to the time it takes to run (~1.5hrs I think 😬 -- I'm not a fan of 'end-to-end' tests for this very reason).

I don't think the cost of running the full test suite should be a concern once the PR has been reviewed and it's at the point of being merged, as that's our last-ditch effort to have confidence in the change-set overall.

fastly/block_fastly_service_v1_logging_kafka.go
kellymclaughlin: :+1: Done

Integralist (approved):

Integralist:
@kellymclaughlin this has been approved, but before I merge could you run the full test suite please.

I would do this myself but I'm in the middle of migrating terraform to Go-Fastly 2.0.0 and so my local setup is all over the place 😬

Integralist:
@kellymclaughlin 👋🏻 I can see there are some conflicts with this file. Are you OK to rebase master and resolve the conflicts (+ run the full test suite). Thanks!

kellymclaughlin:
@Integralist  I finally got a clean run of the full test suite.

```
18:30:33:terraform-provider-fastly(kelly/add-kafka-sasl-fields) $ make testacc
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v  -timeout 360m -ldflags="-X=github.com/fastly/terraform-provider-fastly/version.ProviderVersion=acc"

...

PASS
ok  	github.com/fastly/terraform-provider-fastly/fastly	7053.460s
?   	github.com/fastly/terraform-provider-fastly/scripts/website	[no test files]
```

PR #343, opened 2020-11-26
fix typo in service_compute doc

cookie-s:
should be past participle


phamann (approved):
👍 Good catch - thank you for the fix.

PR #349, opened 2021-01-04
RELEASE.md

Integralist:
Create a release process document for future contributors.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,6 +10,8 @@
 
 ## Notes
 
+Step 3. we utilize [semantic versioning](https://semver.org/) and only include relevant/significant changes within the CHANGELOG.
+
 Step 5. causes a [github action](https://github.com/fastly/terraform-provider-fastly/blob/master/.github/workflows/release.yml) to be triggered which produces a 'draft' release.
 
 Step 7. causes a [github webhook](https://github.com/fastly/terraform-provider-fastly/settings/hooks) to be triggered which produces a release on the [terraform registry](https://registry.terraform.io/providers/fastly/fastly/latest) and which can take a while to publish, so check back later if it doesn't show up within a few minutes.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,14 @@
+# Release Process
+
+1. Merge PR.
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,14 @@
+# Release Process
+
+1. Merge PR.
```
phamann, line 3: Either here after step one or after step 3, it would be good to suggest the user runs the full acceptance test suite against the master branch.

RELEASE.md
```diff
@@ -0,0 +1,14 @@
...
+2. Open a new PR to update CHANGELOG ([example](https://github.com/fastly/terraform-provider-fastly/pull/348)).
```
phamann: Would be good to seem some additional notes here about our changelog conventions. I.e. we follow semver and only include relevant/significant changes in the log etc.

RELEASE.md
```diff
@@ -0,0 +1,14 @@
...
+3. Rebase latest remote master branch locally (`git pull --rebase origin master`).
+4. Tag a new release (`git tag -s vX.Y.Z -m "vX.Y.Z" && git push origin vX.Y.Z`).
+5. Copy/paste CHANGELOG into the [draft release](https://github.com/fastly/terraform-provider-fastly/releases).
+6. Publish draft release.
```
phamann: I suggest we add a final step about communicating the new release via internal channels.

phamann (changes requested):
Looking good. A couple of comments to ensure we follow the normal flow.

RELEASE.md
phamann: I'd take a look at the CLI or Go Fastly docs for inspiration: https://github.com/fastly/go-fastly/blob/master/RELEASE.md#how-to-cut-a-new-release-for-the-go-fastly-http-client

RELEASE.md
Integralist: Ah yes good shout!

RELEASE.md
Integralist: Good shout 👍🏻

Integralist:
@phamann made changes as requested

phamann (approved):
👍🏻 LGTM

PR #356, opened 2021-01-25
Replace old doc generation process with new tfplugindocs tool

Integralist:
Hashicorp have deleted a Makefile target (`website-provider-test`) which we were depending on to generate our [Terraform provider's documentation](https://registry.terraform.io/providers/fastly/fastly/latest/docs) (this was historically handled by the CI action file `.github/workflows/pr.yml` and the `website` job defined there). This means all our Terraform provider PRs will be failing (inc. any new PRs that get opened).

**NOTES**: I need to restructure the directories for documentation in order to use https://github.com/hashicorp/terraform-plugin-docs. I also removed some obsolete files that were related to when the terraform platform was hosted using Ruby.

Because the new `tfplugindocs` tool is using the `Description` field from our own Schema code to generate the schema documentation output, it meant that I needed to manually copy the live documentation values into the code because there was a difference (in a lot of places) between the code and the original _manually_ curated/written documentation.

Contrived example:

```diff
- "service_id": "service Id"            // original code
+ "service_id": "The ID of the Service" // description in the manually written docs.
```

> **NOTE**: the biggest change in this PR (outside of renaming directories and files) is `scripts/generate-docs.go` which is what now generates the documentation.

**CAVEATS**: I _believe_ the original CI action was validating the content (e.g. broken links). The new tool, although it has a `validate` command, appears to only validate the _directory structure_ and not the content. 

This means I've deleted the CI `website` job and the related Makefile targets because they are no longer usable for our purposes. Hopefully in the future the `tfplugindocs validate` command will become _richer_ in functionality and so we can try running it in CI.

**FOLLOW-UP PR**: implement a new `.github/workflows/pr.yml` job that will run `make generate-docs` and then diff the output against what's been changed to be sure any changes to the code results in new documentation being generated.

**REFERENCES**: https://www.terraform.io/docs/registry/providers/docs.html


README.md
```diff
@@ -83,21 +74,27 @@ In order to run the tests with extra debugging context, prefix the `make` comman
 $ TF_LOG=trace make testacc
 ```
 
-Building The Documentation
---------------------------
+## Building The Documentation
+
+The documentation is built from components (go templates) stored in the `docs_src` folder.
```
phamann: The `tfplugindocs` convention is to use a directory named `templates` instead of our old `docs_src`. Is there a reason why you've kept that and not followed convention?

https://github.com/hashicorp/terraform-plugin-docs#conventional-paths

Integralist:
@phamann looks like I may have made some progress 🎉 

```bash
$ tfplugindocs validate
detected static docs directory, running checks
                                                                                                                                                                                                                                              
$ tfplugindocs generate
rendering website for provider "terraform-provider-fastly"
exporting schema from Terraform
compiling provider "fastly"
generating missing resource content
generating template for "fastly_service_dictionary_items_v1"
generating template for "fastly_service_dynamic_snippet_content_v1"
generating template for "fastly_service_v1"
generating template for "fastly_service_waf_configuration"
generating template for "fastly_user_v1"
generating template for "fastly_service_acl_entries_v1"
generating template for "fastly_service_compute"
generating missing data source content
generating template for "fastly_ip_ranges"
generating template for "fastly_waf_rules"
generating missing provider content
generating template for "terraform-provider-fastly"
cleaning rendered website dir
rendering templated website to static markdown
rendering "data-sources/ip_ranges.md.tmpl"
rendering "data-sources/waf_rules.md.tmpl"
rendering "index.md.tmpl"
rendering "resources/service_acl_entries_v1.md.tmpl"
rendering "resources/service_compute.md.tmpl"
rendering "resources/service_dictionary_items_v1.md.tmpl"
rendering "resources/service_dynamic_snippet_content_v1.md.tmpl"
rendering "resources/service_v1.md.tmpl"
rendering "resources/service_waf_configuration.md.tmpl"
rendering "resources/user_v1.md.tmpl"
```

See the changes to the generated docs in the latest commit (00f2482)

This suggests two things:

1. We don't need `scripts/website/parse-templates.go` as `tfplugindocs generates` generates the required Markdown.
2. The new Markdown is significantly different in the sense that we lose contextual information...

```diff
- cidr_blocks - The lexically ordered list of ipv4 CIDR blocks.
+ cidr_blocks (List of String)
```

Another example is...

<img width="319" alt="Screenshot 2021-01-26 at 11 39 07" src="https://user-images.githubusercontent.com/180050/105840604-2e04d300-5fcb-11eb-95bd-7c15022b4a46.png">

But if I check the Terraform code I can see there are descriptions defined...

<img width="726" alt="Screenshot 2021-01-26 at 11 39 18" src="https://user-images.githubusercontent.com/180050/105840622-36f5a480-5fcb-11eb-91f0-f471591afb07.png">

So we need to figure out how can we get that context put _back_ in (as we're losing a lot of useful info otherwise) 🤔 

Two other outstanding questions are:

1. How do we _preview_ these changes? Looks like there is now only a 'basic' preview tool ([here](https://registry.terraform.io/tools/doc-preview)).
2. What other features are missing from the original `website-provider-test` Makefile target? e.g. I thought we had link checkers etc.

/cc @kailan @doramatadora

README.md
```diff
@@ -83,21 +74,27 @@ In order to run the tests with extra debugging context, prefix the `make` comman
...
-The documentation is built from components (go templates) stored in the `website_src` folder.
-Building the documentation copies the full markdown into the `website` folder, ready for deployment to Hashicorp.
+> NOTE: you'll need the [`tfplugindocs`](https://github.com/hashicorp/terraform-plugin-docs) tool for generating HTML from the Markdown to be deployed to Hashicorp.
 
 With the repository cloned to: `$GOPATH/src/github.com/fastly/terraform-provider-fastly`:
 
-* To build the documentation:
-`go run scripts/website/parse-templates.go `
+* To build the documentation Markdown files:
+`go run scripts/website/parse-templates.go`
```
phamann: Is this step needed? I thought `tfplugindocs generate` now handles everything for us?

phamann (changes requested):
Looking good, I think there is some confusion as to what the tool is doing, hence why you're seeing large diffs in the content.

>The new Markdown is significantly different in the sense that we lose contextual information...

I believe this is because the tool is not finding our templates, as you're using `docs_src` directory instead of `templates` and thus it is generating its own ones from the provider schema. From the [how it works](https://github.com/hashicorp/terraform-plugin-docs#how-it-works) read me of the tool: 

- _Generate resource template files, if missing_

>How do we preview these changes? Looks like there is now only a 'basic' preview tool (here).

I'm pretty sure serving is on the roadmap for the new too. In fact you can see a stub [here](https://github.com/hashicorp/terraform-plugin-docs/blob/main/internal/cmd/serve.go). Therefore I suggest we just wait for that and live without it in the meantime.

> I thought we had link checkers etc.

Per above, I assume this feature will come in a later version of the tool.

docs/data-sources/ip_ranges.md
```diff
@@ -1,9 +1,5 @@
 ---
-layout: "fastly"
 page_title: "Fastly: fastly_ip_ranges"
-sidebar_current: "docs-fastly-datasource-ip_ranges"
-description: |-
```
Integralist: The new frontmatter doesn't support these legacy fields:
https://www.terraform.io/docs/registry/providers/docs.html#yaml-frontmatter

docs/data-sources/waf_rules.md
```diff
@@ -145,4 +141,27 @@ The `rules` block supports:
...
+### Read-only
+
+- **rules** (List of Object) The list of rules that results from any given combination of filters. (see [below for nested schema](#nestedatt--rules))
+
+<a id="nestedatt--rules"></a>
+### Nested Schema for `rules`
+
+Read-only:
+
+- **latest_revision_number** (Number)
```
Integralist, line 171: There are a few issues open already with the `tfplugindocs` tool only just being release. I've opened an issue related to nested Schema 'Description' fields not being included in the output.

https://github.com/hashicorp/terraform-plugin-docs/issues/28

docs/data-sources/ip_ranges.md
```diff
@@ -34,3 +30,14 @@ resource "aws_security_group" "from_fastly" {
 * `ipv6_cidr_blocks` - The lexically ordered list of ipv6 CIDR blocks.
 
 [1]: https://docs.fastly.com/guides/securing-communications/accessing-fastlys-ip-ranges
+<!-- schema generated by tfplugindocs -->
```
Integralist, line 32: There is currently some overlap between our own curated "Attributes Reference" and the "Schema" section generated dynamically by the new `tfplugindocs` tool, but until the issue (hashicorp/terraform-plugin-docs#28) is resolved we might want to hold off making any further content changes.

docs/resources/service_compute.md
```diff
@@ -556,3 +552,562 @@ Fastly Service can be imported using their service ID, e.g.
 ```
 $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx
 ```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
```
Integralist, line 69: @phamann this is actually an example of where the generated Schema works quite well and so we might not want to keep our own curated "Arguments Reference".

Where things get complicated is when the tool fails to pull in the `Description` field values (which happens in a few files, and the behaviour is inconsistent as noted in [this issue I reported](https://github.com/hashicorp/terraform-plugin-docs/issues/28)).

Integralist: Also, it gets awkward if there is ever a scenario where we keep the auto-generated Schema section but we still want to add in some additional 'contextual' information.

docs/data-sources/ip_ranges.md
Integralist: UPDATE: `tfplugindocs` dynamically added `description: |-` to the index.md so it would appear there's more frontmatter options than what is actually _documented_.

docs/data-sources/ip_ranges.md
Integralist: After discussing with @phamann I've decided I'm going to review each individual file and manually decide whether the curated content should stay or go. This is because some of the generated documentation files have dynamically generated Schema data that is very good and ends up making our own "Arguments Reference" and "Attributes Reference" sections completely redundant (i.e. lots of duplication).

Integralist:
OK @phamann I think this is ready for a review now. 

As mentioned in the PR description: I'm pushing back the creation of a new CI job to a separate PR.

.github/workflows/pr.yml
```diff
@@ -43,32 +43,3 @@ jobs:
           ${{ runner.os }}-go-mod-
     - name: Test
       run: make test
-  website:
```
phamann, line 46: Slightly concerned that we're now not doing any doc generation or validation in CI, is this intentional?

docs/data-sources/ip_ranges.md
phamann: So should we revert this deletion and keep these if they are in-fact supported?

docs/resources/service_compute.md
```diff
@@ -0,0 +1,621 @@
+---
```
phamann, line 1: Hmmm this file shouldn't exist.

docs/resources/service_v1.md
```diff
@@ -0,0 +1,1115 @@
+---
```
phamann, line 1: Per previous `service_compute` comment, this file shouldn't exist.

fastly/block_fastly_service_v1_package.go
```diff
@@ -23,21 +23,24 @@ func NewServicePackage(sa ServiceMetadata) ServiceAttributeDefinition {
 
 func (h *PackageServiceAttributeHandler) Register(s *schema.Resource) error {
 	s.Schema[h.GetKey()] = &schema.Schema{
-		Type:     schema.TypeList,
-		Required: true,
-		MaxItems: 1,
-		MinItems: 1,
+		Type:        schema.TypeList,
+		Required:    true,
+		Description: "The `package` block supports uploading or modifying Wasm packages for use in a Fastly Compute@Edge service. See Fastly's documentation on [Compute@Edge](https://www.fastly.com/products/edge-compute/serverless)",
```
phamann: I know not your edit, however that link would be better as a DevHub link instead of the marketing site.

scripts/generate-docs.go
```diff
@@ -0,0 +1,307 @@
...
+		{
+			name: "service_v1",
+			path: tempDir + "/resources/service_v1.md.tmpl",
+			Data: PageData{
+				"vcl",
+			},
+		},
+		{
+			name: "service_compute",
+			path: tempDir + "/resources/service_compute.md.tmpl",
```
phamann, line 93: I now realise why these files are showing as completely new in the diff for the output directory. The names of these files should match the resource name. I.e.:

```suggestion
			path: tempDir + "/resources/fastly_service_compute.md.tmpl",
```

Similar comments elided for all of these references in this file.

scripts/generate-docs.go
```diff
@@ -0,0 +1,307 @@
...
+		log.Fatal("Could not get current working directory")
+	}
+	for !strings.HasPrefix(filepath.Base(scriptPath), "terraform-provider-") && scriptPath != "/" {
+		scriptPath = filepath.Clean(scriptPath + "/..")
+	}
+	if scriptPath == "/" {
+		log.Fatal("Script was run outside of fastly provider directory")
+	}
+	return scriptPath
+}
```
phamann: I'm unsure why this is needed? Wouldn't [`os.Getwd()`](https://golang.org/pkg/os/#Getwd) be sufficient, if we know this script is invoked for a make target in the project root directory?

scripts/generate-docs.go
```diff
@@ -0,0 +1,307 @@
...
+				f.Close() // ignore error; Write error takes precedence
+				log.Fatal(err)
+			}
+			if err := f.Close(); err != nil {
+				log.Fatal(err)
+			}
+		}
+		return nil
+	})
+}
```
phamann, line 216: An alternative to injecting this at the bottom of every outputed file, is to have this inline within our template templates wrapped in a `println` template function so our precomlie templating step doesn't execute the statement. I.e. add something like this to the files:

```
{{ println "{{ .SchemaMarkdown | trimspace }}" }}
```

This would give us more fine grain control over where in the file the schema appears. Those is slightly more error prone if we forget to include in a flile.

phamann (changes requested):
Looking good, thank you for doing this, updating all of the schema descriptions must have been painful but a big win.

It would be great if we can find a way to programatically import the description data from the open-api schema 🤔?

Just a couple of outstanding queries.

docs/data-sources/ip_ranges.md
Integralist: sure I'll put back all frontmatter for now.

.github/workflows/pr.yml
Integralist: > is this intentional?

It is. For this PR at least. In the PR description I state...

**FOLLOW-UP PR**: implement a new .github/workflows/pr.yml job that will run make generate-docs and then diff the output against what's been changed to be sure any changes to the code results in new documentation being generated.

...I intend to do that very soon (as in next week) so for the short term if there's any terraform PRs that need merging I can request people run the `make generate-docs` target in the mean time.

scripts/generate-docs.go
Integralist: So although I like the idea of having control over where the schema exists in the file, I think I prefer the current implementation for a couple of reasons...

1. consistency across pages (schema appears at the bottom always rather than _potentially_ changing on a per page basis)
2. it seems like for the most part the original 'reference' sections we had (which the schema section is replacing) would appear last any way.
3. less error prone as you have already stated
4. decouples the `tfplugindocs` specific syntax (e.g. if we ever needed to migrate to a different tool)

docs/resources/service_compute.md
Integralist: why is that though? if I look at `master` I can see https://github.com/fastly/terraform-provider-fastly/blob/master/website/docs/r/service_compute.html.markdown and so that file is effectively named the same. The only difference is the new documentation tool is happy with just `.md` file extension and doesn't require `.html.markdown`.

scripts/generate-docs.go
Integralist: This doesn't look to be how it worked originally...

https://github.com/fastly/terraform-provider-fastly/tree/master/website/docs/r

...in there you can see the original 'output' directory (i.e. `website` has the resource file named `service_compute.html.markdown` which is what we have here except we've replaced the extensions with what what tfplugindocs is expecting.)

templates/data-sources/ip_ranges.md.tmpl
```diff
@@ -1,4 +1,4 @@
-{{define "fastly_ip_ranges"}}---
+{{define "ip_ranges"}}---
```
Integralist, line 1: NOTE: this was originally named `fastly_ip_ranges` because of the original `parse-templates.go` script which tried to generate a 'side menu' and was doing so based off prefixing 'fastly' into a 'Title' field. 

In the new script that generates the documentation this function no longer exists. **Is that a problem?**

```go
func generateMenuItems(pageType string, pages []Page) []MenuItem {
	var menuItems []MenuItem
	for _, p := range pages {
		menuItems = append(menuItems, MenuItem{
			Lookup: fmt.Sprintf("docs-fastly-resource-%s", strings.ReplaceAll(p.name, "_", "-")),
			Link:   fmt.Sprintf("/docs/providers/fastly/%s/%s.html", pageType, p.name),
			Title:  fmt.Sprintf("fastly_%s", p.name),
		})
	}
	return menuItems
}
```

Integralist: @phamann I don't think this part of the original script was actually used in the latest documentation because the links in the side menu of the current documentation ([here](https://registry.terraform.io/providers/fastly/fastly/latest/docs)) are `/docs/resources/service_acl_entries_v1` and **_not_** `/docs/providers/fastly/%s/%s.html`.

This makes me wonder if the side menu is now dynamically generated by Hashicorp. 

## UPDATE

That does appear to be the case when reading https://www.terraform.io/docs/registry/providers/docs.html#navigation-hierarchy

> Navigation Hierarchy: Provider docs are organized by category: resources, data sources, and guides. At a minimum, a provider must contain an index (docs/index.md) and at least one resource or data source. After publishing this provider version, its page on the Terraform Registry would display a navigation which resembles this hierarchy.

...all of which we have.

So I'm not going to put back that logic as it now appears to be confirmed as redundant/obsolete.

phamann (approved):
👍🏻 LGTM

I now see the original implementation didn't name the files correctly either. I still find it weird/inconsistent that the outputed filenames, and thus the generated URLs on the registry, don't match the name of the resource. Would be good to see what the convention with other providers is here - but not a blocker.

Integralist:
Ran the full test suite and it's passing.

PR #359, opened 2021-01-29
Update Go-Fastly dependency to 3.0.0

Integralist:
There is a new major version release for the Go-Fastly dependency. It introduces some breaking changes, which this PR addresses.

**NOTE**: as part of the current TLS implementation work, there will need to be changes added to a 3.x.x release of Go-Fastly and so integrating the TLS features into Terraform will be blocked until this PR is merged.


vendor/github.com/fastly/go-fastly/v3/fastly/client.go
```diff
@@ -46,7 +46,7 @@ const DefaultRealtimeStatsEndpoint = "https://rt.fastly.com"
 var ProjectURL = "github.com/fastly/go-fastly"
 
 // ProjectVersion is the version of this library.
-var ProjectVersion = "2.1.0"
+var ProjectVersion = "3.0.0"
 
 // UserAgent is the user agent for this particular client.
```
doramatadora, line 51:
```suggestion
// UserAgent is the user agent for this client.
```

vendor/github.com/fastly/go-fastly/v3/fastly/errors.go
```diff
@@ -0,0 +1,341 @@
...
+		if e.Meta != nil {
+			fmt.Fprintf(&b, "\n    Meta:   %v", *e.Meta)
+		}
+	}
+
+	return b.String()
+}
+
+// String implements the stringer interface and returns the string representing
+// the string text that includes the status code and corresponding status text.
```
doramatadora, line 333:
```suggestion
// String implements the string interface and returns a string 
// that includes the status code and corresponding status text.
```

doramatadora (approved):
LGTM, with some nitpicky copyedits 🎉

Integralist:
Thanks @doramatadora but unfortunately those edits will need to be made to the Go-Fastly repo. The terraform provider [vendors](https://golang.org/ref/mod#vendoring) its dependencies, so the files you've suggested changes to would be overwritten the next time someone runs `go mod vendor`.

doramatadora:
Thank you for the explanation! In that case, this looks great. I will open a PR in Go-Fastly.

Integralist:
Ran the full test suite and it's passing.

PR #362, opened 2021-02-01
CI: check if docs need to be regenerated

Integralist:
If a change is made to a schema `Description` field and a PR is opened, then we would like to be able to identify to the author of the PR that the documentation needs to be regenerated (e.g. by running `make generate-docs`).

The solution is to have a Github workflow which runs the relevant Makefile target and then runs `git diff` to see if there's any changes to what's in the main branch. If there are changes, it means new documentation files were generated in CI and so the author of the PR should run the relevant Makefile target locally and commit those changes.


Integralist:
@phamann @doramatadora I've fixed to the latest release of goreleaser now. Ready for another review whenever you have a moment.

Integralist:
@doramatadora @phamann ready for review.

GNUmakefile
```diff
@@ -52,10 +52,18 @@ test-compile:
...
 	go test -c $(TEST) $(TESTARGS)
 
-generate-docs:
+dependencies:
+	@echo "Download go.mod dependencies"
+	@go mod download
+
+install-tools: dependencies
+	@echo "Installing tools from tools/tools.go"
+	@cat tools/tools.go | grep _ | awk -F '"' '{print $$2}' | xargs -tI {} go install {}
```
phamann, line 61: Out of interest, why have you chosen this route instead of a bare`go install`?

phamann (approved):
👍🏻 LGTM - my comment is just for curiosity and not blocking.

GNUmakefile
Integralist: To clarify: `go install` needs a repo path e.g. `github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs` and so I can’t do something like `go install tools/tools.go`.

This means I needed to parse whatever is in the `tools/tools.go` file and grab all the imports. 

Admittedly there’s only one import so yeah I could just import it manually via the Makefile like so `go install github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs` but I didn’t want to couple the Makefile to a dependency that may not exist in future, nor have to update it every time a new tool is added.

doramatadora (approved):
LGTM 🚀

GNUmakefile
phamann: Yup, apologies, my understanding of `go install` was incorrect :)

PR #364, opened 2021-02-08
Resources for Custom TLS and Platform TLS products

bengesoff:
4 new resources:
- `fastly_tls_private_key`
- `fastly_tls_certificate`
- `fastly_tls_platform_certificate`
- `fastly_tls_activation`

10 new data sources
- `fastly_tls_activation`
- `fastly_tls_activation_ids`
- `fastly_tls_certificate`
- `fastly_tls_certificate_ids`
- `fastly_tls_configuration`
- `fastly_tls_configuration_ids`
- `fastly_tls_platform_certificate`
- `fastly_tls_platform_certificate_ids`
- `fastly_tls_private_key`
- `fastly_tls_private_key_ids`

This PR depends on the related `go-fastly` SDK PR (https://github.com/fastly/go-fastly/pull/259) and will need its `go.mod` updating once it gets merged. (Currently uses a `replace` directive to point to the OC fork).

The documentation has been updated to use the new format, thanks to the upstream PRs https://github.com/fastly/terraform-provider-fastly/pull/356 and https://github.com/fastly/terraform-provider-fastly/pull/362. This carries with it the same caveats as noted elsewhere with regards to the generated documentation for nested schemas.

There were a few tweaks added on top of the docs generation script. In particular I added some logic to the Makefile to install `tfplugindocs` to a project-local `./bin` directory due to the volatility in the generated documentation. I also removed the module caching from the github workflow because the tool was installed from the `vendor` directory which was already inside the git repo.

The Makefile had a command added for [sweepers](https://www.terraform.io/docs/extend/testing/acceptance-tests/sweepers.html) as well.


fastly/validators.go
```diff
@@ -140,3 +141,16 @@ func validateHTTPSURL() schema.SchemaValidateFunc {
 		return
 	}
 }
+
+func validatePEMBlock(pemType string) schema.SchemaValidateFunc {
+	return func(val interface{}, key string) ([]string, []error) {
+		b, _ := pem.Decode([]byte(val.(string))) // TODO: might want to loop here to support multiple PEM resources in one string for intermediates blob?
```
bengesoff: Outstanding question here about what the contents of the `intermediates_blob` should be. Currently only validates for one certificate but maybe there could be multiple? I asked Alexander on Friday who said he'd check

go.mod
```diff
@@ -2,11 +2,12 @@ module github.com/fastly/terraform-provider-fastly
 
 go 1.14
 
+replace github.com/fastly/go-fastly/v3 v3.0.0 => github.com/opencredo/go-fastly/v3 v3.0.0-20210208131624-9ccb395d6808
```
bengesoff: Temporary redirect to OC fork until https://github.com/fastly/go-fastly/pull/259 is merged

docs/data-sources/tls_configuration.md
```diff
@@ -0,0 +1,56 @@
...
+- **tls_service** (String) Whether the configuration should support the `PLATFORM` or `CUSTOM` TLS service.
+
+### Read-Only
+
+- **created_at** (String) Timestamp (GMT) when the configuration was created.
+- **dns_records** (Set of Object) The available DNS addresses that can be used to enable TLS for a domain. DNS must be configured for a domain for TLS handshakes to succeed. If enabling TLS on an apex domain (e.g. `example.com`) you must create four A records (or four AAAA records for IPv6 support) using the displayed global A record's IP addresses with your DNS provider. For subdomains and wildcard domains (e.g. `www.example.com` or `*.example.com`) you will need to create a relevant CNAME record. (see [below for nested schema](#nestedatt--dns_records))
+- **updated_at** (String) Timestamp (GMT) when the configuration was last updated.
+
+<a id="nestedatt--dns_records"></a>
+### Nested Schema for `dns_records`
```
bengesoff, line 50: Same issue as @Integralist raised previously (https://github.com/hashicorp/terraform-plugin-docs/issues/28) - no description on nested schema. I've left it autogenerated because I didn't think the descriptions were too critical, it is reasonably self-explanatory what the fields are for, and it seemed easier to maintain by leaving it autogenerated with the rest of the docs.

fastly/validators.go
bengesoff: I've just had a response - the suspicion was correct in that there can be multiple PEM blocks in the `intermediates_blob`. Will need to update to account for this, either as an extra on this PR or a followup depending on timings

scripts/generate-docs.go
```diff
@@ -144,7 +200,11 @@ func getTemplate(tmplDir string) *template.Template {
 		}
 		return nil
 	})
-	return template.Must(template.ParseFiles(templateFiles...))
+	template, err := template.ParseFiles(templateFiles...)
```
Integralist, line 219: Ah I didn't notice `template.ParseFiles` returned an error. Good spot in that I wouldn't have been catching any error and so the code would have imploded before reaching `template.Must`.

scripts/generate-docs.go
```diff
@@ -74,6 +74,46 @@ func main() {
 			name: "ip_ranges",
 			path: tempDir + "/data-sources/ip_ranges.md.tmpl",
 		},
+		{
```
Integralist, line 85: It would be nice now that I think about it for these slices to be dynamically populated rather than having to remember to update the script every time a new resource is added.

Integralist: But I wouldn't worry too much about it for this PR.

.github/workflows/pr.yml
```diff
@@ -52,25 +52,6 @@ jobs:
...
-      uses: actions/cache@v2
-      with:
-        path: ~/go/pkg/mod
-        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
-        restore-keys: |
-          ${{ runner.os }}-go-mod-
-    - name: Install dependencies
-      if: steps.cache.outputs.cache-hit != 'true'
-      run: make dependencies
-      shell: bash
```
Integralist, line 73: Any explanation you can give me for why these segments of the workflow have been deleted?

.gitignore
```diff
@@ -7,6 +7,7 @@ terraform.tfstate
 bin/
 modules-dev/
 /pkg/
+playground/
```
Integralist: What does this folder relate to?

GNUmakefile
```diff
@@ -52,18 +52,20 @@ test-compile:
 	fi
 	go test -c $(TEST) $(TESTARGS)
 
-dependencies:
-	@echo "Download go.mod dependencies"
-	@go mod download
-
-install-tools: dependencies
+BIN=$(CURDIR)/bin
+$(BIN)/%:
```
Integralist, line 56: Does this need to be a dynamic target? Could the target still be named `install-tools` and `generate-docs` to reference it? My thinking is that in the updated `generate-docs` we're still adding `$BIN` to `$PATH` so the `go install`'ed tools will still be located there. 

I feel like the dynamic target name (and specifically setting `$(BIN)/tfplugindocs`) is too much generalization without any real value? Although I could be mis-understanding the purpose for this refactor entirely.

GNUmakefile
```diff
@@ -52,18 +52,20 @@ test-compile:
...
+	@cat tools/tools.go | grep _ | awk -F '"' '{print $$2}' | GOBIN=$(BIN) xargs -tI {} go install {}
+
+# Inject ./bin into PATH to allow scripts/generate-docs.go to access local tfplugindocs binary
+generate-docs: $(BIN)/tfplugindocs
+	PATH=$(PATH):$(BIN) go run scripts/generate-docs.go
 
-generate-docs: install-tools
-	go run scripts/generate-docs.go
+validate-docs: $(BIN)/tfplugindocs
+	$(BIN)/tfplugindocs validate
```
Integralist, line 64: Would this not also need `PATH=$(PATH):$(BIN)` ? As you might not necessarily run `generate-docs` first.

docs/data-sources/tls_activation.md
```diff
@@ -0,0 +1,38 @@
...
+### Optional
+
+- **certificate_id** (String) ID of the TLS Certificate used.
+- **configuration_id** (String) ID of the TLS Configuration used.
+- **domain** (String) Domain that TLS was enabled on.
+- **id** (String) Fastly Activation ID. Conflicts with all other filters.
+
+### Read-Only
+
+- **created_at** (String) Timestamp (GMT) when TLS was enabled.
```
Integralist, line 38: Interesting this is GMT? I would have thought UTC would have been the preferred timezone 🤔 
Is the API determining this or is this the provider code that's configured to GMT?

Integralist: Same comment effectively everywhere we use GMT.

fastly/data_source_fastly_tls_configuration_ids.go
```diff
@@ -0,0 +1,44 @@
...
+
+	var ids []string
+	for _, configuration := range configurations {
+		ids = append(ids, configuration.ID)
+	}
+
+	// 2.x upgrade note - `hashcode.String` was removed from the SDK
+	// Code will need to be copied into this repository
+	// https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-hashcode-package
+	d.SetId(fmt.Sprintf("%d", hashcode.String(""))) // hashCode should include any filters set when they are added
```
Integralist: What's the rationale behind passing an empty string to `hashcode.String`?

docs/data-sources/tls_private_key_ids.md
```diff
@@ -0,0 +1,31 @@
...
+data "fastly_tls_private_key" "example" {
+  id = fastly_tls_private_key_ids.demo.ids[0]
+}
+```
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- **id** (String) The ID of this resource.
```
Integralist, line 27: Probably a silly question, but I've seen in a few places this pattern of having an optional `id` attribute for a data source that simply returns multiple data records, and I'm a bit confused by the purpose of it. e.g. when would you want to pass an id for this data source?

docs/resources/tls_activation.md
```diff
@@ -0,0 +1,75 @@
...
+
+resource "fastly_tls_certificate" "demo" {
+  certificate_body = "..."
+  name = "demo-cert"
+  depends_on = [fastly_tls_private_key.demo]
+}
+
+resource "fastly_tls_activation" "test" {
+  certificate_id = fastly_tls_certificate.demo.id
+  domain = "example.com"
```
Integralist, line 48: Just out of interest, but would it possible to reference the `example.com` already defined, via `fastly_service_v1.demo.domain.name` or maybe `fastly_service_v1.demo.domain[0].name` ?

templates/resources/tls_certificate.md.tmpl
```diff
@@ -0,0 +1,63 @@
...
+sidebar_current: "docs-fastly-resource-tls_certificate"
+description: |-
+Uploads a custom TLS certificate
+---
+
+# fastly_tls_certificate
+
+Uploads a custom TLS certificate to Fastly to be used to terminate TLS traffic.
+
+-> Each TLS certificate **must** have its corresponding private uploaded _prior_ to uploading the certificate. This can
```
Integralist:
```suggestion
-> Each TLS certificate **must** have its corresponding private key uploaded _prior_ to uploading the certificate. This can
```

templates/data-sources/tls_activation_ids.md.tmpl
```diff
@@ -0,0 +1,31 @@
...
+---
+
+# fastly_tls_activation_ids
+
+Use this data source to get the list of TLS Activation identifiers in Fastly.
+
+## Example Usage
+
+```hcl
+
```
Integralist: Can we remove the extraneous empty lines at the top/bottom of this code snippet.

templates/resources/tls_platform_certificate.md.tmpl
```diff
@@ -0,0 +1,98 @@
...
+sidebar_current: "docs-fastly-resource-tls_platform_certificate"
+description: |-
+Uploads a TLS certificate to the Platform TLS service
+---
+
+# fastly_tls_platform_certificate
+
+Uploads a TLS certificate to the Fastly Platform TLS service.
+
+-> Each TLS certificate **must** have its corresponding private uploaded _prior_ to uploading the certificate. This can
```
Integralist:
```suggestion
-> Each TLS certificate **must** have its corresponding private key uploaded _prior_ to uploading the certificate. This can
```

fastly/data_source_fastly_tls_activation.go
```diff
@@ -0,0 +1,167 @@
...
+func dataSourceFastlyTLSActivation() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceFastlyTLSActivationRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				Description:   "Fastly Activation ID. Conflicts with all other filters.",
+				ConflictsWith: []string{"certificate_id", "configuration_id", "domain"},
```
Integralist, line 19: I'm unclear on what the purpose of this field is.

> ConflictsWith is a set of schema keys that conflict with this schema. This will only check that they're set in the _config_. This will not raise an error for a malfunctioning resource that sets a conflicting key.

So sounds like the field doesn't do any validation other than ensuring they're _all_ set (from I can tell).

Put another way, what would happen if this line was removed? (as we'd still be checking for the conflicting keys in `getTLSActivationFilters()` if `id` was not provided).

That said, although we're using `d.GetOk` to check for the conflicting keys (which we're relying on in the absence of `id`), is the point of `ConflictsWith` to have terraform error if the user doesn't provide at least one of these fields? But then would we not prefer to use `AtLeastOneOf` rather than expecting them to provide _all_ three alternative attributes?

fastly/data_source_fastly_tls_activation.go
```diff
@@ -0,0 +1,167 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Timestamp (GMT) when TLS was enabled.",
+			},
+		},
+	}
+}
+
+func dataSourceFastlyTLSActivationRead(d *schema.ResourceData, meta interface{}) error {
+
```
Integralist: See my comment below about appropriate line spacing (and presume I've made the same suggestion elsewhere, especially as it seems we're being inconsistent with the spacing)...

```suggestion
```

fastly/data_source_fastly_tls_activation.go
```diff
@@ -0,0 +1,167 @@
...
+	conn := meta.(*FastlyClient).conn
+
+	var activation *fastly.TLSActivation
+
+	if v, ok := d.GetOk("id"); ok {
+
```
Integralist: I'm ok with breathing space around blocks of code, but I'm not sure I'm a fan of extra lines like is shown here where we have a line space after the if conditional and again before the else block. I'd prefer to see those sorts of line spacing removed.

fastly/data_source_fastly_tls_activation.go
```diff
@@ -0,0 +1,167 @@
...
+		filters = append(filters, func(c *fastly.TLSActivation) bool {
+			return c.Domain.ID == v.(string)
+		})
+	}
+
+	return filters
+}
+
+func listTLSActivations(conn *fastly.Client, filters ...TLSActivationPredicate) ([]*fastly.TLSActivation, error) {
+	var activations []*fastly.TLSActivation
```
Integralist, line 111: Is there going to be a future concern of holding a potentially unbounded number of results in-memory?

Considering we appear to only care about the zero-index from the collection, should we not break the loop after we've received a single activation result?

fastly/data_source_fastly_tls_activation_ids.go
```diff
@@ -0,0 +1,73 @@
...
+			"certificate_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "ID of TLS certificate used to filter activations",
+			},
+			"ids": {
+				Type:        schema.TypeSet,
+				Computed:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Set:         schema.HashString,
```
Integralist: Is there a reason we've used `Set` here? I've noticed we've not defined `Set` on other uses of schema.TypeSet in our provider code, so I'm wondering if that's a mistake on our part of if there's a reason why you've used it here.

My understanding is that this function (`schema.HashString`) is used internally by terraform to ensure (i.e. validate) our set data structure definitely consists of unique values.

fastly/data_source_fastly_tls_activation_ids_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/stretchr/testify/require"
+	"strings"
+	"testing"
+)
+
+func TestAccDataSourceFastlyTLSActivationIds_basic(t *testing.T) {
+	domain := fmt.Sprintf("tf-test-%s.com", acctest.RandomWithPrefix("tf-test-"))
```
Integralist: If I'm understanding the `acctest.RandomWithPrefix` function, I presume what we actually want is...

```suggestion
	domain := fmt.Sprintf("%s.com", acctest.RandomWithPrefix("tf-test-"))
```

...as the random name we're generating will already have `"tf-test-"` as part of the prefix.

fastly/sweeper_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/fastly/go-fastly/v3/fastly"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+const testResourcePrefix = "tf-test"
+
+var sweeperClients map[string]*fastly.Client
+
+func TestMain(m *testing.M) {
```
Integralist, line 16: I read the docs on `resource.TestMain` and I still don't understand how it works or what it tests 🤦🏻

Integralist: I see you assign to `sweeperClients` but then in the other functions (called from other tests) it relies on `sweeperClients` being defined so how do we ensure `TestMain` is run ahead of those other tests that also call functions like `sharedClientForRegion`?

fastly/data_source_fastly_tls_activation_ids_test.go
Integralist: Same comment applies in other test files.

fastly/data_source_fastly_tls_certificate.go
```diff
@@ -0,0 +1,223 @@
...
+
+		certificate = certificates[0]
+	}
+
+	return dataSourceFastlyTLSCertificateSetAttributes(certificate, d)
+}
+
+type TLSCertificatePredicate func(*fastly.CustomTLSCertificate) bool
+
+func getTLSCertificateFilters(d *schema.ResourceData) []TLSCertificatePredicate {
```
Integralist, line 119: There's a lot of similarities in this file to the tls_activation file. I'm guessing without having generics in the language there's no _reasonable_ away to consolidate the duplication?

fastly/data_source_fastly_tls_certificate_ids.go
```diff
@@ -0,0 +1,46 @@
...
+
+	var ids []string
+	for _, certificate := range certificates {
+		ids = append(ids, certificate.ID)
+	}
+
+	// 2.x upgrade note - `hashcode.String` was removed from the SDK
+	// Code will need to be copied into this repository
+	// https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-hashcode-package
+	d.SetId(fmt.Sprintf("%d", hashcode.String(""))) // hashCode should include any filters set when they are added
```
Integralist: I don't think I understood the comment:

> // hashCode should include any filters set when they are added

fastly/data_source_fastly_tls_configuration.go
```diff
@@ -0,0 +1,292 @@
...
+	for _, record := range configuration.DNSRecords {
+		DNSRecords = append(DNSRecords, map[string]string{
+			"record_type":  record.RecordType,
+			"record_value": record.ID,
+			"region":       record.Region,
+		})
+	}
+
+	d.SetId(configuration.ID)
+	if err := d.Set("name", configuration.Name); err != nil {
```
Integralist, line 218: When would we want to use partial state? e.g. `d.SetPartial(...)`

I've typically seen it used when dealing with 'resources', but does it ever make sense to use it with 'data sources'? and isn't that sort of what's happening here anyway (e.g. we're only getting partial data back and set into the local state before returning an error)? or does the data not get set into the state file if there is an error returned?

fastly/data_source_fastly_tls_configuration_ids_test.go
```diff
@@ -0,0 +1,21 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccFastlyAccDataSourceTLSConfigurationIDs,
+				Check:  resource.TestCheckResourceAttrSet("data.fastly_tls_configuration_ids.subject", "ids.#"),
+			},
+		},
+	})
+}
+
+const testAccFastlyAccDataSourceTLSConfigurationIDs = `data "fastly_tls_configuration_ids" "subject" {}`
```
Integralist, line 21: I typically have constants at the top of the file. Is there a benefit to have it underneath the code that calls it?

I guess it comes down to how you prefer to read code vs what makes reading code easier.

fastly/data_source_fastly_tls_configuration_ids_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func TestAccFastlyDataSourceTLSConfigurationIDs(t *testing.T) {
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccFastlyAccDataSourceTLSConfigurationIDs,
+				Check:  resource.TestCheckResourceAttrSet("data.fastly_tls_configuration_ids.subject", "ids.#"),
```
Integralist, line 15: I presume this is just validating that there are zero elements within the ids list, is my understanding correct?

fastly/provider.go
```diff
@@ -24,17 +24,31 @@ func Provider() terraform.ResourceProvider {
...
+			"fastly_tls_activation":               dataSourceFastlyTLSActivation(),
+			"fastly_tls_activation_ids":           dataSourceFastlyTLSActivationIds(),
+			"fastly_tls_certificate":              dataSourceFastlyTLSCertificate(),
+			"fastly_tls_certificate_ids":          dataSourceFastlyTLSCertificateIDs(),
+			"fastly_tls_configuration":            dataSourceFastlyTLSConfiguration(),
+			"fastly_tls_configuration_ids":        dataSourceFastlyTLSConfigurationIDs(),
+			"fastly_tls_platform_certificate":     dataSourceFastlyTLSPlatformCertificate(),
+			"fastly_tls_platform_certificate_ids": dataSourceFastlyTLSPlatformCertificateIDs(),
+			"fastly_tls_private_key":              dataSourceFastlyTLSPrivateKey(),
+			"fastly_tls_private_key_ids":          dataSourceTLSPrivateKeyIDs(),
```
Integralist: Should this not be named consistently with the other data sources?
```suggestion
			"fastly_tls_private_key_ids":          dataSourceFastlyTLSPrivateKeyIDs(),
```

That said, I notice the 'resources' listed below aren't named consistently in that they don't mention the provider name `Fastly`. I'm not suggesting we change that as part of this PR as it's an unrelated refactoring, but more from an idiomatic terraform code perspective shouldn't they be `<resource|dataSource>Fastly<Name>`?

e.g.

```diff
- resourceServiceV1
+ resourceFastlyServiceV1
```

fastly/data_source_fastly_tls_private_key_ids.go
```diff
@@ -0,0 +1,47 @@
+package fastly
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/hashcode"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceTLSPrivateKeyIDs() *schema.Resource {
```
Integralist: Should this not be named consistently with the other data sources?
```suggestion
func dataSourceFastlyTLSPrivateKeyIDs() *schema.Resource {
```

I presume we'd want to update the READ function as well: 

```diff
- dataSourceTLSPrivateKeyIdsRead
+ dataSourceFastlyTLSPrivateKeyIdsRead
```

fastly/data_source_fastly_tls_private_key.go
```diff
@@ -0,0 +1,189 @@
...
+	"fmt"
+	"time"
+
+	"github.com/fastly/go-fastly/v3/fastly"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceFastlyTLSPrivateKey() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceTLSPrivateKeyRead,
```
Integralist: Should we keep the CRUD function names consistent with the parent resource?

e.g.
```suggestion
		Read: dataSourceFastlyTLSPrivateKeyRead,
```

fastly/resource_fastly_tls_activation_test.go
```diff
@@ -0,0 +1,160 @@
...
+  depends_on = [fastly_service_v1.test]
+}
+`, serviceName, domain, key, keyName, cert, certName, domain)
+}
+
+func testAccFastlyTLSActivationCheckExists(resourceName string) resource.TestCheckFunc {
+	return func(state *terraform.State) error {
+		conn := testAccProvider.Meta().(*FastlyClient).conn
+
+		r := state.RootModule().Resources[resourceName]
```
Integralist: Isn't there a possibility this will cause a runtime panic if there is no `resourceName` key defined? I appreciate the test will fail and that's a good thing, but that maybe we should handle that error scenario a bit more cleanly if so. What do you reckon?

Integralist (commented):
@bengesoff this PR looks great! I've left some questions and minor suggestions. Let me know your thoughts when you get a chance.

scripts/generate-docs.go
bengesoff: Yes I think I was having some trouble with it when I was moving some of the old style templates over and it wasn't happy about something in the parsing step so it helped to debug that - figured it'd probably be helpful for someone else too.

scripts/generate-docs.go
bengesoff: Yeah that could be a nice enhancement - scanning the directories or something

.github/workflows/pr.yml
bengesoff: Yes this change tied in with some of the other changes relating to the way `tfplugindocs` was being installed. I think previously the makefile was installing the package to `GOROOT` (i.e. `~/go/pkg` and `~/go/bin`), which would explain the steps in the workflow caching these directories. 

However when adapting the upstream docs generation changes I found that the output of the `tfplugindocs` tool was quite volatile, given its alpha status, and thought it made sense to use the `vendor` style dependency management to pin in the exact version of the source code used to generate the docs and reduce disparity between all of the contributor's versions of it. In a similar vein I changed it to install the binary to a project-local `./bin` directory to isolate it in case other terraform plugin projects were using different versions of it on anyone's machine. (I think you also made some very similar changes in parallel, e.g. adding a `tools.go` to allow `go mod` to find it, so I ended up merging my tweaks into that as well).

Anyway long story short, with the source code being installed to `./vendor` and the compiled `tfplugindocs` binary being installed to `./bin`, I didn't see a need to manually cache the external global `GOROOT` directories as the GitHub workflow does. (Also it had stopped working for me when I removed the `make dependencies` target, so I had to fix it somehow! 😉)

Does this make sense do you think? Happy to jump on a call and discuss it further if you think that would be helpful

.gitignore
bengesoff: Not sure I think someone must have been experimenting - I'll remove it

GNUmakefile
bengesoff: This was related to the same change as explained in the comment on `.github/workflows/pr.yml`. As part of moving the installation of `tfplugindocs` to be inside the `terraform-provider-fastly` directory, I used the non-PHONY target to avoid needing to re-run it every time, if the binary already exists. By extension, the dynamic target seemed useful in case other dependencies were added, given that the grep/awk command was already capable of installing the whole `tools.go` file.

That's the rationale anyway - I'm happy to change it to use a PHONY `install-tools` target if you'd prefer, and if I'm not mistaken that would rebuild the tool on each invocation, which could be helpful if it were upgraded? That's assuming that you agree installing to a project-local `terraform-provider-fastly/bin` directory is a good idea anyway, which you might not - also ok!

GNUmakefile
bengesoff: I don't believe so, as it's calling the binary from its relative filepath instead of letting the shell look it up on the system `PATH`, but I might be wrong. In the case of `generate-docs` the script calls it without specifiying the location so the `PATH` is needed there. 

Now you mention it, it could be cleaner to specify `./bin/tfplugindocs` inside the script to avoid modifying the `PATH` for that command, if we decide that a non-global install is the right way to go. Or if could be nice to leave it configurable from the Makefile too so the location is saved in one place, in which case maybe passing an argument in could be best.

docs/data-sources/tls_activation.md
bengesoff: This was taken directly from the API docs I believe!

fastly/data_source_fastly_tls_configuration_ids.go
bengesoff: If I recall correctly, this was mainly as a placeholder anticipating the addition of filters to this data source at some point, similarly to how the [`aws_ami_ids`](https://github.com/hashicorp/terraform-provider-aws/blob/bc48c2178fff1c5b1b142481069e45fcb5870bd5/aws/data_source_aws_ami_ids.go#L111) data source does it, but in this case there are no filters so an empty string was used in the mean time. It could be hardcoded to `0` if you think that would be cleaner, and a `hashcode` or similar could be added in at a later point if additional filters are added?

docs/data-sources/tls_private_key_ids.md
bengesoff: Good spot, that seems to be an artifact of `tfplugindocs` - I agree there isn't a purpose of actually specifying this. What do you think the policy generally should be on tfplugindocsisms like this?

docs/resources/tls_activation.md
bengesoff: I'm not sure this would work, because `fastly_service_v1.demo.domain` refers to a set, not a list, and so can't be indexed directly. Alternatively `tolist(fastly_service_v1.test.domain)[0].name` could be used, or `[for d in fastly_service_v1.domain : d. name][0]` etc. Whichever you think is clearest between these or hardcoding 🙂

fastly/data_source_fastly_tls_activation.go
bengesoff: We used this pattern in all the singular data sources, I believe, to enforce that _either_ the `id` field is used, _or_ 0-3 of the other filters, but not both. This is so that the `id` field can be used in conjunction with the corresponding plural data source, in which case it wouldn't make any sense to specify another filter as it's just a simple ID lookup. Similarly, it wouldn't make sense for a user to specify an ID if they were trying to filter based on another attribute.

So you're right in that it's just to get a nice Terraform error for the user, but it's less about enforcing they're all set, just that none of those listed are also set when `id` is present. Put differently, `ConflictsWith` doesn't care how many of the listed fields are set, as long as none of them are set when `id` is.

fastly/data_source_fastly_tls_activation.go
bengesoff: Good spot, thanks!

fastly/data_source_fastly_tls_activation.go
bengesoff: Agreed, I'm afraid I didn't notice these, will update.

fastly/data_source_fastly_tls_activation.go
bengesoff: This is a good point, yes. In the case of TLS activations we do not share the `list*` function between the singular and plural data sources (because the plural one uses an API filter directly whereas the others do all the filtering after pulling the results down). Therefore in this case, we could just modify this function to break the loop early once the filter has passed. However, in most (if not all) of the other data sources, we do share the `list*` function between singular and plural data sources, to abstract away the API's inconsistent pagination methods, so we might lose the consistency among the data sources in this case.

Do you think the consistency with other data sources takes precedence, or should we take the easy code optimisation here while we can?

fastly/data_source_fastly_tls_activation_ids.go
bengesoff: Well spotted - I've just looked this up and it's apparently not needed anymore as Terraform:

> now does the right thing, even with nested data structures

so I will go ahead and remove these!

fastly/sweeper_test.go
bengesoff: I'll be honest I haven't really looked into how it works myself - I saw it looked very similar to [what the docs recommend](https://www.terraform.io/docs/extend/testing/acceptance-tests/sweepers.html) and left it at that! I believe the testing framework (called by the Makefile with `go test ./fastly -v -sweep=ALL ........`) takes care of running `TestMain` etc in the correct order.

If you're asking more about what the purpose is, the link above explains it quite well, but essentially it's just a "cleanup" script to delete any leaked resources that were provisioned by a test and that didn't get correctly deleted. You can run `make sweep` at some point after running the tests and it'll take care of deleting anything that matches a certain criteria, which in this case is the `testResourcePrefix` prefix

fastly/data_source_fastly_tls_certificate.go
bengesoff: This was definitely one that seemed a good candidate for generics, agreed! I don't believe so, no, if we maintain the "_reasonable_" caveat. Off the top of my head the main option would be to pass around `interface{}`s instead of the `fastly.*` types and either forfeit type safety or have a load of type switches around obscuring the logic. If you think of anything, let me know, otherwise I guess we're waiting until December or whenever generics gets added!

fastly/data_source_fastly_tls_certificate_ids.go
bengesoff: Maybe the wording isn't clear - it's supposed to indicate that, if additional filters are added to the data source (such as `tls_activation_ids`) then the filters should be used instead of the empty string. I'll try and clear this up

fastly/data_source_fastly_tls_configuration.go
bengesoff: I don't think it's ever needed as it's deprecated in SDK 1.x and removed in SDK 2.x. If there's an error then the state should get refreshed on the next run anyway so there shouldn't be any need to partially fill the state if I'm not mistaken? The only edge case I can possibly think of is if there is some one-time piece of information that won't be available on the next call to `Read` but I don't believe that applies anywhere here.

fastly/data_source_fastly_tls_configuration_ids_test.go
bengesoff: I think this is underneath because it's a test piece of HCL config and they seem to normally be at the bottom of the files when they are functions, so would make sense to group them even if some of them don't have any variable data. Happy to shift it around if you'd prefer.

fastly/data_source_fastly_tls_configuration_ids_test.go
bengesoff: I think it's just validating that the `ids` field has an attribute for its length and therefore loaded without error, not that it is specifically zero. There's not really anything interesting to test here because we don't know what configurations, if any, might be present and we can't create one and look for it, as we do in the other resources (eg activations)

fastly/provider.go
bengesoff: Ahh yes good point, I did this check on filenames before opening the PR but missed the function names 😩

bengesoff:
@Integralist I have fixed&resolved some comments, and replied to the rest. Thanks for the feedback, caught some things that had slipped through and raised some good points!

.github/workflows/pr.yml
Integralist: No that's fine. Thanks for the background information 👍🏻

GNUmakefile
Integralist: OK that makes sense. I think avoiding reinstalling the tool every time is better any way and we can suffer not being on the bleeding edge all the time.

docs/data-sources/tls_activation.md
Integralist: Fair enough then 🙂

fastly/data_source_fastly_tls_configuration_ids.go
Integralist: > anticipating the addition of filters to this data source at some point

if it's likely we're going to see filters come in the future, then I think this is fine to leave 'as is', otherwise I'd suggest we set some kind of value to be hashed.

docs/data-sources/tls_private_key_ids.md
Integralist: Hmm, I guess we just leave it in if that's what the plugin is doing (I mean it is the _official_ tool now so it makes sense not to fight against it).

docs/resources/tls_activation.md
Integralist: OK good to know. I asked only for my own learning of the terraform hcl syntax (of which I'm still very green on). 👍🏻

fastly/data_source_fastly_tls_activation.go
Integralist: > ConflictsWith doesn't care how many of the listed fields are set, as long as none of them are set when id is.

Gotcha. Thanks! 👍🏻

fastly/data_source_fastly_tls_activation.go
Integralist: I think consistency is fine in this case. If there's a performance issue that arises we can address it when it happens.

fastly/data_source_fastly_tls_certificate.go
Integralist: 👍🏻

fastly/data_source_fastly_tls_certificate_ids.go
Integralist: Awesome, thanks!

fastly/data_source_fastly_tls_configuration.go
Integralist: > removed in SDK 2.x

oh really! lulz

fastly/data_source_fastly_tls_configuration_ids_test.go
Integralist: No that's fine. I didn't really pay attention to the fact it was mock hcl, and I agree that those always appear to be at the bottom of the test file 👍🏻

GNUmakefile
bengesoff: Just a heads up here - I've pushed in a change to avoid modifying `PATH` for the `generate-docs`

Integralist (approved):
@bengesoff LGTM.

I'll get this merged and then if you could let me know once you've rebased the latest master into your #365 PR.

Thanks!

PR #365, opened 2021-02-08
Managed TLS Subscriptions Resources

bengesoff:
2 new resources:
- `fastly_tls_subscription`
- `fastly_tls_subscription_validation`

3 new data sources:
- `fastly_tls_domain`
- `fastly_tls_subscription`
- `fastly_tls_subscription_ids`

This PR depends on https://github.com/fastly/terraform-provider-fastly/pull/364.

Does not include support for `PATCH` endpoint yet, will add in a separate PR.


docs/resources/tls_subscription.md
```diff
@@ -0,0 +1,170 @@
...
+sidebar_current: "docs-fastly-resource-tls_subscription"
+description: |-
+Enables TLS on a domain using a managed certificate
+---
+
+# fastly_tls_subscription
+
+Enables TLS on a domain using a certificate managed by Fastly.
+
+To make it work, DNS records need to be modified on the domain being secured, in order to respond to the ACME domain ownership challenge.
```
Integralist:
```suggestion
DNS records need to be modified on the domain being secured, in order to respond to the ACME domain ownership challenge.
```

Integralist: This is rendered as one big chunk of text which is hard to read through. Can we split this up into three distinct sections:

```
DNS records need to be modified on the domain being secured, in order to respond to the ACME domain ownership challenge. There are two options for doing this: the `managed_dns_challenge`, which is the default method; and the `managed_http_challenges`, which points production traffic to Fastly.

~> See the [Fastly documentation](https://docs.fastly.com/en/guides/serving-https-traffic-using-fastly-managed-certificates#verifying-domain-ownership) for more information on verifying domain ownership.

The example below demonstrates usage with AWS Route53 to configure DNS, and the `fastly_tls_subscription_validation` resource to wait for validation to complete.
```

NOTE: I amended the last paragraph to reflect the 'suggested' change below.

docs/resources/tls_subscription.md
```diff
@@ -0,0 +1,170 @@
...
+There are two options for doing this: the `managed_dns_challenge`, which is the default method; and the `managed_http_challenges`, which points production traffic to Fastly.
+See the [Fastly documentation](https://docs.fastly.com/en/guides/serving-https-traffic-using-fastly-managed-certificates#verifying-domain-ownership) for more information on verifying domain ownership.
+The example below shows example usage with AWS Route53 to configure DNS, and the `fastly_tls_subscription_validation` resource to wait for validation to complete.
```
Integralist:
```suggestion
The example below demonstrates usage with AWS Route53 to configure DNS, and the `fastly_tls_subscription_validation` resource to wait for validation to complete.
```

docs/resources/tls_subscription.md
```diff
@@ -0,0 +1,170 @@
...
+
+# Set up DNS record for managed DNS domain validation method
+resource "aws_route53_record" "domain_validation" {
+  name            = fastly_tls_subscription.example.managed_dns_challenge.record_name
+  type            = fastly_tls_subscription.example.managed_dns_challenge.record_type
+  zone_id         = data.aws_route53_zone.demo.id
+  allow_overwrite = true
+  records         = [fastly_tls_subscription.example.managed_dns_challenge.record_value]
+  ttl             = 60
+}
```
Integralist: I think this code would read better if we moved the `aws_route53_zone ` and `aws_route53_record` resources further down so that `fastly_service_v1` is shown first, followed by `fastly_tls_subscription`, and _then_ show the aws resources, followed lastly by `fastly_tls_subscription_validation`.

fastly/resource_fastly_tls_subscription.go
```diff
@@ -0,0 +1,233 @@
...
+)
+
+func resourceFastlyTLSSubscription() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceFastlyTLSSubscriptionCreate,
+		Read:   resourceFastlyTLSSubscriptionRead,
+		Delete: resourceFastlyTLSSubscriptionDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
```
Integralist, line 19: Are you able to clarify how 'importing' works in terraform? 

From what I can tell the importer works by calling the function assigned to `State` and in this case all we're doing is passing the ID the user provides as part of the `terraform import <resource.instance> <id>` command into the `schema.ImportStatePassthrough` function which itself is a simple one line function that returns a `schema.ResourceData` instance with just the ID assigned.

So I imagine the `terraform import` cli subcommand must then also internally trigger a call to the relevant READ function (in this case `resourceFastlyTLSSubscriptionRead`), and so that's what ends up populating the terraform state.

docs/resources/tls_subscription_validation.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Set up DNS record for managed DNS domain validation method
+resource "aws_route53_record" "domain_validation" {
+  name            = fastly_tls_subscription.example.managed_dns_challenge.record_name
+  type            = fastly_tls_subscription.example.managed_dns_challenge.record_type
+  zone_id         = data.aws_route53_zone.demo.id
+  allow_overwrite = true
+  records         = [fastly_tls_subscription.example.managed_dns_challenge.record_value]
+  ttl             = 60
+}
```
Integralist: As mentioned earlier I'd like to see these resources moved down under `fastly_tls_subscription`.

fastly/data_source_fastly_tls_domain.go
```diff
@@ -0,0 +1,133 @@
...
+	"fmt"
+
+	"github.com/fastly/go-fastly/v3/fastly"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceFastlyTLSDomain() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceTLSDomainsRead,
+
```
Integralist:
```suggestion
```

fastly/data_source_fastly_tls_domain.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"fmt"
+
+	"github.com/fastly/go-fastly/v3/fastly"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceFastlyTLSDomain() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceTLSDomainsRead,
```
Integralist: As per the other PR: let's keep the naming of functions consistent so in this case:
```suggestion
		Read: dataSourceFastlyTLSDomainsRead,
```

If you could apply this throughout the PR please.

fastly/data_source_fastly_tls_domain.go
```diff
@@ -0,0 +1,133 @@
...
+	return nil
+}
+
+func findTLSDomain(conn *fastly.Client, d *schema.ResourceData) (*fastly.TLSDomain, error) {
+	var filters []func(*fastly.TLSDomain) bool
+
+	domain := d.Get("domain").(string)
+	filters = append(filters, func(d *fastly.TLSDomain) bool {
+		return d.ID == domain
+	})
```
Integralist: Is this designed this way to allow for more filters to be added at a later stage? If so I wonder if it's worth removing the logic until it's needed (e.g. just pass a single `filter` function into `listTLSDomains`, and subsequently into `filterTLSDomain`).

> Note: I'm presuming nothing else is calling `listTLSDomains` or `filterTLSDomain` and is passing them a slice of filter functions.

Integralist (approved):
@bengesoff Thanks for the PR! Just a couple of minor comments/questions. Otherwise approved.

docs/resources/tls_subscription.md
bengesoff: Yes sounds good, I'll copy the suggested changes across manually so it's included in the templates too 👍🏼

fastly/resource_fastly_tls_subscription.go
bengesoff: I think you're pretty much spot on, yes, as far as I understand it.

In our case the entire state can be constructed from just the ID in the Read function so the `schema.ImportStatePassthrough` is recommended. If the implementation of Read required other attributes to be set or derived beforehand then something more custom would be required in the State importer to set this up.

So yes I think the process is, as you described it: run the `Importer.State` function with whatever was provided to `terraform import`; then call the `Read` function to populate the state. There's a separate test for this as well, where it imports a resource it has just created and checks that the state matches

fastly/data_source_fastly_tls_domain.go
bengesoff: Yes for extensibility in adding more filters, and also consistency with the other data sources (they mostly follow the same pattern of having a `listTLS*` function hiding the pagination and passing in filters).

I've made a change, let me know what you think, happy to go further if you'd like 🙂

Integralist (approved):

PR #366, opened 2021-02-10
Ensure schema.Set uses custom SetDiff algorithm

Integralist:
**Background**: we implemented `SetDiff` to identify when two elements across _distinct_ sets have the same key. This helps avoid recreating an element (i.e. DELETE, then CREATE) which only needs an UPDATE.

**Problem**: not all schema.Set objects are using the `SetDiff` algorithm.

**Caveats**: there is a separate concern with some resources being deleted/recreated. Specifically resources that require a lookup key of "name" and who are also expecting the name field to be updated (e.g. updating the name of an ACL). This will be addressed in a separate PR.

**Notes**: I've not updated `block_fastly_waf_configuration_v1_active_rule.go` as it appears to be doing some kind of filtering of data which would be incompatible with the SetDiff implementation. Nor have I updated the `block_fastly_waf_configuration_v1_exclusion.go` as the nested structuring of the resources was quite complicated to follow and so it was safer not to mess with it for now.

### DIFF
--- a/fastly/block_fastly_service_v1_acl.go
+++ b/fastly/block_fastly_service_v1_acl.go
@@ -83,33 +83,14 @@ func (h *ACLServiceAttributeHandler) Process(d *schema.ResourceData, latestVersi
 		}
 	}
 
-	// UPDATE modified resources
+	// UPDATE modified resources (NOT IMPLEMENTED)
 	//
-	// NOTE: although the go-fastly API client enables updating of a resource by
+	// Although the go-fastly API client enables updating of a resource by
 	// its 'name' attribute, this isn't possible within terraform due to
 	// constraints in the data model/schema of the resources not having a uid.
-	for _, resource := range diffResult.Modified {
-		resource := resource.(map[string]interface{})
-
-		opts := gofastly.UpdateACLInput{
-			ServiceID:      d.Id(),
-			ServiceVersion: latestVersion,
-			Name:           resource["name"].(string),
-		}
-
-		// only attempt to update attributes that have changed
-		modified := setDiff.Filter(resource, oldSet)
-
-		if v, ok := modified["name"]; ok {
-			opts.NewName = v.(string)
-		}
-
-		log.Printf("[DEBUG] Update ACL Opts: %#v", opts)
-		_, err := conn.UpdateACL(&opts)
-		if err != nil {
-			return err
-		}
-	}
+	//
+	// Because of this we do not implement any logic for updating the dictionary
+	// resource, only CREATE and DELETE functionality.
 
 	return nil
 }
--- a/fastly/block_fastly_service_v1_acl.go
+++ b/fastly/block_fastly_service_v1_acl.go
@@ -33,11 +33,18 @@ func (h *ACLServiceAttributeHandler) Process(d *schema.ResourceData, latestVersi
 	oldACLSet := oldACLVal.(*schema.Set)
 	newACLSet := newACLVal.(*schema.Set)
 
-	remove := oldACLSet.Difference(newACLSet).List()
-	add := newACLSet.Difference(oldACLSet).List()
+	setDiff := NewSetDiff(func(acl interface{}) (interface{}, error) {
+		// Use the acl name as the key
+		return acl.(map[string]interface{})["name"], nil
+	})
+
+	diffResult, err := setDiff.Diff(oldACLSet, newACLSet)
+	if err != nil {
+		return err
+	}
 
 	// Delete removed ACL configurations
-	for _, vRaw := range remove {
+	for _, vRaw := range diffResult.Deleted {
 		val := vRaw.(map[string]interface{})
 		opts := gofastly.DeleteACLInput{
 			ServiceID:      d.Id(),
@@ -58,7 +65,7 @@ func (h *ACLServiceAttributeHandler) Process(d *schema.ResourceData, latestVersi
 	}
 
 	// POST new ACL configurations
-	for _, vRaw := range add {
+	for _, vRaw := range diffResult.Added {
 		val := vRaw.(map[string]interface{})
 		opts := gofastly.CreateACLInput{
 			ServiceID:      d.Id(),
--- a/fastly/block_fastly_service_v1_acl.go
+++ b/fastly/block_fastly_service_v1_acl.go
@@ -168,20 +168,20 @@ func (h *ACLServiceAttributeHandler) Register(s *schema.Resource) error {
 func flattenACLs(aclList []*gofastly.ACL) []map[string]interface{} {
 	var al []map[string]interface{}
 	for _, acl := range aclList {
-		// Convert VCLs to a map for saving to state.
-		vclMap := map[string]interface{}{
+		// Convert ACLs to a map for saving to state.
+		aclMap := map[string]interface{}{
 			"acl_id": acl.ID,
 			"name":   acl.Name,
 		}
 
 		// prune any empty values that come from the default string value in structs
-		for k, v := range vclMap {
+		for k, v := range aclMap {
 			if v == "" {
-				delete(vclMap, k)
+				delete(aclMap, k)
 			}
 		}
 
-		al = append(al, vclMap)
+		al = append(al, aclMap)
 	}
 
 	return al
--- a/fastly/block_fastly_service_v1_acl.go
+++ b/fastly/block_fastly_service_v1_acl.go
@@ -30,15 +30,15 @@ func (h *ACLServiceAttributeHandler) Process(d *schema.ResourceData, latestVersi
 		newACLVal = new(schema.Set)
 	}
 
-	oldACLSet := oldACLVal.(*schema.Set)
-	newACLSet := newACLVal.(*schema.Set)
+	oldSet := oldACLVal.(*schema.Set)
+	newSet := newACLVal.(*schema.Set)
 
-	setDiff := NewSetDiff(func(acl interface{}) (interface{}, error) {
-		// Use the acl name as the key
-		return acl.(map[string]interface{})["name"], nil
+	setDiff := NewSetDiff(func(resource interface{}) (interface{}, error) {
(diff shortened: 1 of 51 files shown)


fastly/resource_fastly_service_acl_entries_v1.go
```diff
@@ -141,14 +141,21 @@ func resourceServiceAclEntriesV1Update(d *schema.ResourceData, meta interface{})
...
-		oes := oe.(*schema.Set)
-		nes := ne.(*schema.Set)
+		oldSet := oe.(*schema.Set)
+		newSet := ne.(*schema.Set)
 
-		removeEntries := oes.Difference(nes).List()
-		addEntries := nes.Difference(oes).List()
+		setDiff := NewSetDiff(func(resource interface{}) (interface{}, error) {
+			// Use the resource ID as the key
+			return resource.(map[string]interface{})["id"], nil
```
Integralist: This is the computed ID within the "entry" object which is what I _think_ is required. 
The reason I selected that as the key is because just before this code is `d.HasChange("entry")`.
@phamann you might be best to sanity check me here.

fastly/block_fastly_waf_configuration_v1_exclusion.go
```diff
@@ -112,20 +112,25 @@ func updateWAFRuleExclusions(d *schema.ResourceData, meta interface{}, wafID str
...
-	oss := os.(*schema.Set)
-	nss := ns.(*schema.Set)
+	oldSet := os.(*schema.Set)
+	newSet := ns.(*schema.Set)
 
-	add := nss.Difference(oss).List()
-	remove := oss.Difference(nss).List()
+	setDiff := NewSetDiff(func(resource interface{}) (interface{}, error) {
+		// Use the resource name as the key
+		return resource.(map[string]interface{})["name"], nil
```
Integralist: This is the "name" attribute within the top level `wafRuleExclusion` resource which is what I _think_ is required. 
The reason I selected that as the key is because just before this code is `d.GetChange("rule_exclusion")`.
The `rule_exclusion` attribute exists in a different file that assigns to it `wafRuleExclusion`.
@phamann you might be best to sanity check me here.

triblondon (commented):
I'm not qualified to comment on this so if Patrick can review that would be really appreciated!

fastly/block_fastly_service_v1_director.go
```diff
@@ -30,14 +30,21 @@ func (h *DirectorServiceAttributeHandler) Process(d *schema.ResourceData, latest
...
 	}
 
-	ods := od.(*schema.Set)
-	nds := nd.(*schema.Set)
+	oldSet := od.(*schema.Set)
+	newSet := nd.(*schema.Set)
 
-	removeDirector := ods.Difference(nds).List()
-	addDirector := nds.Difference(ods).List()
+	setDiff := NewSetDiff(func(resource interface{}) (interface{}, error) {
```
Integralist, line 36: @phamann Hmm. Might tap you up Monday (or whenever you have time, if any) for a pairing session to help catch some test errors.

phamann (changes requested):
Hey @Integralist. Thanks for doing this, its been a very long-running need in the project! 

I started reviewing but soon realised you haven't added the `Update` code path for any of the resources just `Add` and `Delete`, hence using the current implementation any updates to resources will not be applied and is the core benefit of moving to this pattern - this may also be why you have failing tests. 

Use the backend block as the example, specifically these lines: https://github.com/fastly/terraform-provider-fastly/blob/master/fastly/block_fastly_service_v1_backend.go#L75-L92

Hope that helps :)

Integralist:
FYI @phamann in the latest commit https://github.com/fastly/terraform-provider-fastly/pull/366/commits/ee842edff84dc2288ed31c6f9567f7b7e511e146 I've added some documentation around the caveats I encountered and for which we discussed offline. I didn't want those to get missed in the review as I feel they're important to call out.

Integralist:
@phamann I've updated the PR description to reflect the files I'm not including (for various reasons).

fastly/block_fastly_service_v1_acl.go
```diff
@@ -72,6 +82,16 @@ func (h *ACLServiceAttributeHandler) Process(d *schema.ResourceData, latestVersi
...
 	}
+
+	// UPDATE modified resources (NOT IMPLEMENTED)
+	//
+	// Although the go-fastly API client enables updating of a resource by
+	// its 'name' attribute, this isn't possible within terraform due to
+	// constraints in the data model/schema of the resources not having a uid.
+	//
+	// Because of this we do not implement any logic for updating the dictionary
+	// resource, only CREATE and DELETE functionality.
```
Integralist, line 93: As a temporary measure (until we get all `TypeSet` replaced with `TypeList`) I've updated the documentation to reflect that updating the `name` attribute will result in the resource being DELETED and then CREATED (thus losing all data contained within that resource).

Integralist:
@pschulte @bridgetlane 👋🏻 I'm looking for other engineers experienced with either our terraform provider or terraform in general to give a review of this PR.

fastly/block_fastly_service_v1_backend.go
```diff
@@ -91,6 +94,111 @@ func (h *BackendServiceAttributeHandler) Process(d *schema.ResourceData, latestV
 		}
 	}
 
+	// UPDATE modified resources
+	//
+	// NOTE: although the go-fastly API client enables updating of a resource by
+	// its 'name' attribute, this isn't possible within terraform due to
+	// constraints in the data model/schema of the resources not having a uid.
+	for _, resource := range diffResult.Modified {
```
phamann, line 83: This file already processes updates in a `diffResult.Modified` loop on line 79. So assume this is a mistake, as we don't want to apply the updates twice 😉

fastly/block_fastly_service_v1_director.go
```diff
@@ -106,6 +117,84 @@ func (h *DirectorServiceAttributeHandler) Process(d *schema.ResourceData, latest
...
+		_, err := conn.UpdateDirector(&opts)
+		if err != nil {
+			return err
+		}
+
+		if v, ok := modified["backends"]; ok {
+			backends := v.(*schema.Set).List()
+			if len(backends) > 0 {
+				for _, backend := range backends {
+					opts := gofastly.CreateDirectorBackendInput{
```
phamann, line 182: Hmmm. It seems our API doesn't have an `PUT` to update a director backend and thus we also don't have a `gofaslty.UpdateDirectorBackend()`. It would be good to investigate what the consequences of recreating the relationship in an update loop is, if they haven't changed - hopefully the action is idempotent.

fastly/block_fastly_service_v1_director.go
```diff
@@ -106,6 +117,84 @@ func (h *DirectorServiceAttributeHandler) Process(d *schema.ResourceData, latest
...
+						ServiceID:      d.Id(),
+						ServiceVersion: latestVersion,
+						Director:       resource["name"].(string),
+						Backend:        backend.(string),
+					}
+
+					log.Printf("[DEBUG] Director Backend Create opts: %#v", opts)
```
phamann: Related to the above comment, Is `create` the correct verb here?

fastly/block_fastly_service_v1_snippet.go
```diff
@@ -33,19 +33,32 @@ func (h *SnippetServiceAttributeHandler) Process(d *schema.ResourceData, latestV
...
+		return t["name"], nil
+	})
+
+	diffResult, err := setDiff.Diff(oldSet, newSet)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("\n\nold values: %+v\n\n", oldSet)
+	fmt.Printf("\n\nnew values: %+v\n\n", newSet)
+	fmt.Printf("\n\ndiffResult: %+v\n\n", diffResult)
```
phamann: I assume these were accidentally left in after debugging and not intentional?

```suggestion
```

phamann (approved):
👍🏻 Looking good - there are a couple of minor questions, some things to double check and some inconsistencies. However approving now to unblock you.

I left most comments inline, however a couple of other overarching observations:
- It seems there is some inconsistency when you've chosen to note in the public  documentation about the lack of ability to update a name.  For instance non of the logging endpoints have this note - Is this intentional? I'd suggest we do this everywhere if we are going to do it.
-  The original implementation of this pattern in [fastly/block_fastly_service_v1_backend.go](https://github.com/fastly/terraform-provider-fastly/pull/366/files#diff-8ddf1a06c3b1cf572fb5eb2c4a43fa695e64c322d17f4dd227d0943fa3c326ea) abstracted the building of the update struct fields to a `buildUpdateBackendInput()` method, and you've inlined the logic into `Process()`- I personally feel the former pattern makes process easier to read/grok and enables us to unit test the update logic with mock data if we want in the future.

Integralist:
re:  

> The original implementation of this pattern in fastly/block_fastly_service_v1_backend.go abstracted the building of the update struct fields to a buildUpdateBackendInput() method, and you've inlined the logic into Process()- I personally feel the former pattern makes process easier to read/grok and enables us to unit test the update logic with mock data if we want in the future.

That's reasonable. I've made a ticket for doing this work in a separate PR.

fastly/block_fastly_service_v1_director.go
Integralist: @phamann I created a go-fastly script to validate what happens if we try to create a backend that already exists, and it seems we'll get a `409 Conflict` error.

One way to avoid any errors is to ignore 409's but because go-fastly currently doesn't return any useful contextual information about the error that came back from the API (note: I have made a ticket for this annoyance) I would need to do something like:

```go
// Ignore a 409 Conflict
if !strings.Contains(err.Error(), "409 - Conflict") {
    return err
}
```

Integralist: I should add, the tests appear to update a backend's name and it's picked up as 'modified' so it looks like we wouldn't end up trying to recreate existing director backends.

PR #370, opened 2021-02-18
PATCH endpoint for TLS subscriptions

bengesoff:
Adds support for the `PATCH` endpoint for updating TLS Subscriptions. Currently a Limited Availability feature so this won't work for all customers.

Similar to the `force_destroy` flag there is also a `force_update` flag which gets passed through to the API to enable making changes to production traffic. This is so you have to opt-in to this behaviour and take responsibility for the risk this may have. It is worth noting that there doesn't seem to be any case where **not** using this flag has any effect.

It appears the `PATCH` endpoint only supports updating the `configuration_id`, `domains`, or `common_name` so there is a `CustomizeDiff` present to trigger a recreation of the subscription if anything else changes. Therefore the `PATCH` endpoint is only used if the changes are contained to these three fields.


fastly/resource_fastly_tls_subscription_test.go
```diff
@@ -22,14 +22,15 @@ func init() {
...
+	domain2 := fmt.Sprintf("%sALT.test", name)
 
 	resourceName := "fastly_tls_subscription.subject"
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccResourceFastlyTLSSubscriptionConfig(name, domain),
+				Config: testAccResourceFastlyTLSSubscriptionConfig(name, domain, domain2, domain),
```
Integralist: I'd prefer more descriptive variables to be passed into these functions (e.g. have them align with the actual signature defined). 

```suggestion
				Config: testAccResourceFastlyTLSSubscriptionConfig(name, domain1, domain2, commonName),
```

I appreciate that it's going to result in a semi-pointless variable assigned (e.g. `commonName = domain`) but I just feel it's much easier to understand what's going on at a glance without having to look at the terraform config function itself.

> Note: I presume we'd also have (based on my other comments): `commonName1 = domain1` and `commonName2 = domain2` as the next test step passes `domain2` as the 'common name'.

fastly/resource_fastly_tls_subscription_test.go
```diff
@@ -22,14 +22,15 @@ func init() {
 func TestAccResourceFastlyTLSSubscription(t *testing.T) {
 	name := acctest.RandomWithPrefix(testResourcePrefix)
 	domain := fmt.Sprintf("%s.test", name)
```
Integralist: Could we rename this for the sake of my OCD 🙂 (and also update the `testAccResourceFastlyTLSSubscriptionConfig` to use `domain1` as well please)
```suggestion
	domain1 := fmt.Sprintf("%s.test", name)
```

Integralist (commented):
@bengesoff LGTM. Just a couple of minor comments.

bengesoff:
@Integralist fixed based on your feedback, and confirmed the test still passes 🙂

PR #371, opened 2021-02-18
Test speedup

bengesoff:
Mostly a find/replace of `resource.Test` to `resource.ParallelTest` with a couple of exceptions:

- **`TestAccFastlyServiceV1_splunk_env`** and **`TestAccFastlyServiceV1_syslog_useTls`** have to stay serial because they set/reset environment variables within the test to exercise the `schema.EnvDefaultFunc`, which interferes with the other tests
- **`TestAccFastlyServiceWAFVersionV1Validation`** had to do `t.Parallel` directly because `resource.Test` was getting called in a loop as part of a table driven test

Total test time reduced from ~2:13:00 to ~00:21:00 (sample size of 1, `GOMAXPROCS=8`). This is excluding the running of the kinesis logging tests which would add a few more minutes. They could potentially add some unforeseen test failures which need fixing too, but it's unlikely as it doesn't look like they do anything with env variables etc

### DIFF
--- a/fastly/block_fastly_waf_configuration_v1_exclusion_test.go
+++ b/fastly/block_fastly_waf_configuration_v1_exclusion_test.go
@@ -92,8 +92,9 @@ func TestAccFastlyServiceWAFVersionV1FlattenWAFRuleExclusions(t *testing.T) {
 }
 
 func TestAccFastlyServiceWAFVersionV1Validation(t *testing.T) {
-	// Must set test to parallel outside of resource.ParallelTest due to the for loop. Otherwise t.Parallel() would be
-	// called multiple times, leading to a panic
+	// As we use a 'table test' which executes a `resource.Test` multiple times within a for-loop, we don't utilise the
+	// `resource.ParallelTest` function but instead call t.Parallel(). The use of t.Parallel() must happen outside of 
+	// the for-loop otherwise it would be executed multiple times, leading to a runtime panic.
 	t.Parallel()
 
 	name := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
--- a/fastly/block_fastly_waf_configuration_v1_exclusion_test.go
+++ b/fastly/block_fastly_waf_configuration_v1_exclusion_test.go
@@ -92,6 +92,10 @@ func TestAccFastlyServiceWAFVersionV1FlattenWAFRuleExclusions(t *testing.T) {
 }
 
 func TestAccFastlyServiceWAFVersionV1Validation(t *testing.T) {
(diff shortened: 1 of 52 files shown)


Integralist:
I ran these tests and had a lot of `429 Too Many Requests` errors (as well as errors related to TLS resource creation, which my account wasn't allowed to run without upgrading it -- i.e. I'd had to talk to the TLS team). There was also a few errors related to TLS queries coming back empty, which I presume was also related to my account not having access to TLS.

bengesoff:
It could be an idea to limit the parallelism by default in the Makefile incarnation of the tests, but to allow this to be increased for those who can (either with access to sufficiently configured accounts, or with sufficient bandwidth)

fastly/block_fastly_waf_configuration_v1_exclusion_test.go
```diff
@@ -92,6 +92,10 @@ func TestAccFastlyServiceWAFVersionV1FlattenWAFRuleExclusions(t *testing.T) {
 }
 
 func TestAccFastlyServiceWAFVersionV1Validation(t *testing.T) {
+	// Must set test to parallel outside of resource.ParallelTest due to the for loop. Otherwise t.Parallel() would be
+	// called multiple times, leading to a panic
```
Integralist:
```suggestion
	// As we use a 'table test' which executes a `resource.Test` multiple times within a for-loop, we don't utilise the
	// `resource.ParallelTest` function but instead call t.Parallel(). The use of t.Parallel() must happen outside of 
	// the for-loop otherwise it would be executed multiple times, leading to a runtime panic.
```

Integralist (approved):
@bengesoff Code LGTM.

**Note**: I ran this code locally and (after increasing rate limits on my test account and activating Platform TLS access) I found the test suite ran in 24mins! 🎉 That said there were some failures but they look to be related to there still being some outstanding Platform TLS feature flags needing to be activated by the TLS team.

Just two things I'd like to see implemented before merging (we discussed this offline):

1. Add a 'parallelism' flag to control GOMAXPROCS when running `go test`.
2. Add comments to the relevant tests that are staying serialised.

Integralist:
I'm reaching out to the relevant TLS team to get the full TLS feature set added to my testing account so I can confirm the full test suite is passing before I merge. Then the only remaining task is for @bengesoff to add the `parallelism` flag.

bengesoff:
I've gone with a test parallelism of 4 by default as it seems like a happy medium. Another approach could be to go with a parallelism of 1 by default to cater for the worst case, and to allow people to tune this higher if they need the performance gain, with the awareness that it could affect timeouts etc?

PR #372, opened 2021-02-24
Add force_destroy flag to ACLs and Dicts to allow deleting non-empty lists

bengesoff:
Updates service ACL and Dictionary blocks with extra protection to prevent the deletion of populated data that is dynamically added outside of a Terraform workflow.

Checks for non-empty ACL entries and Dictionary items before deleting ACLs and Dictionaries to reduce potential loss when updating either collection name.

Returns an error to the terraform user if `force_destroy` is not set on the ACL/Dictionary prior to updating.

### DIFF
--- a/fastly/block_fastly_service_v1_acl_test.go
+++ b/fastly/block_fastly_service_v1_acl_test.go
@@ -78,7 +78,7 @@ func TestAccFastlyServiceV1_acl(t *testing.T) {
 			},
 			{
 				Config: testAccServiceV1Config_acl(name, aclNameUpdated),
-				Check:  testAccAddACLEntries(&acl),
+				Check:  testAccAddACLEntries(&acl), // triggers side-effect of adding an ACL Entry
 			},
 			{
 				Config:      testAccServiceV1Config_acl(name, aclName),
--- a/fastly/block_fastly_service_v1_acl_test.go
+++ b/fastly/block_fastly_service_v1_acl_test.go
@@ -3,6 +3,7 @@ package fastly
 import (
 	"fmt"
 	"reflect"
+	"regexp"
 	"testing"
 
 	gofastly "github.com/fastly/go-fastly/v3/fastly"
@@ -42,10 +43,18 @@ func TestResourceFastlyFlattenAcl(t *testing.T) {
 
 func TestAccFastlyServiceV1_acl(t *testing.T) {
 	var service gofastly.ServiceDetail
-	name := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	var acl gofastly.ACL
+	name := acctest.RandomWithPrefix(testResourcePrefix)
 	aclName := fmt.Sprintf("acl_%s", acctest.RandString(10))
 	aclNameUpdated := fmt.Sprintf("acl_updated_%s", acctest.RandString(10))
 
+	// Six part test:
+	// 1. Create service with 2 ACLs
+	// 2. Rename both the ACLs, should succeed because the ACLs are empty
+	// 3. Keep both ACLs the same and add an entry to one of them
+	// 4. Try to rename the ACLs, expect to fail with "list not empty error"
+	// 5. Without renaming the ACLs, set force_destroy=true to skip the deletion check
+	// 6. Try to rename the ACLs again, expect to succeed
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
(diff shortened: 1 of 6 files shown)


Integralist (changes requested):
LGTM. But would be great to get a couple of tests in the relevant ACL/Dictionary test files to validate the new behaviour.

fastly/block_fastly_service_v1_acl_test.go
```diff
@@ -42,10 +43,18 @@ func TestResourceFastlyFlattenAcl(t *testing.T) {
...
 	aclName := fmt.Sprintf("acl_%s", acctest.RandString(10))
 	aclNameUpdated := fmt.Sprintf("acl_updated_%s", acctest.RandString(10))
 
+	// Six part test:
+	// 1. Create service with 2 ACLs
+	// 2. Rename both the ACLs, should succeed because the ACLs are empty
+	// 3. Keep both ACLs the same and add an entry to one of them
+	// 4. Try to rename the ACLs, expect to fail with "list not empty error"
+	// 5. Without renaming the ACLs, set force_destroy=true to skip the deletion check
+	// 6. Try to rename the ACLs again, expect to succeed
```
Integralist, line 57: ❤️  the descriptive breakdown here. This has been lacking in the majority of the tests 👍🏻

fastly/block_fastly_service_v1_acl_test.go
```diff
@@ -55,31 +64,55 @@ func TestAccFastlyServiceV1_acl(t *testing.T) {
...
 					testAccCheckServiceV1Exists("fastly_service_v1.foo", &service),
-					testAccCheckFastlyServiceV1Attributes_acl(&service, name, "a_"+aclNameUpdated),
-					testAccCheckFastlyServiceV1Attributes_acl(&service, name, "b_"+aclNameUpdated),
+					testAccCheckFastlyServiceV1Attributes_acl(&service, name, "a_"+aclNameUpdated, &acl),
+					testAccCheckFastlyServiceV1Attributes_acl(&service, name, "b_"+aclNameUpdated, &acl),
+				),
+			},
+			{
+				Config: testAccServiceV1Config_acl(name, aclNameUpdated),
+				Check:  testAccAddACLEntries(&acl),
```
Integralist:
```suggestion
				Check:  testAccAddACLEntries(&acl), // triggers side-effect of adding an ACL Entry
```

fastly/block_fastly_service_v1_dictionary_test.go
```diff
@@ -58,42 +68,47 @@ func TestAccFastlyServiceV1_dictionary(t *testing.T) {
...
+					testAccCheckFastlyServiceV1Attributes_dictionary(&service, &dictionary, name, updatedDictName, false),
+				),
+			},
+			{
+				Config: testAccServiceV1Config_dictionary(name, updatedDictName, backendName, domainName),
+				Check:  testAccAddDictionaryItems(&dictionary),
+			},
+			{
+				Config:      testAccServiceV1Config_dictionary(name, dictName, backendName, domainName),
+				ExpectError: regexp.MustCompile(""),
```
Integralist: Should we not get an error like we do with the ACL tests?

e.g.

```go
regexp.MustCompile("Cannot delete.*list is not empty.*")
```

fastly/block_fastly_service_v1_dictionary_test.go
```diff
@@ -58,42 +68,47 @@ func TestAccFastlyServiceV1_dictionary(t *testing.T) {
...
+				Config: testAccServiceV1Config_dictionary(name, updatedDictName, backendName, domainName),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckServiceV1Exists("fastly_service_v1.foo", &service),
-					testAccCheckFastlyServiceV1Attributes_dictionary(&service, name, dictName, true),
+					testAccCheckFastlyServiceV1Attributes_dictionary(&service, &dictionary, name, updatedDictName, false),
+				),
+			},
+			{
+				Config: testAccServiceV1Config_dictionary(name, updatedDictName, backendName, domainName),
+				Check:  testAccAddDictionaryItems(&dictionary),
```
Integralist:
```suggestion
				Check:  testAccAddDictionaryItems(&dictionary), // triggers side-effect of adding a Dictionary Item
```

Integralist (commented):
LGTM @bengesoff just a couple of comments.

fastly/block_fastly_service_v1_acl_test.go
bengesoff: Thanks! Thought it was especially necessary given the longer-form test

fastly/block_fastly_service_v1_dictionary_test.go
bengesoff: Oh that's a facepalm right there! Not sure where that's gone 😂

bengesoff: Rerunning the test with this added back in 🤦🏻‍♂️

Integralist (approved):

PR #373, opened 2021-02-24
Add service test sweeper

bengesoff:
Sometimes errors in acceptance tests can leave leaked services in the Fastly account. This sweeper adds support for cleaning these up with the `make sweep` command. This goes through the services and deletes the ones whose names start with `tf-test-` (the convention that the tests all use). The TLS sweepers work in the same way.


Integralist (approved):
LGTM. Thanks @bengesoff

PR #376, opened 2021-03-01
Better sensitive value handling in Google pub/sub logging provider

robwil:
This allows users to provide secret values for the Google pub/sub log integration, in a similar way that is already implemented for BigQuery and GCS (via an optional environment variable). Importantly, it also marks the private key as a sensitive value so it won't be logged during plan / apply.

cc @jw2340

### DIFF
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -411,9 +411,12 @@ Required:
 
 - **name** (String) The unique name of the Google Cloud Pub/Sub logging endpoint. It is important to note that changing this attribute will delete and recreate the resource
 - **project_id** (String) The ID of your Google Cloud Platform project
-- **secret_key** (String) Your Google Cloud Platform account secret key. The `private_key` field in your service account authentication JSON
 - **topic** (String) The Google Cloud Pub/Sub topic to which logs will be published
-- **user** (String) Your Google Cloud Platform service account email address. The `client_email` field in your service account authentication JSON
+
+Optional:
+
+- **secret_key** (String, Sensitive) Your Google Cloud Platform account secret key. The `private_key` field in your service account authentication JSON. You may optionally provide this secret via an environment variable, `FASTLY_GOOGLE_PUBSUB_SECRET_KEY`.
+- **user** (String) Your Google Cloud Platform service account email address. The `client_email` field in your service account authentication JSON. You may optionally provide this via an environment variable, `FASTLY_GOOGLE_PUBSUB_EMAIL`.
 
 
 <a id="nestedblock--logging_heroku"></a>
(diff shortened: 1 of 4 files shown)


docs/resources/service_compute.md
```diff
@@ -411,9 +411,12 @@ Required:
 
 - **name** (String) The unique name of the Google Cloud Pub/Sub logging endpoint. It is important to note that changing this attribute will delete and recreate the resource
 - **project_id** (String) The ID of your Google Cloud Platform project
-- **secret_key** (String) Your Google Cloud Platform account secret key. The `private_key` field in your service account authentication JSON
 - **topic** (String) The Google Cloud Pub/Sub topic to which logs will be published
-- **user** (String) Your Google Cloud Platform service account email address. The `client_email` field in your service account authentication JSON
+
+Optional:
```
Integralist, line 416: So this was actually quite interesting to me, so I raised a query with the official terraform documentation maintainers:

https://github.com/hashicorp/terraform-plugin-docs/issues/42

NOTE: this isn't a blocker to the PR getting approved/merged (as it's outside the control of both the contributor and ourselves as maintainers of this terraform provider), it's just something I noticed and thought I'd share.

Integralist (commented):
Thanks @robwil for opening this PR. 

The code looks fine to me. I left one comment which is an external conversation that I thought might be interesting to share. Otherwise it would be good if we could validate this change set with a test. Is that something you can add please.

docs/resources/service_compute.md
robwil: One thing worth noting here is that the field does in fact become optional from the perspective of the user, since if you don't explicitly pass it then it will default to `""` (empty string). In practice, we can leave `secret_key` out of the TF config block entirely.

But I do see your point that since it will always have a value, and is required, it is a little strange to call it "optional" in the documentation.

robwil:
The reason I left a test out is because none of the existing envvar overrides are covered with tests, so there wasn't a clear example to follow (as someone unfamiliar with the test setup here). But I can take a shot at it!

robwil:
I just added an attempt at a unit test. I didn't see too many other examples of this, but the way I did it seemed like the most straight-forward way to test that `secret_key` could receive a value from an envvar override.

fastly/block_fastly_service_v1_logging_googlepubsub_test.go
```diff
@@ -58,6 +60,85 @@ func TestResourceFastlyFlattenGooglePubSub(t *testing.T) {
...
+		t.Fatalf("Error matching:\nexpected: \"\"\ngot: %#v", result)
+	}
+
+	// Actually set env var and expect it to be used to determine user
+	envVarKey := "FASTLY_GOOGLE_PUBSUB_EMAIL"
+	mockValue := "example@my-project.iam.gserviceaccount.com"
+	originalEnvValue := os.Getenv(envVarKey)
+	defer func() {
+		os.Setenv(envVarKey, originalEnvValue)
+	}()
```
Integralist, line 88: I appreciate this detail 🙂

Integralist (approved):
This looks good to me. Thanks @robwil ❤️

PR #377, opened 2021-03-05
Fix panic caused by incorrect type assert.

Integralist:
**Problem**: the gzip resource was [recently migrated to using the `DiffSet` algorithm for determining changes](https://github.com/fastly/terraform-provider-fastly/pull/366), which introduced a bug where we incorrectly were type asserting a Set into a String which would cause a panic at runtime.

**Notes**: the tests didn't catch this because the pre-existing gzip tests were written in such a way that the 'update' test was actually causing the gzip resource to be deleted and recreated anew (because it was modifying the 'name' field) and so I've modified the tests to address this.

### DIFF
--- a/fastly/block_fastly_service_v1_gzip.go
+++ b/fastly/block_fastly_service_v1_gzip.go
@@ -127,10 +127,24 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 		// this and so we've updated the below code to convert the type asserted
 		// int into a uint before passing the value to gofastly.Uint().
 		if v, ok := modified["content_types"]; ok {
-			opts.ContentTypes = gofastly.String(v.(string))
+			set := v.(*schema.Set)
+			if len(set.List()) > 0 {
+				var s []string
+				for _, elem := range set.List() {
+					s = append(s, elem.(string))
+				}
+				opts.ContentTypes = gofastly.String(strings.Join(s, " "))
+			}
 		}
 		if v, ok := modified["extensions"]; ok {
-			opts.Extensions = gofastly.String(v.(string))
+			set := v.(*schema.Set)
+			if len(set.List()) > 0 {
+				var s []string
+				for _, elem := range set.List() {
+					s = append(s, elem.(string))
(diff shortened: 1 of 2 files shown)


fastly/block_fastly_service_v1_gzip.go
```diff
@@ -127,10 +127,24 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
...
 		// int into a uint before passing the value to gofastly.Uint().
 		if v, ok := modified["content_types"]; ok {
-			opts.ContentTypes = gofastly.String(v.(string))
+			set := v.(*schema.Set)
+			if len(set.List()) > 0 {
+				var s []string
+				for _, elem := range set.List() {
+					s = append(s, elem.(string))
+				}
+				opts.ContentTypes = gofastly.String(strings.Join(s, " "))
```
phamann, line 136: Noting here for reviewers, this API accepts the types as a space-seperated list, so this is correct :+1:
https://developer.fastly.com/reference/api/vcl-services/gzip/

phamann (approved):
👍🏻 LGTM

bridgetlane (approved):

fastly/block_fastly_service_v1_gzip_test.go
```diff
@@ -272,3 +291,32 @@ resource "fastly_service_v1" "foo" {
...
+    name    = "%s"
+    comment = "tf-testing-domain"
+  }
+
+  backend {
+    address = "aws.amazon.com"
+    name    = "amazon docs"
+  }
+
+  gzip {
```
Integralist, line 310: I want to clarify why this new test validates the actual 'update' mechanics...

The original test _thought_ it was validating a 'modification' to the gzip resource, but because they changed the _name_ field (along with other fields) it meant that the gzip resource was actually being 'deleted' and then 'recreated' so the 'modification' logic (which is where the panic happens) was never actually run.

To ensure the modification/update logic is executed I created a new test that updated the gzip resource but left the name the same and just modified some of the other fields.

PR #382, opened 2021-03-12
Allow developer overrides to enable Terraform to locate and use locally built provider

trentrosenbaum:
This PR is focused on enabling Terraform to discover a locally built provider binary through the developer overrides mechanism.  The PR introduces a new script to generate a overrides files that references a local provider binary.
Updates include

- Update Makefile to build binary into local bin directory
- Generate a developer overrides file alongside provider binary
- Returns environment variable line to instruct Terraform to use overrides file and binary
- Updates README.md to cover local build and execution.

### DIFF
--- a/scripts/generate-dev-overrides.sh
+++ b/scripts/generate-dev-overrides.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 BIN_DIR=$PWD/bin
+OVERRIDES_FILENAME=developer_overrides.tfrc
 
-cat << EOF > $BIN_DIR/developer_overrides.tfrc
+cat << EOF > $BIN_DIR/$OVERRIDES_FILENAME
 provider_installation {
 
   dev_overrides {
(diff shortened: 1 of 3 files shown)


scripts/generate-dev-overrides.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+BIN_DIR=$PWD/bin
+
+cat << EOF > $BIN_DIR/developer_overrides.tfrc
```
Integralist: I'd prefer to shift this name (i.e. `developer_overrides.tfrc`) into a variable as it's used three times.

Integralist (approved):
One small comment but otherwise LGTM.

scripts/generate-dev-overrides.sh
trentrosenbaum: Updated the script to use a variable when referencing the overrides filename.

Issue #179, opened 2019-07-09
Always invalidates and recreates backends


querry43:
Terraform is updating the backends of two fastly services every time it is run.  This behavior changed when upgrading from Terraform 0.11.14 to 0.12.3 and from the fastly provider version 0.5 to 0.8.  Additionally, I use terraform workspaces for different environments (dev, prod, etc) and this only happens in some workspaces.

```
      - backend {
          - address               = "my.address" -> null
          - auto_loadbalance      = false -> null
          - between_bytes_timeout = 10000 -> null
          - connect_timeout       = 10000 -> null
          - error_threshold       = 0 -> null
          - first_byte_timeout    = 50000 -> null
          - max_conn              = 500 -> null
          - name                  = "django-frontend" -> null
          - port                  = 443 -> null
          - ssl_cert_hostname     = "my-hostname" -> null
          - ssl_check_cert        = true -> null
          - use_ssl               = true -> null
          - weight                = 100 -> null
        }
      + backend {
          + address               = "my.address"
          + auto_loadbalance      = false
          + between_bytes_timeout = 10000
          + connect_timeout       = 10000
          + error_threshold       = 0
          + first_byte_timeout    = 50000
          + max_conn              = 500
          + name                  = "django-frontend"
          + port                  = 443
          + ssl_cert_hostname     = "my-hostname"
          + ssl_check_cert        = true
          + use_ssl               = true
          + weight                = 100
        }
```

### Terraform Version
```
Terraform v0.12.3
+ provider.aws v2.17.0
+ provider.fastly v0.8.0
+ provider.template v2.1.2
```

### Affected Resource(s)
- fastly_service_v1

### Expected Behavior
The fastly service should not change.

### Actual Behavior
The fastly service is being updated every apply.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. `terraform apply`

querry43:
I am not sure how to securely share configs and trace output (~300k lines).  If there are useful snippets, I can sanitize and post here.

querry43:
I think this could be related to ssl parameters.  Here is a complete backend config block after a few changes from my initial configuration.  In this case, the port is 80 so all ssl parameters should be false or null.

```
  backend {
    name = "java-backend-primary"

    address = var.java_backend_addresses[0]
    port    = var.java_backend_port

    healthcheck        = "Backend Healthcheck"
    max_conn           = 400
    connect_timeout    = 15000
[...]

PR #385, opened 2021-03-25
Misleading diff bug, output variable refresh bug, support for plugin debug mode

bengesoff:
A few changes in one PR here (the first two are small):

- **Plugin debug mode:** Added a CLI flag to `main.go` to support `--debug`. This allows the provider to be run in its own process outside of `terraform` itself. The provider can then be run with a debugger with breakpoints etc
- **Misleading diff bug:** Added a `Default: ""` to the `override_host` attribute in the backend block. In my testing it appeared that this removed the weird bug where all of the backends were marked as recreated in the diff. For example this issue https://github.com/fastly/terraform-provider-fastly/issues/179. This was being caused by the internal TypeSet diffing logic and the shimming between the legacy flatmap typesystem in the `terraform-plugin-sdk` and the newer `cty`-based typesystem that Terraform Core now uses. When not set, the proposed state from the config was being set to `nil`, but the prior state was being shimmed to `""` and the set hash was therefore different. 
- **Output variable refresh bug:** I added a `CustomizeDiff` function to the service resource to mark the `cloned_version` and `active_version` attributes as recomputed when any of the other attributes change (with the exception of `name`, `comment` and `version_comment`). This solves the problem where an output variable referencing these attributes wouldn't be updated after an `apply` and would require another `apply` or `refresh` for this to propagate through.
- Removed `StateFunc`s from all of the logging blocks as these were causing unnecessary whitespace diffs for the `CustomizeDiff`s and forcing the tests to fail with non-empty plans (the version was being marked as recomputed). I couldn't find any adverse effects from doing this, but would appreciate some confirmation here as there were a lot of comments about a known issue and there might have been a good reason that they were added in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,22 +44,47 @@ Alongside the newly built binary a file called `developer_overrides.tfrc` will b
 back details for setting the `TF_CLI_CONFIG_FILE` environment variable that will enable Terraform to use your locally built provider binary.
 * HashiCorp - [Development Overrides for Provider developers](https://www.terraform.io/docs/cli/config/config-file.html#development-overrides-for-provider-developers). 
 
-Alternatively, you can run the provider directly and tell Terraform how to communicate it with an environment variable.
-This is useful for attaching a debugger like [delve](https://github.com/go-delve/delve) to the provider.
-To do this, make sure the provider is built and present in the local `bin` directory (see above).
-Then, run the provider executable directly with the `--debug` flag.
+### Debugging the provider
+
+The previous method with `dev_overrides` should be sufficient for most development use including testing your local changes with actual Terraform code.
+However, sometimes it can be helpful to run the provider in debug mode if you need to attach a debugger like [delve](https://github.com/go-delve/delve) to solve a particular issue.
+
+The way Terraform normally works is that it starts the provider in a subprocess and connects to it using GRPC over a local socket.
+(For more information on this, see [hashicorp/go-plugin](https://github.com/hashicorp/go-plugin#architecture)).
+For debugging, it is possible to bypass this and execute the provider in a separate process, then tell Terraform how to communicate with it.
+The benefit of this is that the provider can be launched with a debugger attached in the same way that the debugger would attach to any normal executable.
+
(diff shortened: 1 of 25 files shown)


README.md
```diff
@@ -44,6 +44,23 @@ Alongside the newly built binary a file called `developer_overrides.tfrc` will b
 back details for setting the `TF_CLI_CONFIG_FILE` environment variable that will enable Terraform to use your locally built provider binary.
 * HashiCorp - [Development Overrides for Provider developers](https://www.terraform.io/docs/cli/config/config-file.html#development-overrides-for-provider-developers). 
 
+Alternatively, you can run the provider directly and tell Terraform how to communicate it with an environment variable.
+This is useful for attaching a debugger like [delve](https://github.com/go-delve/delve) to the provider.
+To do this, make sure the provider is built and present in the local `bin` directory (see above).
+Then, run the provider executable directly with the `--debug` flag.
```
Integralist: Couple of questions about this:

1. Is this preferred over configuring `dev_overrides` and if so, why?
2. I've only really used the GUI version of delve so how does the TUI version use `TF_REATTACH_PROVIDERS`?

bengesoff: To your first question: I wouldn't say it's preferred, just that they have different purposes. The `dev_overrides` method is probably best for general purpose testing and exercising of the plugin as it more closely matches typical Terraform use. This second method is good for debugging mainly (hard to do otherwise). There is some overlap though, so maybe it's confusing to mention them together? We could add a `### Debugging`-type section to make it clear that `dev_overrides` is the primary mechanism?

I probably haven't explained this well enough, maybe I should include commands actually using delve to make it clearer? Basically you run the plugin through delve completely normally (`dlv exec ./bin/tf-prov-fastly -- --debug` or something), which will output some logs to stdout, including the line with the env var. You then copy paste it into another shell, export it, and run `terraform apply` completely normally. The `terraform` binary itself will read the env var and go find the unix socket itself instead of running the plugin in a subprocess as it normally would. Does that make more sense? I can get it working with delve TUI and update the instructions to reflect that if it helps

Integralist:
@bengesoff the docs are much better now, thank you!

PR #386, opened 2021-03-26
Add debug mode

bengesoff:
Allows provider to be run in debugger in separate process, and to allow Terraform to connect to it. The README changes should explain it better than this description


Integralist (approved):
This is a fantastic addition. Thanks @bengesoff

PR #387, opened 2021-03-26
Fix misleading diff bug for backends

bengesoff:
Added a `Default: ""` to the `override_host` attribute in the backend block. In my testing it appeared that this removed the weird bug where all of the backends were marked as recreated in the diff. For example this issue https://github.com/fastly/terraform-provider-fastly/issues/179. This was being caused by the internal TypeSet diffing logic and the shimming between the legacy flatmap typesystem in the `terraform-plugin-sdk` and the newer `cty`-based typesystem that Terraform Core now uses. When not set, the proposed state from the config was being set to `nil`, but the prior state was being shimmed to `""` and the set hash was therefore different.

### DIFF
--- a/fastly/block_fastly_service_v1_backend.go
+++ b/fastly/block_fastly_service_v1_backend.go
@@ -318,6 +318,7 @@ func (h *BackendServiceAttributeHandler) Register(s *schema.Resource) error {
 		"override_host": {
 			Type:        schema.TypeString,
 			Optional:    true,
+			Default:     "",
 			Description: "The hostname to override the Host header",
 		},
 		"shield": {


Integralist (commented):
@bengesoff so this didn't seem to work for me. Here are the steps I took to validate the PR...

1. Clone your fork of the provider.
2. checkout the relevant PR branch.
3. `make build`
4. `export TF_CLI_CONFIG_FILE ...` in new shell where I have some HCL code.
5. `terraform apply` initial code that defines a backend.
6. `terraform apply` change to code that updates the backend name.

It's at that point the diff shows...

```tf
  # fastly_service_v1.test_service will be updated in-place
  ~ resource "fastly_service_v1" "test_service" {
        id             = "5bzeLrJAO46eKffLudHsLS"
        name           = "Misleading Backend Diff"
        # (5 unchanged attributes hidden)

      - backend {
          - address               = "httpbin.org" -> null
          - auto_loadbalance      = true -> null
          - between_bytes_timeout = 10000 -> null
          - connect_timeout       = 1000 -> null
          - error_threshold       = 0 -> null
          - first_byte_timeout    = 15000 -> null
          - max_conn              = 200 -> null
          - name                  = "httpbin" -> null
          - port                  = 80 -> null
          - ssl_check_cert        = true -> null
          - use_ssl               = false -> null
          - weight                = 100 -> null
        }
      + backend {
          + address               = "httpbin.org"
          + auto_loadbalance      = true
          + between_bytes_timeout = 10000
          + connect_timeout       = 1000
          + error_threshold       = 0
          + first_byte_timeout    = 15000
          + max_conn              = 200
          + name                  = "httpbin--updated"
          + port                  = 80
          + ssl_check_cert        = true
          + use_ssl               = false
          + weight                = 100
        }

        # (1 unchanged block hidden)
    }
```

Integralist (approved):

PR #388, opened 2021-03-26
Output variable refresh bug

bengesoff:
- **Output variable refresh bug:** I added a `CustomizeDiff` function to the service resource to mark the `cloned_version` and `active_version` attributes as recomputed when any of the other attributes change (with the exception of `name`, `comment` and `version_comment`). This solves the problem where an output variable referencing these attributes wouldn't be updated after an `apply` and would require another `apply` or `refresh` for this to propagate through.
- Removed `StateFunc`s from all of the logging blocks as these were causing unnecessary whitespace diffs for the `CustomizeDiff`s and forcing the tests to fail with non-empty plans (the version was being marked as recomputed). I couldn't find any adverse effects from doing this, but would appreciate some confirmation here as there were a lot of comments about a known issue and there might have been a good reason that they were added in.

### DIFF
--- a/fastly/base_fastly_service_v1.go
+++ b/fastly/base_fastly_service_v1.go
@@ -6,9 +6,9 @@ import (
 	"log"
 	"time"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-
 	gofastly "github.com/fastly/go-fastly/v3/fastly"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
@@ -148,6 +148,25 @@ func resourceService(serviceDef ServiceDefinition) *schema.Resource {
 		UpdateContext: resourceUpdate(serviceDef),
 		DeleteContext: resourceDelete(serviceDef),
 		Importer:      resourceImport(serviceDef),
+		CustomizeDiff: customdiff.All(
+			customdiff.ComputedIf("cloned_version", func(_ context.Context, d *schema.ResourceDiff, _ interface{}) bool {
+				// If anything other than name, comment and version_comment has changed, the current version will be
+				// cloned in resourceServiceUpdate so set it as recomputed. These three fields can be updated without
+				// creating a new version
+				for _, changedKey := range d.GetChangedKeysPrefix("") {
+					if changedKey == "name" || changedKey == "comment" || changedKey == "version_comment" {
+						continue
+					}
+					return true
+				}
+				return false
+			}),
+			customdiff.ComputedIf("active_version", func(_ context.Context, d *schema.ResourceDiff, _ interface{}) bool {
+				// If cloned_version is recomputed and we are automatically activating new versions (controlled with the
+				// activate flag) then the active_version will be recomputed too.
+				return d.HasChange("cloned_version") && d.Get("activate") == true
+			}),
+		),
 
 		Schema: map[string]*schema.Schema{
 			"name": {
@@ -388,6 +407,15 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta int
 		// for their own attributes.
 		for _, a := range serviceDef.GetAttributeHandler() {
 			if a.MustProcess(d, initialVersion) {
+				// Check if the Update has been cancelled and return early if so
+				if err := ctx.Err(); err != nil {
+					if errors.Is(err, context.Canceled) {
+						return nil
+					}
+
+					return diag.FromErr(err)
+				}
+
 				if err := a.Process(d, latestVersion, conn); err != nil {
 					return diag.FromErr(err)
 				}
--- a/fastly/block_fastly_service_v1_bigquerylogging_test.go
+++ b/fastly/block_fastly_service_v1_bigquerylogging_test.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"reflect"
-	"strings"
 	"testing"
 
 	gofastly "github.com/fastly/go-fastly/v3/fastly"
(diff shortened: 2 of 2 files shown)


fastly/base_fastly_service_v1.go
```diff
@@ -147,6 +148,25 @@ func resourceService(serviceDef ServiceDefinition) *schema.Resource {
...
+						continue
+					}
+					return true
+				}
+				return false
+			}),
+			customdiff.ComputedIf("active_version", func(_ context.Context, d *schema.ResourceDiff, _ interface{}) bool {
+				// If cloned_version is recomputed and we are automatically activating new versions (controlled with the
+				// activate flag) then the active_version will be recomputed too.
+				return d.HasChange("cloned_version") && d.Get("activate") == true
```
bengesoff, line 167: Just had a thought here - might need to check `d.HasChange("activate")`, as this could change the active version too. (Could it? Or does there need to be another change for this to take effect?)

bengesoff:
Rebased onto master, just going to retest

fastly/base_fastly_service_v1.go
bengesoff: Nope, the service is only activated if there's a change in a nested block. If the activate block was outside of this condition then the following could be applicable but it isn't.

```diff
diff --git a/fastly/base_fastly_service_v1.go b/fastly/base_fastly_service_v1.go
index 3ca10b3f..9982cdaa 100644
--- a/fastly/base_fastly_service_v1.go
+++ b/fastly/base_fastly_service_v1.go
@@ -164,7 +164,10 @@ func resourceService(serviceDef ServiceDefinition) *schema.Resource {
 			customdiff.ComputedIf("active_version", func(_ context.Context, d *schema.ResourceDiff, _ interface{}) bool {
 				// If cloned_version is recomputed and we are automatically activating new versions (controlled with the
 				// activate flag) then the active_version will be recomputed too.
-				return d.HasChange("cloned_version") && d.Get("activate") == true
+				updatingAndActivating := d.HasChange("cloned_version") && d.Get("activate") == true
+				activatingExisting := d.Get("active_version") != d.Get("cloned_version") && d.HasChange("activate")
+
+				return updatingAndActivating || activatingExisting
 			}),
 		),
 ```

**Comment by bengesoff** [comment]
Full acceptance test suite passing for me ✅

**Review by Integralist — APPROVED** [review]
LGTM but going to request another Terraform SME who has historical context to take a look as well.

**Comment by mccurdyc** [comment]
Hmm it's hard to remember why we did this, trying to find context:

- https://github.com/fastly/terraform-provider-fastly/pull/268

**Comment by Integralist** [comment]
This was discussed internally.

@bengesoff will attempt to replicate issue with logging blocks, and if is unable to replicate we'll be OK proceeding with the understanding that this PR will include extra documentation to explain to users how to trim whitespace for their keys if this indeed does bubble back up as an issue in the future.

**Comment by bengesoff** [comment]
Update from my logging whitespace investigations:

As feared, the API handling of whitespace in the PEM blobs doesn’t appear to be consistent. It appears that at least Elasticsearch and Kafka logging both strip the whitespace off, whereas BigQuery and others preserve whatever is given to them. This means that the `trimSpaceStateFunc`  **does** have some effect, but only in a small proportion of the cases in which it’s used.

The result of this being that we essentially have three options:
- Stick with the fix of deleting the `StateFunc`s everywhere and document the need for `trimspace()` if the strings have whitespace
- Keep the `StateFunc`s in and close the PR, which will leave the output variable refresh bug at large (due to `CustomiseDiff` not being an option in this case)
- Pass on the inconsistency to the relevant team at Fastly responsible for these APIs and get the best of both worlds

**Comment by bengesoff** [comment]
In addition to the first bullet point of my previous comment: we could also augment the documentation with a `ValidateFunc` to ensure that `trimspace()` has been used, e.g.

```go
func validateStringTrimmed(value interface{}, p cty.Path) diag.Diagnostics {
	if value != strings.TrimSpace(value.(string)) {
		return diag.FromErr(p.NewErrorf("string must not contain trailing whitespace"))
	}

	return nil
}
```
This might be a good alternative to removing the `StateFunc` altogether as it will lead to a good error message.

Integralist:
Hi @bengesoff 

I was discussing this in more detail with @phamann and we came to the conclusion that regardless of what we do in the Terraform provider, the backend APIs need to be consistent by trimming excess whitespace.

Additionally Patrick wondered if we could maybe utilise `DiffSupressFunc` (instead of `ValidateFunc`) and have the diff function check a trim of the relevant cert/key attribute, i.e. it's comparing a trimmed version of the HCL input against a trimmed version of the state stored value, so we would only show a diff if there was actually a difference.

But again, we'd still need the data to be stored Fastly's side trimmed.

Let me know you thoughts.

bengesoff:
Hi @Integralist 

Just to clarify, you guys reckon the direction is to _always_ trim rather than _never_ trim the whitespace?

I did try the `DiffSuppressFunc` actually, with the same idea but unfortunately it won't work. It comes back to the `TypeSet` problem where TF thinks the whole block is being created/deleted 😞 certainly an attractive idea though as otherwise we'll still have to push the burden onto the user

Integralist:
@bengesoff OK no worries, thanks for that feedback.

re:

> always trim rather than never trim the whitespace?

The thinking here is that only the BigQuery and GCS endpoints are different, and so adding trimming to those endpoints is a smaller blast radius than the reverse which would be removing trimming from everywhere. If the APIs do this then it would mean we'd use your [ValidateFunc suggestion](https://github.com/fastly/terraform-provider-fastly/pull/388#issuecomment-820550316) and have this change as part of a v1 release as this change would be 'breaking' in the sense that we're now forcing the requirement to trim the content back onto users where before it was handled internally by the provider.

bengesoff:
That makes sense, fair enough! Are you waiting for a final decision on this still?

Integralist:
> That makes sense, fair enough! Are you waiting for a final decision on this still?

Yup! There's an internal ticket that I'm following to see how this progresses. I'll let you know whenever we have an update.

Integralist:
@bengesoff FYI - might not affect this PR but the `master` branch is now named `main`.

Integralist:
This PR should now be safe without having to wait for our internal API work (thanks to https://github.com/fastly/terraform-provider-fastly/pull/439).

Issue #239, opened 2020-05-26
Image Optimizer support


Sytten:
Unless I am missing something, the image optimizer is not configurable from terraform. I think it would be nice to support it.

chenrui333:
Unless I am wrong, I dont see the support in go-fastly lib. https://github.com/fastly/go-fastly/search?q=image&unscoped_q=image

Integralist:
Hi @Sytten 

I added a comment to the go-fastly API client here:
https://github.com/fastly/go-fastly/issues/191#issuecomment-1260967231

I have a similar question in relation to Terraform.

deg4uss3r:
We enable the default settings via https://github.com/fastly/terraform-provider-fastly/pull/832 we can close this now :D

PR #390, opened 2021-03-30
When updating service, clone the cloned_version instead of active_version

bengesoff:
## Problem
When the provider updates a Fastly service, it clones a version, makes the changes to that, then optionally activates it depending on the value of the `activate` attribute. When `activate` is `true`, it doesn't matter whether `cloned_version` or `active_version` is used as the source of this clone as they are both equal (unless something failed between cloning and activating). However, when `activate` is `false`, and the user is making incremental updates to a service via "draft" versions, the `cloned_version` could be further ahead of the `active_version`. The current behaviour in this case is to keep cloning the active version each time. However, the Terraform state will track the `cloned_version` due to this code in the service Read function:
```
	// If we are importing or `activate` is set to false, temporarily set the
	// service.ActiveVersion number to the latest version supplied via the get
	// service version details call. This is to ensure we still read all of the
	// state below.
        isInactive := d.Get("activate").(bool) == false
        if s.ActiveVersion.Number == 0 && isImport || isInactive {
	        s.ActiveVersion.Number = s.Version.Number
        }
```
Therefore the plan that Terraform makes for each intermediate version to get to the next is essentially based on an incorrect view of the state.
```
   TF State
┌────┬────┬────┐
│    │    │    │
│    ▼    ▼    ▼
1    2    3    4
│    ▲    ▲    ▲
│    │    │    │
├────┘    │    │
│         │    │
├─────────┘    │
│              │
└──────────────┘
   Fastly Cloning
```
I don't believe this would actually be an issue if we didn't have TypeSets involved behind the nested blocks, because a state refresh as part of a reapply should generate a correct plan to get the new version into the desired state. However, for reasons previously discussed, TypeSets are a fact of life for now. The fact that the `SetDiff` algorithm is then used to conditionally Update a block instead of a Delete/Create operation means that there are certain sequences of updates where the order of operations matters. 

The `backend` block is a good example of this as it will always use the `UpdateBackend` method in go-fastly unless the `name` attribute has changed. In this case, an unfortunate combination of updates and name changes can cause unexpected results.

## Solution
The solution is to use the `cloned_version` instead of the `active_version` to clone from when making updates, so that the new version is identical to the TF state/previous HCL instead of the potentially arbitrary active version.

This also exposed a problem with the ACL test where the domain/backend names were being recalculated on each test step so this is fixed too. There is a better description in d1f15f21f0d0d76bb4745845caa80068f05be36a.

## Implications
As the ACL test failure demonstrates, there is a behaviour change in this fix, and the previous behaviour could be being depended on. Therefore, on a technicality, this is a breaking change fix.

### DIFF
--- a/fastly/base_fastly_service_v1.go
+++ b/fastly/base_fastly_service_v1.go
@@ -234,7 +234,7 @@ func resourceRead(serviceDef ServiceDefinition) schema.ReadContextFunc {
 // while injecting the ServiceDefinition into the true Update functionality.
 func resourceUpdate(serviceDef ServiceDefinition) schema.UpdateContextFunc {
 	return func(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-		return resourceServiceUpdate(ctx, d, meta, serviceDef)
+		return resourceServiceUpdate(ctx, d, meta, serviceDef, false)
 	}
 }
 
@@ -278,11 +278,19 @@ func resourceServiceCreate(ctx context.Context, d *schema.ResourceData, meta int
 	}
 
 	d.SetId(service.ID)
-	return resourceServiceUpdate(ctx, d, meta, serviceDef)
+
+	// If the service was just created, there is an empty Version 1 available
+	// that is unlocked and can be updated.
+	err = d.Set("cloned_version", 1)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return resourceServiceUpdate(ctx, d, meta, serviceDef, true)
 }
 
 // resourceServiceUpdate provides service resource Update functionality.
-func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}, serviceDef ServiceDefinition) diag.Diagnostics {
+func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}, serviceDef ServiceDefinition, isCreate bool) diag.Diagnostics {
 	if err := validateVCLs(d); err != nil {
 		return diag.FromErr(err)
 	}
@@ -314,18 +322,11 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta int
 		}
 	}
 
-	// Update the active version's comment. No new version is required for this.
+	// Update the cloned version's comment. No new version is required for this.
 	if d.HasChange("version_comment") && !needsChange {
-		latestVersion := d.Get("active_version").(int)
-		if latestVersion == 0 {
-			// If the service was just created, there is an empty Version 1 available
-			// that is unlocked and can be updated.
-			latestVersion = 1
-		}
-
 		opts := gofastly.UpdateVersionInput{
 			ServiceID:      d.Id(),
-			ServiceVersion: latestVersion,
+			ServiceVersion: d.Get("cloned_version").(int),
 			Comment:        gofastly.String(d.Get("version_comment").(string)),
 		}
 
@@ -340,17 +341,11 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta int
 
 	if needsChange {
 		var latestVersion int
-		// If the service has not yet been activated then we do not need to clone a version as version 1 will be
-		// unlocked and we can apply changes to that instead
-		if d.Get("active_version") == 0 {
+		if isCreate {
 			initialVersion = true
 			// If the service was just created, there is an empty Version 1 available
 			// that is unlocked and can be updated.
 			latestVersion = 1
-			err := d.Set("cloned_version", latestVersion)
-			if err != nil {
-				return diag.FromErr(err)
-			}
 		} else {
 			latestVersion = d.Get("cloned_version").(int)
 			// Clone the latest version, giving us an unlocked version we can modify.
--- a/fastly/base_fastly_service_v1.go
+++ b/fastly/base_fastly_service_v1.go
@@ -346,6 +346,10 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta int
 			// If the service was just created, there is an empty Version 1 available
 			// that is unlocked and can be updated.
 			latestVersion = 1
+			err := d.Set("cloned_version", latestVersion)
+			if err != nil {
+				return diag.FromErr(err)
+			}
 		} else {
 			// Clone the latest version, giving us an unlocked version we can modify.
 			log.Printf("[DEBUG] Creating clone of version (%d) for updates", latestVersion)
@@ -474,10 +478,6 @@ func resourceServiceRead(ctx context.Context, d *schema.ResourceData, meta inter
 		return diag.FromErr(err)
 	}
 	err = d.Set("active_version", s.ActiveVersion.Number)
-	if err != nil {
-		return diag.FromErr(err)
-	}
-	err = d.Set("cloned_version", s.Version.Number)
 	if err != nil {
 		return diag.FromErr(err)
 	}
--- a/fastly/base_fastly_service_v1.go
+++ b/fastly/base_fastly_service_v1.go
@@ -182,12 +182,11 @@ func resourceService(serviceDef ServiceDefinition) *schema.Resource {
 			// gets set whenever Terraform detects changes and clones the currently
 			// activated version in order to modify it. Active Version and Cloned
 			// Version can be different if the Activate field is set to false in order
-			// to prevent the service from being activated. It is not used internally,
-			// but it is exported for users to see after running `terraform apply`.
+			// to prevent the service from being activated.
 			"cloned_version": {
 				Type:        schema.TypeInt,
 				Computed:    true,
-				Description: "The latest cloned version by the provider. The value gets only set after running `terraform apply`",
+				Description: "The latest cloned version by the provider",
 			},
 
 			"activate": {
@@ -207,7 +206,7 @@ func resourceService(serviceDef ServiceDefinition) *schema.Resource {
 
 	// This loops over all the attribute handlers in the service definition and calls Register.
 	// Register adds schema attributes to the overall schema for the resource. This allows each AttributeHandler to
-	// define it's own attributes while allowing the overall set to be composed.
+	// define its own attributes while allowing the overall set to be composed.
 	for _, a := range serviceDef.GetAttributeHandler() {
 		a.Register(s) // Mutates s, adding handler-specific schema items to the list.
 	}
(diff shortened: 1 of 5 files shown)


bengesoff:
I've since added changes to when `cloned_version` actually gets set across the Read and Update functions. The logic flow in these functions is already quite convoluted so I've tried not to do too much to refactor, other than fix the cloning logic. 

In making these changes there were quite a lot of knock-on test failures for different reasons. The most significant was the ACL test failure mentioned in the main PR description. This reared its head again and took a bit of digging to get to the bottom of but in the end was a fairly simple fix.

Integralist:
**UPDATE**: I've been testing this PR and there appears to be issues with the current implementation. 

To clarify, these are my testing steps: 

1. Create an "active" service version `1` with one backend.
2. Switch to `activate = false` and apply an update to the defined backend (i.e. changing the backend name).
3. Apply another change to the defined backend (again, making a name change).
4. In both cases, changing the name of the backend will cause it to be deleted and a new backend created.
5. There should always be one backend at the end of the apply.

The intention is to ensure that a draft version `2` is cloned from the active version `1`, while a draft version `3` should be cloned from `draft 2` and that the diffs between them are correct.

I've tested the provider in two separate ways:

1. Running this PR branch version as a normal provider (using `TF_CLI_CONFIG_FILE`).
2. Running this PR branch version in 'debug' mode so I can attach a debugger (such as delve) to it.

I initially noticed the diffs between draft 2 and 3 were not correct. This hinted at draft 3 not being a clone of draft 2 but being (incorrectly) cloned from the active version 1. I then was able to confirm this was the case.

Although on some occasions it has worked correctly (i.e. the draft 3 was correctly cloned from draft 2), so this suggests to me there is a race condition somewhere.

We're discussing this internally and doing further investigation.

bengesoff:
I'm going to push some more changes to this in a sec, as chatting through with @Integralist, we've identified yet another edge case that could cause weird cloning behaviour. In particular, if the user creates a service outside of Terraform, adds some versions to it, and doesn't activate it, then imports it to Terraform and starts making changes, the current behaviour will be to mutate version 1 until something gets activated. It might feel more natural to clone the latest version import and add changes on top of that instead.

I'm going to change this to add a `d.Set("cloned_version", 1)` in the Create function, and then remove the logic from the Update function which checks `active_version == 0` and overrides the cloning. This will mean that it will just repeatedly create new drafts for each change made, similarly to the current normal case where we always clone from `cloned_version` for every change. We will add some features in a later PR to mitigate the draft version spamming.

fastly/base_fastly_service_v1.go
```diff
@@ -340,13 +340,14 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta int
 	initialVersion := false
 
 	if needsChange {
-		latestVersion := d.Get("active_version").(int)
-		if latestVersion == 0 {
+		var latestVersion int
+		if isCreate {
```
Integralist, line 344: I don't think we need the `if` block now. Could we not just now rely on the `else` block which does...

```
latestVersion = d.Get("cloned_version").(int)
```

Because inside of `resourceServiceCreate` we now set `cloned_version`.

bengesoff: I thought this, but if we do then we end up cloning immediately after creation, if that makes sense? So on a normal creation we'd go: "create service, set cloned_version=1, call Update function, apply comment etc to version 1, clone, set cloned_version=2, make updates to nested blocks" all in a single `apply` and end up with two versions straight away

Integralist: Ok that makes sense. Thanks!

christophermancini (approved):
"The logic flow in these functions is already quite convoluted so I've tried not to do too much to refactor, other than fix the cloning logic."

Agree, there are several functions in here that deserve some attention to break them up into smaller, more testable functions. At any rate, your solution to the problem seems like a great choice and I didn't notice anything concerning with the code. Well done and thanks for getting this sorted.

bridgetlane (changes requested):
I have some concerns about this change. I can see the following problem:
As a Fastly customer, I often would troubleshoot on my own and with Fastly support by manually cloning a version of my service and making changes (or Fastly support would do this) and then look at the compiled TF. At all of these times, I would not want my state to be pulled from whatever new changes the cloned version has. What if this option was configurable, allowing the default behavior to stay the same but be overriden by customers who want to always pull from cloned?

I would be interested in making this setting configurable, so the user can decide what their state is tracking (active or cloned).

I also think this is a fundamentally breaking change to how our TF works and we should think about how we might communicate with customers about this. It's possible my concern is a niche one, and the majority of customers would be happy pulling state from cloned.

bengesoff:
@christophermancini - thanks for taking the time to look through! Hopefully we'll get these functions cleaned up in a future PR.

@bridgetlane - I think I understand your concerns, but please let me know if not. I believe you're saying that if you manually make changes and clone a new draft version outside of Terraform (eg. via the Fastly web UI), then you would not want to see this version pulled into Terraform. If that's the case then I don't believe this change should violate those expectations because:
- if `"activate"` is `true`, then the version would only be pulled if it was also activated, which it doesn't sound like it would be in the example workflow you set out
- if `"activate"` is `false` then it would only pull the state from whatever `"cloned_version"` is set to, which itself is only ever set during import, service creation, or while Terraform is cloning in response to some updates, and should therefore only reference versions that Terraform created (i.e. not versions that were cloned outside TF)

Does this seem right? In any case, thank you for your feedback, it's great to throw different workflows and perspectives at this! 👍🏼 please let me know if I haven't understood your comment correctly, or if I haven't explained this very well.

bridgetlane:
@bengesoff Thanks for the reply/explanation, and that actually does address my concern! Good stuff. Thanks for the work on this

bridgetlane (approved):

Integralist:
@phamann @mccurdyc @jbhanrahan just a heads up to say we've now had two separate reviewers give a 👍🏻 on this PR. I need to have a chat with the author separately about the `1.0.0` this PR is going to entail. In the mean time if y'all still want to review this and give feedback then there is still time.

Integralist (approved):

Integralist:
Discussed internally and it seems we're OK with these changes and that they should be safe within the 0.x.x release range.

Thanks @bengesoff

JiriKovar:
I think this change just broke our way of working :) (correct me if I'm wrong please).

What we do is that we allow our teams to make changes in Fastly services both manually (i.e. if something cannot wait, or there is no support in Terraform yet), and also through Terraform directly. It used to be OK as Terraform simply used the last active version, but it's not the case anymore. 

The behavior we now see is that when someone manually clones a version, makes changes and activates the version, Terraform ignores it, clones the last version known to Terraform and effectively overrides/removes the manual changes, which is a huge problem in case of APIs that are not supported in Terraform, like the [Image Optimization API](https://github.com/fastly/terraform-provider-fastly/issues/239) .

So now we have 3 options:
1. Revert to older version of the provider (which is a currently implemented quick fix for us)
2. Always import manually cloned versions to Terraform state (which is a problem, because not everybody authorized to Fastly is also authorized to touch the Terraform state, nor does she/he knows how to correctly do it)
3. Work around the issue by implementing an automated import of the last active version to the Terraform state (which would be our approach in case that you decide this scenario is not supported)

TIA

Integralist:
Hi @JiriKovar thanks for sharing this detailed context. I'm going to bring this back to the team to discuss. I'll feedback here once we've had some time to think about the problem. Thanks!

JiriKovar:
Thank you very much, @Integralist  !

Hopefully this might help you at least a bit:
* Versions used during the aforementioned incident:
  - `Terraform v0.15.1`
  - `- Using fastly/fastly v0.28.2`
* Here's a screenshot of history in Fastly UI: 
2021-05-07 10_57_18-Fastly_ Configure (deliver-assets-us-01)
* And thanks to creating snapshot of Terraform state after each apply, I can confirm that the relevant part of the terraform state was:

  - Before the terraform apply in the red rectangle: 

            "activate": true,
            "active_version": 56,
            "cloned_version": 56,
            "comment": "Managed by Terraform"
            
  - After the terraform apply in the red rectangle: 
 
            "activate": true,
            "active_version": 58,
            "cloned_version": 58,
            "comment": "Managed by Terraform"

Integralist:
Hi @JiriKovar apologies for the delay replying.

I was chatting with a colleague this morning who stated that IO can only be configured via VCL (see https://developer.fastly.com/reference/io/#enabling-image-optimization). And so other than running a manual Terraform import, the solution would be to not use the UI to update the VCL configuration for the IO service (as you can add/update VCL via Terraform).

Let me know if this helps or if I'm misunderstanding the issue. Thanks!

JiriKovar:
@Integralist Unfortunately that is not the case - IO can be configured via UI (yes, some parts need to be configured in VCL, but not everything) and for the record reconfiguration of these variables was the operation that was reverted in the aforementioned scenario: 
image

Integralist:
Thanks for the update @JiriKovar the extra information was appreciated. 

Would it be possible to open a customer support ticket (support@fastly.com, or via the Fastly Support Slack if you have access) and reference this issue so we can get our team to help dig into this more for you. Thanks!

JiriKovar:
I've created a request asking them to either contact you directly (since I've noticed you are a Fastly employee), or jump into this discussion if I should take a part. 

Thank you very much. 

For the record:

-  Terraform 0.15.4 has changed its way of working with changes done manually outside of Terraform, which might be helpful here: https://github.com/hashicorp/terraform/releases 

- I'm sorry to say this, but since we are using multiple Terraform providers, I noticed that all of them refreshes state from the "active version" (the version that is currently working) and then makes diffs and builds on it. No matter that it might have been changed manually (much like described in the release notes in the link above). Even when I tried, I cant find a way to describe the "last version known to Terraform" to match this general concept that spans at least over all other providers we use. IMHO the root of the problem is with the `activate: false` flag - I have not come across any other Terraform provider that does this - `terraform apply` command is an equivalent of saying "I want the this environment to work according to the current definition". And the `activate: false` both breaks this concept as well as brings complexity described above by @bengesoff . If the version applied by Terraform was always activated, then it would work the same as with other providers (i.e. AWS, Azure), which would make using the last active version a no-brainer, while disregarding manual changes that were not applied yet. Please consider not copying Fastly functionality that much and following general concepts of Terraform a bit more.

Integralist:
@JiriKovar we've merged a change that I'm going to be cutting into a new release shortly which should hopefully address the comments you've raised.

JiriKovar:
Sounds like exactly what we need. Thank you very much!

PR #398, opened 2021-04-15
Add support for file_max_bytes configuration for azure logging endpoint

kellymclaughlin:
Support file_max_bytes configuration option for azure blob storage logging endpoint. Follows from fastly/go-fastly#255.

### DIFF
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -202,6 +202,7 @@ Required:
 
 Optional:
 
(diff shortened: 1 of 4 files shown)


Integralist:
@kellymclaughlin you'll need to run `make generate-docs` to get the last CI check to pass.

Integralist (approved):
LGTM but just need the CI to pass, which requires running `make generate-docs`.

kellymclaughlin:
Thanks @Integralist! I forgot about `make generate-docs`, but took care of that now.

PR #399, opened 2021-04-15
deps(go): use go1.16

chenrui333:
Bump golang to 1.16


Integralist:
Closing in favour of #447

PR #400, opened 2021-04-16
Automatically opt in to `ignore_changes`-esque behaviour for versionless resources (ACL entries and dict items)

bengesoff:
The versionless resources (ACL entries and Dictionary items) are sometimes used in a way whereby they are "seeded" via Terraform and then updated/managed externally via the API or UI console. For this, the documentation suggests using `ignore_changes`, a built-in Terraform meta-argument that allows the user to specify fields to ignore and from which to allow the state to drift.

However, sometimes this isn't obvious or the user doesn't understand this suggestion until it is too late, and data ends up getting lost. This happens because the user makes changes elsewhere and doesn't use `ignore_changes`, so Terraform takes action to remove the state drift and deletes their changes. This data is then unrecoverable.

This PR addresses this by changing the default behaviour of these resources to ignore the changes, and only allows the "dangerous" behaviour (which is currently the default) by explicitly opting in with a `manage_entries` option.


Integralist:
If it's OK @bengesoff for you to run `make generate-docs` so at least CI shows all green 😉

bengesoff:
Haha yes sorry @Integralist, I was planning to do that later on but I forgot how unpleasant it must be for you to look at 😏 two secs

Integralist:
@bengesoff FYI - might not affect this PR but the `master` branch is now named `main`.

Integralist (approved):
LGTM. But this will be considered a breaking change and so we'll defer merging this until we're ready to release a v1.0.0 of the Terraform provider.

Integralist:
**Question**: what's the situation with using `ignore_changes` with this implementation? 

There will be customers who have `ignore_changes` set already and so I presume if a customer pulls the version of the provider with this change set, then the `ignore_changes` will just become a redundant directive that can be deleted but otherwise won't cause any adverse affects.

bengesoff:
Yeah it should just be redundant - the `ignore_changes` will still apply, and `manage_entries` will default to false. Of course if someone does not have `ignore_changes` set, the default will now be changing so they will see a difference

Hiieu:
The changes in this PR have been merged into the `v1` branch in the OC fork: https://github.com/opencredo/terraform-provider-fastly/pull/110.

Integralist:
Closing as per the above comment. This is because we're consolidating all changes destined for `v1.0.0` of the Fastly Terraform provider into the [opencredo/v1](https://github.com/opencredo/terraform-provider-fastly/tree/v1) branch and will then soon be merged into the Fastly [v1](https://github.com/fastly/terraform-provider-fastly/tree/v1) branch, which will at some near point in the future make its way into `main` for a new release to be cut.

PR #410, opened 2021-05-10
Add director support for compute resource

bridgetlane:
Directors now work with C@E services. Simple PR to add directors to the compute service definition.

### DIFF
--- a/fastly/block_fastly_service_v1_director_test.go
+++ b/fastly/block_fastly_service_v1_director_test.go
@@ -193,6 +193,86 @@ func TestAccFastlyServiceV1_directors_basic(t *testing.T) {
 	})
 }
 
+func TestAccFastlyServiceV1_directors_basic_compute(t *testing.T) {
+	var service gofastly.ServiceDetail
+	name := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	domainName1 := fmt.Sprintf("fastly-test.tf-%s.com", acctest.RandString(10))
+
+	createdDir1 := gofastly.Director{
+		ServiceVersion: 1,
+		Name:           "mydirector",
+		Type:           3,
+		Quorum:         75,
+		Capacity:       100,
+		Retries:        5,
+	}
+
+	updatedDir1 := gofastly.Director{
+		ServiceVersion: 1,
+		Name:           "mydirector",
+		Type:           4,
+		Quorum:         30,
+		Capacity:       25,
+		Retries:        10,
+	}
+
+	createdDir2 := gofastly.Director{
+		ServiceVersion: 1,
+		Name:           "unchangeddirector",
+		Type:           3,
+		Quorum:         75,
+		Capacity:       100,
+		Retries:        5,
+	}
+
+	// Updated director should be the same as the created ones
+	updatedDir2 := createdDir2
+
+	updatedDir3 := gofastly.Director{
+		ServiceVersion: 1,
+		Name:           "myotherdirector",
+		Type:           3,
+		Quorum:         75,
+		Capacity:       100,
+		Retries:        5,
+	}
+
+	resource.Test(t, resource.TestCase{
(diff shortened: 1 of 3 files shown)


fastly/block_fastly_service_v1_director_test.go
```diff
@@ -193,6 +193,86 @@ func TestAccFastlyServiceV1_directors_basic(t *testing.T) {
...
+						[]*gofastly.Director{&createdDir1, &createdDir2}),
+					resource.TestCheckResourceAttr(
+						"fastly_service_compute.foo", "name", name),
+					resource.TestCheckResourceAttr(
+						"fastly_service_compute.foo", "director.#", "2"),
+				),
+			},
+
+			{
+				Config: testAccServiceV1DirectorsComputeConfig_update(name, domainName1),
```
Integralist, line 261: Can we remove the use of `v1` in the name so the functions are a little more aligned with each other.

e.g. `testAccServiceDirectorsComputeConfig` doesn't say `V1`.

fastly/block_fastly_service_v1_director_test.go
```diff
@@ -193,6 +193,86 @@ func TestAccFastlyServiceV1_directors_basic(t *testing.T) {
 	})
 }
 
+func TestAccFastlyServiceV1_directors_basic_compute(t *testing.T) {
```
Integralist, line 200: I typically find when reading these sorts of tests, that I have to re-read them multiple times to be sure I understand what they're doing. 

When updating the Terraform provider, if I'm working on a particular test suite that is a little hazy in its clarity, I tend to add a doc string to clarify what the test is doing and validating. Would it be possible to do that here? For example...

```go
// This test validates that two directors are created successfully
// (createdDir1 and createdDir2), and in the next Terraform run the first
// director is updated (updatedDir1) while the second director is unchanged
// (updatedDir2) and a third director is added (updatedDir3).
```

This process also typically leads me to realise that the naming of the variables could be improved to make the implementation clearer. For example...

- `createdDir1` -> `dir1`
- `createdDir2` -> `dir2`
- `updatedDir1` -> `dir1Update`
- `updatedDir2` -> `<deleted>` (†)
- `updatedDir3` -> `dir3`

> † I _think_ `updatedDir2 := createdDir2` is redundant? I don't believe the test would fail if `dir2` was passed into `testAccCheckFastlyServiceV1DirectorsAttributes` instead of a new variable assigned a copy of the original struct.

fastly/block_fastly_service_v1_director_test.go
```diff
@@ -326,3 +406,106 @@ resource "fastly_service_v1" "foo" {
...
+
+  director {
+    name = "myotherdirector"
+    type = 3
+    backends = [ "origin x", "origin y" ]
+  }
+
+  package {
+    filename = "test_fixtures/package/valid.tar.gz"
+	source_code_hash = filesha512("test_fixtures/package/valid.tar.gz")
```
Integralist: Nit: looks like the indentation is a bit off?

Integralist (commented):
Left a couple of minor comments. I've presumed from the implementation that the director support in C@E is all internal as it appears we're fundamentally using the same Director API endpoints (via go-fastly).

fastly/block_fastly_service_v1_director_test.go
bridgetlane: Everything else has a V1 so I actually think that's a miss on my part, I'll add it. Thanks!

fastly/block_fastly_service_v1_director_test.go
bridgetlane: Great idea

bridgetlane: Updated, let me know if that covers it!

Integralist (approved):
LGTM. Thanks for making those tweaks @bridgetlane

PR #411, opened 2021-05-12
v0.30.0

Integralist:


triblondon (approved):
The markdown refs are a bit fragile.  I'd recommend using footnote links:

```
A thing that has a <sup>[note-1](#note1)</sup>

<a name="note1">note-1</a>: Footnote content goes here
```

PR #414, opened 2021-06-03
Update default initial value for health check.

derekmarcotte:
Takes into account the change where the VCL is lazy-loaded on nodes as
needed.  The lazy-load may cause the first couple of health check
attempts to fail, and can be enough to mark the backend as unhealthy
until enough successful checks occur to reach the threshold again.

It is highly recommended that the initial and threshold values be the
same for health checks going forward.


HawkSE:
I can't approve this, but I do want to suggest that this be changed.

Integralist:
Thanks @HawkSE I'll get this approved and merged.

Once merged I'll need to coordinate this change with a new Terraform release, which might not happen immediately as we usually like to bundle up a few different changes within a release.

PR #418, opened 2021-06-10
Only set cloned_version after the version has been successfully validated

### LINKED ISSUE #369: state out of sync when there's a draft version created after the active version
### Terraform Version
```
$ terraform -v
Terraform v0.14.6
+ provider registry.terraform.io/fastly/fastly v0.24.0
```

### Affected Resource(s)
- fastly_service_v1

```hcl
resource "fastly_service_v1" "bug_report" {
    # active_version = 1
    default_ttl    = 3600
    id             = "[redacted]"
    name           = "bug report"
    activate       = false

    backend {
        address               = "httpbin.org"
        auto_loadbalance      = false
        between_bytes_timeout = 10000
        connect_timeout       = 1000
        error_threshold       = 0
        first_byte_timeout    = 15000
        max_conn              = 200
        name                  = "Host 1"
        port                  = 443
        ssl_cert_hostname     = "httpbin.org"
        ssl_check_cert        = true
        ssl_sni_hostname      = "httpbin.org"
        use_ssl               = true
        weight                = 100
    }

    domain {
        name = "test-domain-1.ps-pantheon.com"
    }

    snippet {
        content  = <<-EOT
            if (req.url !~ "^/anything") {
              set req.url = "/anything" req.url;
            }
        EOT
        name     = "url rewrite"
        priority = 100
        type     = "recv"
    }
}

```

### Steps to Reproduce
1. Import an active Fastly service (TF file above is the initial active state, call it v1).
2. Update a few things, eg. add a domain and change a snippet.
3. Using `activate = false`, run `terraform apply` and confirm to create a new draft service (v2).
4. DO NOT activate v2.
5. Update the TF script again, eg. add another snippet.
6. Run `terraform apply` again.

Example of first edit (step 2):
```diff
     domain {
         name = "test-domain-1.ps-pantheon.com"
     }
+    domain {
+        name = "test-domain-2.ps-pantheon.com"
+    }
 
     snippet {
         content  = <<-EOT
-            if (req.url !~ "^/anything") {
-              set req.url = "/anything" req.url;
+            if (req.url !~ "^/(status|cookies)") {
+              if (req.url !~ "^/anything") {
+                set req.url = "/anything" req.url
+              }
             }
         EOT
         name     = "url rewrite"
```

Note the missing semicolon in the new snippet which will cause Fastly to error out when activating.

Screenshot of diff from v2 to v1:
Screen Shot 2021-02-18 at 10 27 14

Diff of `.tf` file after fixing error:
```diff
         content  = <<-EOT
             if (req.url !~ "^/(status|cookies)") {
               if (req.url !~ "^/anything") {
-                set req.url = "/anything" req.url
+                set req.url = "/anything" req.url;
               }
             }
         EOT
```

Diff of v3 and v1 from the Fastly UI:
Screen Shot 2021-02-18 at 10 30 33

Note that the new domain created on v2 no longer appears in the changeset.

`active_version` is `1` in `terraform.tfstate` (which is correct).

### Expected Behavior
When TF refreshes the state on a `plan` or `apply`, it should use v1 as the baseline and determine that the domain and two snippets need to be created. The changes that are applied on the new draft should be _all_ of the changes specified in the `.tf` file that are not present in the active version.

### Actual Behavior
Only the most recently added snippet is created. The diff of v3 to v1 in the Fastly UI is the delta of changes between v2 (not activated) and v3 and does not contain all of the changes requested by the `.tf` file.

I suspect based on [these lines](https://github.com/fastly/terraform-provider-fastly/blob/8f24204024233504fa9efb2099f11374885ca9eb/fastly/base_fastly_service_v1.go#L456-L463) that the presence of an unactivated draft means that the provider will fall back to the draft when calculating a diff between the `.tf` file and the current state. The problem is that _isn't_ the current state; the real current state that needs to be the baseline is the active version.

### Potential Workarounds
1. If a draft service is created but not activated and another TF push is required (eg. there was a typo in the draft service [v2]), clone the active version (v1) into a new version (v3), activate v3, and then push a new draft (v4) with the fixes. Not ideal because you end up with a draft service you should not activate every time this happens.
2. (not personally tested but I've heard this works) Set `active_version = x` in the `.tf` file where `x` is the actual active version of the config, and most important keep it up-to-date. The downside to doing this is that the `.tf` file is not automatically updated after a successful push, and if left committed in the repo you run the risk of calculating a diff against an out-of-date config, especially if someone goes into the Fastly UI and makes manual changes.

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- #358 may be related

bengesoff:
Fixes #369.

Also adds the test case which previously failed with a 409 error prior to this fix (do we want to keep this?)

Beforehand, if there was an error midway through an update, and `activate` was true, there would be a half-updated draft version left on the service, which wouldn't be activated. The next apply would then read the active version but clone this half-updated version and try to apply the plan to it. This could cause a 409 error in some cases as the provider would be applying changes to a different version from which it had read.

This issue only applied when `activate` was true because, when `activate` was false, the provider would both read and clone from `cloned_version`, and could plan accordingly. The `activate=true` problem came from the false assumption that `cloned_version` and `active_version` would always be equal, but this isn't true when there's an error during the apply!

The fix for this is to only set `cloned_version` after the version validation has successfully passed.
This means that the next apply after an unsuccessful one will clone from the active version, as expected.

### DIFF
--- a/fastly/resource_fastly_service_v1_test.go
+++ b/fastly/resource_fastly_service_v1_test.go
@@ -551,6 +551,11 @@ func TestAccFastlyServiceV1_createDefaultTTL(t *testing.T) {
 	})
 }
 
+// TestAccFastlyServiceV1_brokenSnippet tests that a service can still be updated after it has failed during an apply.
+// This avoids a bug when activate=true, where setting an invalid snippet causes the resourceServiceUpdate function to
+// return early before activating the version. This broke the assumption that cloned_version always tracks the active
+// version when activate=true, and means that the version we read from, and the one we clone from in order to make changes,
+// are different, meaning the plan is applied to a different version and 409 conflict errors can occur.
 func TestAccFastlyServiceV1_brokenSnippet(t *testing.T) {
 	var service gofastly.ServiceDetail
 	name := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
--- a/fastly/resource_fastly_service_v1_test.go
+++ b/fastly/resource_fastly_service_v1_test.go
@@ -551,6 +551,42 @@ func TestAccFastlyServiceV1_createDefaultTTL(t *testing.T) {
 	})
 }
 
+func TestAccFastlyServiceV1_brokenSnippet(t *testing.T) {
+	var service gofastly.ServiceDetail
+	name := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	domain := fmt.Sprintf("fastly-test.tf-%s.test", acctest.RandString(10))
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccCheckServiceV1Destroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccServiceV1Config_brokenSnippet(name, domain, "backend1", `if (req.url !~ "^/anything") {
+                       set req.url = "/anything" req.url;
+                     }`),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckServiceV1Exists("fastly_service_v1.foo", &service),
+				),
+			},
+			{
+				Config: testAccServiceV1Config_brokenSnippet(name, domain, "backend2", `if (req.url !~ "^/anything") {
+                       set req.url = "/anything" req.url
+                     }`),
+				ExpectError: regexp.MustCompile(`Invalid configuration for Fastly Service`),
+			},
+			{
+				Config: testAccServiceV1Config_brokenSnippet(name, domain, "backend2", `if (req.url !~ "^/anything") {
+                       set req.url = "/anything" req.url;
+                     }`),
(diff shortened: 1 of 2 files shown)


fastly/resource_fastly_service_v1_test.go
```diff
@@ -551,6 +551,42 @@ func TestAccFastlyServiceV1_createDefaultTTL(t *testing.T) {
 	})
 }
 
+func TestAccFastlyServiceV1_brokenSnippet(t *testing.T) {
```
Integralist, line 559: @bengesoff Could add a doc string that just gives a summary of what the test is validating for future travellers.

Tests are written at a time when the full context of what is going on is still fresh, but if I was to come back to this test it would be hard for me to know what it was trying to validate and why.

So something like…

```go
// TestAccFastlyServiceV1_brokenSnippet validates X and it does this by doing Y first, then Z and ensuring ABC happens.
```

Integralist (approved):
LGTM. Thanks!

PR #432, opened 2021-06-25
Add stale if error settings

bridgetlane:
Adding settings here: https://developer.fastly.com/reference/api/vcl-services/settings/

Fixes #432


fastly/block_fastly_service_v1_settings.go
```diff
@@ -73,5 +82,17 @@ func (h *SettingsServiceAttributeHandler) Register(s *schema.Resource) error {
 		Computed:    true,
 		Description: "The default hostname",
 	}
+	s.Schema["stale_if_error"] = &schema.Schema{
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     false,
+		Description: "The default hostname",
```
Integralist: Need to update the description.

fastly/block_fastly_service_v1_settings.go
```diff
@@ -73,5 +82,17 @@ func (h *SettingsServiceAttributeHandler) Register(s *schema.Resource) error {
...
+	}
+	s.Schema["stale_if_error_ttl"] = &schema.Schema{
+		Type:        schema.TypeInt,
+		Optional:    true,
+		Default:     43200,
+		Description: "The default hostname",
```
Integralist: Need to update the description.

fastly/resource_fastly_service_compute_test.go
```diff
@@ -104,6 +104,10 @@ func TestAccFastlyServiceCompute_basic(t *testing.T) {
 					testAccCheckServiceV1Exists("fastly_service_compute.foo", &service),
 					resource.TestCheckResourceAttr(
 						"fastly_service_compute.foo", "name", name),
+					resource.TestCheckResourceAttr(
+						"fastly_service_compute.foo", "default_ttl", "0"),
+					resource.TestCheckResourceAttr(
+						"fastly_service_compute.foo", "stale_if_error", "true"),
```
Integralist, line 110: I presume we also want to validate that `stale_if_error_ttl` is set to zero by the configuration (as the default value of `43200 ` is being validated in the following new test).

Integralist: Oh I see you have another test for that (`TestAccFastlyServiceCompute_createZeroDefaultTTL`) but maybe worth still adding a check for the attribute here too?

bridgetlane:
Thank you for the review @Integralist! Updated accordingly

fastly/block_fastly_service_v1_settings.go
```diff
@@ -86,13 +86,13 @@ func (h *SettingsServiceAttributeHandler) Register(s *schema.Resource) error {
 		Type:        schema.TypeBool,
 		Optional:    true,
 		Default:     false,
-		Description: "The default hostname",
+		Description: "Whether or not stale if error should be true",
```
Integralist: I would prefer us to use the same language as on the developer reference for consistency/accuracy:

> Enables serving a stale object if there is an error

fastly/block_fastly_service_v1_settings.go
```diff
@@ -86,13 +86,13 @@ func (h *SettingsServiceAttributeHandler) Register(s *schema.Resource) error {
...
 	}
 	s.Schema["stale_if_error_ttl"] = &schema.Schema{
 		Type:        schema.TypeInt,
 		Optional:    true,
 		Default:     43200,
-		Description: "The default hostname",
+		Description: "The TTL for stale if error",
```
Integralist: I would prefer us to use the same language as on the developer reference for consistency/accuracy:

> The default time-to-live (TTL) for serving the stale object for the version

Integralist (commented):
Two small suggestions @bridgetlane

fastly/resource_fastly_service_v1_test.go
```diff
@@ -614,6 +614,18 @@ func TestAccFastlyServiceV1_defaultHost(t *testing.T) {
 						"fastly_service_v1.foo", "default_ttl", "3400"),
 					resource.TestCheckResourceAttr(
 						"fastly_service_v1.foo", "stale_if_error_ttl", "43200"),
+						"fastly_service_v1.foo", "default_host", defaultHost),
```
Integralist, line 618: @bridgetlane looks like a bad merge. I think you need an extra `resource.TestCheckResourceAttr(` added above this line.

Integralist (changes requested):

fastly/resource_fastly_service_v1_test.go
Integralist: that should hopefully fix all the other lint errors as I think they're all falling out from this first syntax error.

Integralist (approved):
This looks OK to me. @bridgetlane have you run the full test suite on this PR yet?

Integralist:
FYI I'm using go 1.17 and I needed to run `gofmt -w tools/tools.go` to update the build constraint flag so the latest format is supported:

```diff
+//go:build tools
 // +build tools
```

Integralist:
@bridgetlane @smaeda-ks 

Running the test suite I noticed: `TestAccFastlyServiceCompute_basic` is failing...

```
resource_fastly_service_compute_test.go:96: 
Step 1/2 error: Check failed: 
Check 5/11 error: fastly_service_compute.foo: 
Attribute ‘stale_if_error_ttl’ expected “43200”, got “0"
```

fastly/resource_fastly_service_compute_test.go
```diff
@@ -104,6 +104,12 @@ func TestAccFastlyServiceCompute_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(
+						"fastly_service_compute.foo", "stale_if_error_ttl", "43200"),
```
smaeda-ks:
```suggestion
						"fastly_service_compute.foo", "stale_if_error_ttl", "0"),
```

this should be 0

fastly/resource_fastly_service_v1_test.go
```diff
@@ -610,6 +610,10 @@ func TestAccFastlyServiceV1_defaultHost(t *testing.T) {
 				Config: testAccServiceV1Config_default_host(name, domain, defaultHost),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckServiceV1Exists("fastly_service_v1.foo", &service),
+					resource.TestCheckResourceAttr(
+						"fastly_service_v1.foo", "default_ttl", "3400"),
```
smaeda-ks:
```suggestion
						"fastly_service_v1.foo", "default_ttl", "3600"),
```

our default is 3600, not 3400 (considering `testAccServiceV1Config_default_host` config is not setting `default_ttl`)

fastly/block_fastly_service_v1_settings.go
```diff
@@ -25,6 +25,17 @@ func (h *SettingsServiceAttributeHandler) Process(d *schema.ResourceData, latest
 		DefaultTTL: uint(d.Get("default_ttl").(int)),
 	}
 
+	val := uint(d.Get("stale_if_error_ttl").(int))
+	opts.StaleIfErrorTTL = &val
```
smaeda-ks:
```suggestion
```

fastly/block_fastly_service_v1_settings.go
```diff
@@ -25,6 +25,17 @@ func (h *SettingsServiceAttributeHandler) Process(d *schema.ResourceData, latest
 		DefaultTTL: uint(d.Get("default_ttl").(int)),
```
smaeda-ks:
```suggestion
		DefaultTTL:      uint(d.Get("default_ttl").(int)),
		StaleIfErrorTTL: gofastly.Uint(uint(d.Get("stale_if_error_ttl").(int))),
```

smaeda-ks: Can you also update the comment accordingly?
https://github.com/fastly/terraform-provider-fastly/blob/v0.35.0/fastly/block_fastly_service_v1_settings.go#L24-L25
something like:

```
		// We have the same default value that is provided by
		// the Fastly API, so it's safe to include here
```

smaeda-ks (changes requested):
@bridgetlane or @Integralist I've added some suggestions including fixes for test cases. Since I don't have write access to this repo I can't add commits. Can either of you review and apply those suggestions? Then we should be able to merge this. Thanks

Integralist:
To help expedite these changes I'm going to close this PR and it will be replaced by...
https://github.com/fastly/terraform-provider-fastly/pull/475

Both Bridget and Shohei are credited as the authors of the work respectively.

PR #433, opened 2021-06-26
use TypeList schema for gzip block attributes

smaeda-ks:
Unfortunately, the value of `extensions` and `content_types` parameters for Gzip endpoint is just a list of strings (space-separated):
https://developer.fastly.com/reference/api/vcl-services/gzip/#create-gzip-config

and since they are defined as `schema.TypeSet` in the provider, ordering is [not retained](https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset) when sending them in the API request:
https://github.com/fastly/terraform-provider-fastly/blob/v0.32.0/fastly/block_fastly_service_v1_gzip.go#L196-L207

For example, this gzip block configuration:

```
  gzip {
    name          = "Generated by default gzip policy"
    extensions    = ["js", "css", "html", "eot", "ico", "otf", "ttf", "json", "svg"]
    content_types = ["text/html", "application/x-javascript", "text/css", "application/javascript", "text/javascript", "application/json", "application/vnd.ms-fontobject", "application/x-font-opentype", "application/x-font-truetype", "application/x-font-ttf", "application/xml", "font/eot", "font/opentype", "font/otf", "image/svg+xml", "image/vnd.microsoft.icon", "text/plain", "text/xml"]
  }
```

would produce this in terminal output:
Screen Shot 2021-06-26 at 16 36 57

and would result in this in the actual configuration:
Screen Shot 2021-06-26 at 16 37 56

Using `schema.TypeList` fixes this.
https://www.terraform.io/docs/extend/schemas/schema-types.html#typelist

That said, this will be a breaking change as the first run after the provider version update will modify the gzip object to change the string order.

While the current implementation causes no real issues (at least it produces the valid configuration from the service VCL point of view), retaining the ordering makes users much easier to review and compare the actual changes.

### DIFF
--- a/fastly/block_fastly_service_v1_gzip.go
+++ b/fastly/block_fastly_service_v1_gzip.go
@@ -78,21 +78,11 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 		}
 
 		if v, ok := resource["content_types"]; ok {
-			var cl []string
-			d := v.([]interface{})
-			for _, elem := range d {
-				cl = append(cl, elem.(string))
-			}
-			opts.ContentTypes = strings.Join(cl, " ")
+			opts.ContentTypes = sliceToString(v.([]interface{}))
 		}
 
 		if v, ok := resource["extensions"]; ok {
-			var el []string
-			d := v.([]interface{})
-			for _, elem := range d {
-				el = append(el, elem.(string))
-			}
-			opts.Extensions = strings.Join(el, " ")
+			opts.Extensions = sliceToString(v.([]interface{}))
 		}
 
 		log.Printf("[DEBUG] Fastly Gzip Addition opts: %#v", opts)
@@ -117,35 +107,13 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 		}
 
 		// NOTE: []interface{} is not comparable in Filter function
-		// covert it into string in advance and restore the original state after Filter
-		origContentTypes := make([]interface{}, len(resource["content_types"].([]interface{})))
-		origExtensions := make([]interface{}, len(resource["extensions"].([]interface{})))
-		copy(origContentTypes, resource["content_types"].([]interface{}))
-		copy(origExtensions, resource["extensions"].([]interface{}))
-
-		var cl []string
-		for _, elem := range resource["content_types"].([]interface{}) {
-			cl = append(cl, elem.(string))
-		}
-		resource["content_types"] = strings.Join(cl, " ")
-
-		var el []string
-		for _, elem := range resource["extensions"].([]interface{}) {
-			el = append(el, elem.(string))
-		}
-		resource["extensions"] = strings.Join(el, " ")
+		// covert it into string in advance
+		resource["content_types"] = sliceToString(resource["content_types"].([]interface{}))
+		resource["extensions"] = sliceToString(resource["extensions"].([]interface{}))
 
 		// only attempt to update attributes that have changed
 		modified := setDiff.Filter(resource, oldSet)
 
-		resource["content_types"] = origContentTypes
-		resource["extensions"] = origExtensions
-
-		// NOTE: where we transition between interface{} we lose the ability to
-		// infer the underlying type being either a uint vs an int. This
-		// materializes as a panic (yay) and so it's only at runtime we discover
-		// this and so we've updated the below code to convert the type asserted
-		// int into a uint before passing the value to gofastly.Uint().
 		if v, ok := modified["content_types"]; ok {
 			opts.ContentTypes = gofastly.String(v.(string))
 		}
@@ -238,7 +206,7 @@ func flattenGzips(gzipsList []*gofastly.Gzip) []map[string]interface{} {
 			for _, ev := range e {
 				et = append(et, ev)
 			}
-			ng["extensions"] = schema.NewSet(schema.HashString, et)
+			ng["extensions"] = et
 		}
 
 		if g.ContentTypes != "" {
@@ -247,7 +215,7 @@ func flattenGzips(gzipsList []*gofastly.Gzip) []map[string]interface{} {
 			for _, cv := range c {
 				ct = append(ct, cv)
 			}
-			ng["content_types"] = schema.NewSet(schema.HashString, ct)
+			ng["content_types"] = ct
 		}
 
 		// prune any empty values that come from the default string value in structs
@@ -261,4 +229,12 @@ func flattenGzips(gzipsList []*gofastly.Gzip) []map[string]interface{} {
 	}
 
 	return gl
+}
+
+func sliceToString(src []interface{}) string {
+	var result []string
+	for _, el := range src {
+		result = append(result, el.(string))
+	}
+	return strings.Join(result, " ")
 }
--- a/fastly/block_fastly_service_v1_gzip.go
+++ b/fastly/block_fastly_service_v1_gzip.go
@@ -78,23 +78,21 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 		}
 
 		if v, ok := resource["content_types"]; ok {
-			if len(v.(*schema.Set).List()) > 0 {
-				var cl []string
-				for _, c := range v.(*schema.Set).List() {
-					cl = append(cl, c.(string))
-				}
-				opts.ContentTypes = strings.Join(cl, " ")
+			var cl []string
+			d := v.([]interface{})
+			for _, elem := range d {
+				cl = append(cl, elem.(string))
 			}
+			opts.ContentTypes = strings.Join(cl, " ")
 		}
 
 		if v, ok := resource["extensions"]; ok {
-			if len(v.(*schema.Set).List()) > 0 {
-				var el []string
-				for _, e := range v.(*schema.Set).List() {
-					el = append(el, e.(string))
-				}
-				opts.Extensions = strings.Join(el, " ")
+			var el []string
+			d := v.([]interface{})
+			for _, elem := range d {
+				el = append(el, elem.(string))
 			}
+			opts.Extensions = strings.Join(el, " ")
 		}
 
 		log.Printf("[DEBUG] Fastly Gzip Addition opts: %#v", opts)
@@ -118,8 +116,30 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 			Name:           resource["name"].(string),
 		}
 
+		// NOTE: []interface{} is not comparable in Filter function
+		// covert it into string in advance and restore the original state after Filter
+		origContentTypes := make([]interface{}, len(resource["content_types"].([]interface{})))
+		origExtensions := make([]interface{}, len(resource["extensions"].([]interface{})))
+		copy(origContentTypes, resource["content_types"].([]interface{}))
+		copy(origExtensions, resource["extensions"].([]interface{}))
+
+		var cl []string
+		for _, elem := range resource["content_types"].([]interface{}) {
+			cl = append(cl, elem.(string))
+		}
+		resource["content_types"] = strings.Join(cl, " ")
+
+		var el []string
+		for _, elem := range resource["extensions"].([]interface{}) {
+			el = append(el, elem.(string))
+		}
+		resource["extensions"] = strings.Join(el, " ")
+
 		// only attempt to update attributes that have changed
 		modified := setDiff.Filter(resource, oldSet)
+
+		resource["content_types"] = origContentTypes
+		resource["extensions"] = origExtensions
 
 		// NOTE: where we transition between interface{} we lose the ability to
 		// infer the underlying type being either a uint vs an int. This
@@ -127,24 +147,10 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 		// this and so we've updated the below code to convert the type asserted
 		// int into a uint before passing the value to gofastly.Uint().
 		if v, ok := modified["content_types"]; ok {
-			set := v.(*schema.Set)
-			if len(set.List()) > 0 {
-				var s []string
-				for _, elem := range set.List() {
-					s = append(s, elem.(string))
-				}
-				opts.ContentTypes = gofastly.String(strings.Join(s, " "))
-			}
+			opts.ContentTypes = gofastly.String(v.(string))
 		}
 		if v, ok := modified["extensions"]; ok {
-			set := v.(*schema.Set)
-			if len(set.List()) > 0 {
-				var s []string
-				for _, elem := range set.List() {
-					s = append(s, elem.(string))
-				}
-				opts.Extensions = gofastly.String(strings.Join(s, " "))
-			}
+			opts.Extensions = gofastly.String(v.(string))
 		}
 		if v, ok := modified["cache_condition"]; ok {
 			opts.CacheCondition = gofastly.String(v.(string))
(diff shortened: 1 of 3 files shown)


bengesoff:
I don't think this passes the acceptance tests in its current form, and I think some of the reshuffling with the slices before `setDiff.Filter` isn't actually needed. GitHub wouldn't let me comment on certain lines in the diff so I've made a diff with some suggestions instead, hope that's ok! The summary of the changes below are as follows:

- The looping and string appending pattern is repeated a few times so could be abstracted into a `sliceToString` function to ease readability
- The copying of the slices before `setDiff.Filter` and restoring afterwards looks to be superfluous as the `resource` variable isn't used after it's been mutated, so this can be simplified
- The `flattenGzips` function needs to be updated because it still uses `schema.NewSet`, which causes the `Read` function to panic - and the unit test for this function can also be simplified given that it doesn't need to account for a TypeSet anymore
- The `TestAccFastlyServiceV1_gzips_basic` acceptance test needs updating because the order of the lists now matters

```diff
diff --git a/fastly/block_fastly_service_v1_gzip.go b/fastly/block_fastly_service_v1_gzip.go
index 7e93c4dd..c6f43e97 100644
--- a/fastly/block_fastly_service_v1_gzip.go
+++ b/fastly/block_fastly_service_v1_gzip.go
@@ -78,21 +78,11 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 		}
 
 		if v, ok := resource["content_types"]; ok {
-			var cl []string
-			d := v.([]interface{})
-			for _, elem := range d {
-				cl = append(cl, elem.(string))
-			}
-			opts.ContentTypes = strings.Join(cl, " ")
+			opts.ContentTypes = sliceToString(v.([]interface{}))
 		}
 
 		if v, ok := resource["extensions"]; ok {
-			var el []string
-			d := v.([]interface{})
-			for _, elem := range d {
-				el = append(el, elem.(string))
-			}
-			opts.Extensions = strings.Join(el, " ")
+			opts.Extensions = sliceToString(v.([]interface{}))
 		}
 
 		log.Printf("[DEBUG] Fastly Gzip Addition opts: %#v", opts)
@@ -118,34 +108,12 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 
 		// NOTE: []interface{} is not comparable in Filter function
 		// covert it into string in advance and restore the original state after Filter
-		origContentTypes := make([]interface{}, len(resource["content_types"].([]interface{})))
-		origExtensions := make([]interface{}, len(resource["extensions"].([]interface{})))
-		copy(origContentTypes, resource["content_types"].([]interface{}))
-		copy(origExtensions, resource["extensions"].([]interface{}))
-
-		var cl []string
-		for _, elem := range resource["content_types"].([]interface{}) {
-			cl = append(cl, elem.(string))
-		}
-		resource["content_types"] = strings.Join(cl, " ")
-
-		var el []string
-		for _, elem := range resource["extensions"].([]interface{}) {
-			el = append(el, elem.(string))
-		}
-		resource["extensions"] = strings.Join(el, " ")
+		resource["content_types"] = sliceToString(resource["content_types"].([]interface{}))
+		resource["extensions"] = sliceToString(resource["extensions"].([]interface{}))
 
 		// only attempt to update attributes that have changed
 		modified := setDiff.Filter(resource, oldSet)
 
-		resource["content_types"] = origContentTypes
-		resource["extensions"] = origExtensions
-
-		// NOTE: where we transition between interface{} we lose the ability to
-		// infer the underlying type being either a uint vs an int. This
-		// materializes as a panic (yay) and so it's only at runtime we discover
-		// this and so we've updated the below code to convert the type asserted
-		// int into a uint before passing the value to gofastly.Uint().
 		if v, ok := modified["content_types"]; ok {
 			opts.ContentTypes = gofastly.String(v.(string))
 		}
@@ -238,7 +206,7 @@ func flattenGzips(gzipsList []*gofastly.Gzip) []map[string]interface{} {
 			for _, ev := range e {
 				et = append(et, ev)
 			}
-			ng["extensions"] = schema.NewSet(schema.HashString, et)
+			ng["extensions"] = et
 		}
 
 		if g.ContentTypes != "" {
@@ -247,7 +215,7 @@ func flattenGzips(gzipsList []*gofastly.Gzip) []map[string]interface{} {
 			for _, cv := range c {
 				ct = append(ct, cv)
 			}
-			ng["content_types"] = schema.NewSet(schema.HashString, ct)
+			ng["content_types"] = ct
 		}
 
 		// prune any empty values that come from the default string value in structs
@@ -262,3 +230,11 @@ func flattenGzips(gzipsList []*gofastly.Gzip) []map[string]interface{} {
 
 	return gl
 }
+
+func sliceToString(src []interface{}) string {
+	var result []string
+	for _, el := range src {
+		result = append(result, el.(string))
+	}
+	return strings.Join(result, " ")
+}
diff --git a/fastly/block_fastly_service_v1_gzip_test.go b/fastly/block_fastly_service_v1_gzip_test.go
index 1fa90f1b..411dbd55 100644
--- a/fastly/block_fastly_service_v1_gzip_test.go
+++ b/fastly/block_fastly_service_v1_gzip_test.go
@@ -8,7 +8,6 @@ import (
 	gofastly "github.com/fastly/go-fastly/v3/fastly"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
@@ -27,7 +26,7 @@ func TestResourceFastlyFlattenGzips(t *testing.T) {
 			local: []map[string]interface{}{
 				{
 					"name":       "somegzip",
-					"extensions": schema.NewSet(schema.HashString, []interface{}{"css"}),
+					"extensions": []interface{}{"css"},
 				},
 			},
 		},
@@ -47,13 +46,13 @@ func TestResourceFastlyFlattenGzips(t *testing.T) {
 			local: []map[string]interface{}{
 				{
 					"name":          "somegzip",
-					"extensions":    schema.NewSet(schema.HashString, []interface{}{"css", "json", "js"}),
-					"content_types": schema.NewSet(schema.HashString, []interface{}{"text/html"}),
+					"extensions":    []interface{}{"css", "json", "js"},
+					"content_types": []interface{}{"text/html"},
 				},
 				{
 					"name":          "someothergzip",
-					"extensions":    schema.NewSet(schema.HashString, []interface{}{"css", "js"}),
-					"content_types": schema.NewSet(schema.HashString, []interface{}{"text/html", "text/xml"}),
+					"extensions":    []interface{}{"css", "js"},
+					"content_types": []interface{}{"text/html", "text/xml"},
 				},
 			},
 		},
@@ -61,43 +60,9 @@ func TestResourceFastlyFlattenGzips(t *testing.T) {
 
 	for _, c := range cases {
 		out := flattenGzips(c.remote)
-		// loop, because deepequal wont work with our sets
-		expectedCount := len(c.local)
-		var found int
-		for _, o := range out {
-			for _, l := range c.local {
-				if o["name"].(string) == l["name"].(string) {
-					found++
-					if o["extensions"] == nil && l["extensions"] != nil {
-						t.Fatalf("output extensions are nil, local are not")
-					}
-
-					if o["extensions"] != nil {
-						oex := o["extensions"].(*schema.Set)
-						lex := l["extensions"].(*schema.Set)
-						if !oex.Equal(lex) {
-							t.Fatalf("Extensions don't match, expected: %#v, got: %#v", lex, oex)
-						}
-					}
-
-					if o["content_types"] == nil && l["content_types"] != nil {
-						t.Fatalf("output content types are nil, local are not")
-					}
-
-					if o["content_types"] != nil {
-						oct := o["content_types"].(*schema.Set)
-						lct := l["content_types"].(*schema.Set)
-						if !oct.Equal(lct) {
-							t.Fatalf("ContentTypes don't match, expected: %#v, got: %#v", lct, oct)
-						}
-					}
-
-				}
-			}
-		}
 
-		if found != expectedCount {
-			t.Fatalf("Found and expected mismatch: %d / %d", found, expectedCount)
+		if !reflect.DeepEqual(out, c.local) {
+			t.Fatalf("Error matching:\nexpected: %#v\ngot: %#v", c.local, out)
 		}
 	}
 }
@@ -110,20 +75,20 @@ func TestAccFastlyServiceV1_gzips_basic(t *testing.T) {
 	log1 := gofastly.Gzip{
 		ServiceVersion: 1,
 		Name:           "gzip file types",
-		Extensions:     "js css",
+		Extensions:     "css js",
 		CacheCondition: "testing_condition",
 	}
 
 	log2 := gofastly.Gzip{
 		ServiceVersion: 1,
 		Name:           "gzip extensions",
-		ContentTypes:   "text/css text/html",
+		ContentTypes:   "text/html text/css",
 	}
 
 	log3 := gofastly.Gzip{
 		ServiceVersion: 1,
 		Name:           "all",
-		Extensions:     "js html css",
+		Extensions:     "css js html",
 		ContentTypes:   "text/javascript application/x-javascript application/javascript text/css text/html",
 	}
 
@@ -131,7 +96,7 @@ func TestAccFastlyServiceV1_gzips_basic(t *testing.T) {
 		ServiceVersion: 1,
 		Name:           "all",
 		Extensions:     "css",
-		ContentTypes:   "text/javascript application/x-javascript",
+		ContentTypes:   "application/x-javascript text/javascript",
 	}
 
 	resource.ParallelTest(t, resource.TestCase{
@@ -279,12 +244,11 @@ resource "fastly_service_v1" "foo" {
     extensions = ["css", "js", "html"]
 
     content_types = [
-      "text/html",
-      "text/css",
+      "text/javascript",
       "application/x-javascript",
-      "text/css",
       "application/javascript",
-      "text/javascript",
+      "text/css",
+      "text/html",
     ]
   }
 ```

Let me know if this makes sense - and also if you need help getting the acceptance tests running, happy to help if I can!

**Comment by smaeda-ks** [comment]
@bengesoff Thank you so much. Applied your suggestions.

> The flattenGzips function needs to be updated because it still uses schema.NewSet, which causes the Read function to panic - and the unit test for this function can also be simplified given that it doesn't need to account for a TypeSet anymore

Ugh, sorry, I wasn't correctly runnings tests. I should have noticed that.

> and I think some of the reshuffling with the slices before setDiff.Filter isn't actually needed

Right, I was probably doing something weird, which gave me "Error: Provider produced inconsistent result after apply" and made me writing those unnecessary lines of code 🤦

File: fastly/block_fastly_service_v1_gzip.go [hunk 1]
```diff
@@ -118,33 +106,19 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 			Name:           resource["name"].(string),
 		}
 
+		// NOTE: []interface{} is not comparable in Filter function
+		// covert it into string in advance
```
bengesoff: Just noticed this typo here as well
```suggestion
		// convert it into string in advance
```

smaeda-ks: 🤦

bengesoff:
@smaeda-ks no problem!

> > and I think some of the reshuffling with the slices before setDiff.Filter isn't actually needed
> 
> Right, I was probably doing something weird, which gave me "Error: Provider produced inconsistent result after apply" and made me writing those unnecessary lines of code 🤦

I wouldn't be surprised if the issue with `flattenGzips` was causing this error actually

smaeda-ks:
Yeah, that's what I thought too after reading your comment. Thanks for pointing that out.

fastly/block_fastly_service_v1_gzip.go
```diff
@@ -118,33 +106,19 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 			Name:           resource["name"].(string),
 		}
 
+		// NOTE: []interface{} is not comparable in Filter function
+		// convert it into string in advance
+		resource["content_types"] = sliceToString(resource["content_types"].([]interface{}))
+		resource["extensions"] = sliceToString(resource["extensions"].([]interface{}))
+
 		// only attempt to update attributes that have changed
 		modified := setDiff.Filter(resource, oldSet)
```
bengesoff, line 115: Just thought of a possible issue here as well - you're updating `resource[...]` to be a `string`, but `oldSet` will still have it as `[]interface{}` so will probably always be included in `modified`. It's a bit confusing but I don't think it causes any errors because it will just include the existing value in the update every time. 

There's probably two options to rectify that: either just set `resource["content_types"]` etc to `nil` with a comment to explain that it doesn't matter if you redundantly update these values; or you could update `setDiff.Filter` to be more intelligent than just a simple `!=` equality check (i.e. checking the types before comparing, and adding a better check for more complex values like slices)

smaeda-ks: I was thinking of that pattern as well at first. But, I wasn't sure if that's a good idea to check the types for all inputs unconditionally. That seems a bit inefficient processing.

Like, we could make the `Filter` function take optional parameters (e.g., AFOP) and do some custom conversions/diff based on the block type, for instance. This way, `Filter` function is still backward-compatible.

bengesoff: How about something like this? Avoids a panic by using `DeepEqual` which checks the types for you, and makes it clear that we're ignoring the resulting `modified` variable for `content_types` and `extensions` by always updating them. I haven't run the whole acceptance test suite to double check but seems like it would work?

```diff
diff --git a/fastly/block_fastly_service_v1_gzip.go b/fastly/block_fastly_service_v1_gzip.go
index 756a3cec..25813411 100644
--- a/fastly/block_fastly_service_v1_gzip.go
+++ b/fastly/block_fastly_service_v1_gzip.go
@@ -104,22 +104,13 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 			ServiceID:      d.Id(),
 			ServiceVersion: latestVersion,
 			Name:           resource["name"].(string),
+			ContentTypes:   gofastly.String(sliceToString(resource["content_types"].([]interface{}))),
+			Extensions:     gofastly.String(sliceToString(resource["extensions"].([]interface{}))),
 		}
 
-		// NOTE: []interface{} is not comparable in Filter function
-		// convert it into string in advance
-		resource["content_types"] = sliceToString(resource["content_types"].([]interface{}))
-		resource["extensions"] = sliceToString(resource["extensions"].([]interface{}))
-
 		// only attempt to update attributes that have changed
 		modified := setDiff.Filter(resource, oldSet)
 
-		if v, ok := modified["content_types"]; ok {
-			opts.ContentTypes = gofastly.String(v.(string))
-		}
-		if v, ok := modified["extensions"]; ok {
-			opts.Extensions = gofastly.String(v.(string))
-		}
 		if v, ok := modified["cache_condition"]; ok {
 			opts.CacheCondition = gofastly.String(v.(string))
 		}
diff --git a/fastly/diff.go b/fastly/diff.go
index c337c065..8151af66 100644
--- a/fastly/diff.go
+++ b/fastly/diff.go
@@ -2,6 +2,7 @@ package fastly
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -132,7 +133,7 @@ func (h *SetDiff) Filter(modified map[string]interface{}, oldSet *schema.Set) ma
 
 		if m["name"].(string) == modified["name"].(string) {
 			for k, v := range m {
-				if v != modified[k] {
+				if !reflect.DeepEqual(v, modified[k]) {
 					filtered[k] = modified[k]
 				}
 			}
```

Integralist:
@smaeda-ks just checking in to see if there's anymore feedback you need from @bengesoff 

Thanks!

StarpTech:
Hi, what's the status here?

smaeda-ks:
301 to https://github.com/fastly/terraform-provider-fastly/pull/510

Issue #75, opened 2018-04-25
Can't override default_host = "" of Terraform


junichi-hata:
Hello,

### Terraform Version
Terraform v0.11.7
+ provider.fastly v0.1.4

### Affected Resource(s)
fastly_service_v1

### Terraform Configuration Files
```hcl
resource "fastly_service_v1" "myservice" {
  name = "xxxxxxxxxxx.global.ssl.fastly.net"

  domain {
    name = "xxxxxxxxxxx.global.ssl.fastly.net"
  }

  backend {
    address = "10.0.0.1"
    name    = "Temp"
    port    = 80
  }

  default_ttl   = 10
  force_destroy = true
  default_host = ""

  vcl {
    name    = "main"
    main    = "true"
    content = "${file("main.vcl")}"
  }
}
```

### Expected Behavior
`Override Host` header of Fastly should be deleted

### Actual Behavior
`No changes. Infrastructure is up-to-date.` by `terraform apply`

### Steps to Reproduce
0.  set `default_host = ""` in `override.tf`
1. `terraform apply`

### Important Factoids
When I set `default_host = false`, it will set `0` for `Override Host` of Fastly UI, but it deletes in actual vcl line of `set req.http.host = xxxx`

Integralist:
Thank you for taking the time to open this issue.

I appreciate this was reported a long time ago (and I appreciate your patience awaiting our feedback) but I believe we may now have a fix in place to resolve the issue you're describing. We merged it this morning as the new minor version `v0.21.3`, here are some details about the changes:

- [Migrate Terraform provider to use latest Go-Fastly 2.0.0 release](https://github.com/fastly/terraform-provider-fastly/releases/tag/v0.21.3).
- [Key changes in Go-Fastly 2.0.0](https://github.com/fastly/terraform-provider-fastly/issues/85) that I think relate to this issue.

If possible, could you try upgrading to the latest version (`v0.21.3`) and let me know if this indeed resolves the issue.

Thanks again.

smaeda-ks:
This has been fixed in v0.33.0:
https://github.com/fastly/terraform-provider-fastly/blob/main/CHANGELOG.md

Sorry for taking such a long time! We appreciate your feedback.

PR #434, opened 2021-06-26
fix default_host not being removed

smaeda-ks:
fix for: https://github.com/fastly/terraform-provider-fastly/issues/85 and https://github.com/fastly/terraform-provider-fastly/issues/75

To clarify, this is to allow go-fastly client to send `...&general.default_host=&...` in query parameters when it's removed. This lets our API backend actually removes this field value from DB.

In this case, `opts.DefaultHost` must exist with an empty value.

### DIFF
--- a/fastly/block_fastly_service_v1_settings.go
+++ b/fastly/block_fastly_service_v1_settings.go
@@ -26,6 +26,8 @@ func (h *SettingsServiceAttributeHandler) Process(d *schema.ResourceData, latest
 
 	if attr, ok := d.GetOk("default_host"); ok {
 		opts.DefaultHost = gofastly.String(attr.(string))
+	} else if d.HasChanges("default_host") {
+		opts.DefaultHost = gofastly.String("")
 	}
 
 	log.Printf("[DEBUG] Update Settings opts: %#v", opts)
@@ -70,7 +72,6 @@ func (h *SettingsServiceAttributeHandler) Register(s *schema.Resource) error {
 	s.Schema["default_host"] = &schema.Schema{
 		Type:        schema.TypeString,
 		Optional:    true,
-		Computed:    true,
 		Description: "The default hostname",
 	}
 	return nil
--- a/fastly/block_fastly_service_v1_settings.go
+++ b/fastly/block_fastly_service_v1_settings.go
@@ -19,15 +19,10 @@ func (h *SettingsServiceAttributeHandler) Process(d *schema.ResourceData, latest
 	opts := gofastly.UpdateSettingsInput{
 		ServiceID:      d.Id(),
 		ServiceVersion: latestVersion,
+		DefaultHost:    gofastly.String(d.Get("default_host").(string)),
 		// default_ttl has the same default value of 3600 that is provided by
 		// the Fastly API, so it's safe to include here
 		DefaultTTL: uint(d.Get("default_ttl").(int)),
-	}
-
-	if attr, ok := d.GetOk("default_host"); ok {
-		opts.DefaultHost = gofastly.String(attr.(string))
-	} else if d.HasChanges("default_host") {
-		opts.DefaultHost = gofastly.String("")
 	}
 
 	log.Printf("[DEBUG] Update Settings opts: %#v", opts)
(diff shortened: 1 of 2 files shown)


fastly/block_fastly_service_v1_settings.go
```diff
@@ -26,6 +26,8 @@ func (h *SettingsServiceAttributeHandler) Process(d *schema.ResourceData, latest
 
 	if attr, ok := d.GetOk("default_host"); ok {
```
bengesoff: Would this still work if you removed the `if` statement and just set `DefaultHost` to `gofastly.String(d.Get("default_host").(string)` on the initial `opts` declaration above? Given that it defaults to `""` it might be clearer if it still works?

bengesoff (commented):
Might be worth writing an acceptance test to catch this, what do you think?

fastly/block_fastly_service_v1_settings.go
smaeda-ks: Yeah, that works and clearer. Thanks for the suggestion. Applied it.

fastly/block_fastly_service_v1_settings.go
```diff
@@ -19,15 +19,12 @@ func (h *SettingsServiceAttributeHandler) Process(d *schema.ResourceData, latest
 	opts := gofastly.UpdateSettingsInput{
 		ServiceID:      d.Id(),
 		ServiceVersion: latestVersion,
+		DefaultHost:    gofastly.String(d.Get("default_host").(string)),
```
Integralist, line 22: @smaeda-ks @bengesoff 

I might just be confused about the problem but is the change required here actually to update go-fastly so that the `DefaultHost` is no longer a pointer reference (and to not `omitempty`), then we update the terraform provider to use the new go-fastly release?

https://github.com/fastly/go-fastly/blob/6bf96a5a4c2ba87230e30ddd8683626aa0828bc4/fastly/settings.go#L57

That way the terraform logic can stay the same and it means that if it doesn't execute the setting of `opts.DefaultHost`, then go-fastly would use a zero value of `""` for that field.

Otherwise we're trying to side-step the go-fastly implementation when that implementation appears to be incorrect.

bengesoff: I think I see what you're saying, but might it be relevant to update other Settings and leave `DefaultHost` set how it is? And for that you'd need a pointer and `omitempty` so you can have a `nil` pointer and omit it from the update?

smaeda-ks: I think go-fastly is fine. `general.default_host` is an optional parameter so when there's no change the client doesn't need to send this parameter at all, and `omitempty` should be there.

But in order to "delete" this setting, the client must send `...&general.default_host=&...` as a query parameter. Right now, when `default_host` is set to `""` or deleted from TF file, we're just not sending this parameter at all, and has no effect on that value (via PUT request).

Integralist: Ah OK I see 👍🏻

Integralist (approved):

Integralist (commented):
I concur with @bengesoff and think it would be good idea to have a test to validate these changes (or update an existing test).

smaeda-ks:
OK. I can try writing tests (for my learning).

smaeda-ks:
@Integralist @bengesoff 
OK, I've added a test. I also ran the test and verified it passed.

```
% FASTLY_API_KEY=xxxx TESTARGS='-run=TestAccFastlyServiceV1_defaultHost' make testacc

--- PASS: TestAccFastlyServiceV1_defaultHost (98.39s)
PASS
ok  	github.com/fastly/terraform-provider-fastly/fastly	98.757s
```

and with the old code, it fails as expected:

```
% FASTLY_API_KEY=xxxx TESTARGS='-run=TestAccFastlyServiceV1_defaultHost' make testacc

=== RUN   TestAccFastlyServiceV1_defaultHost
=== PAUSE TestAccFastlyServiceV1_defaultHost
=== CONT  TestAccFastlyServiceV1_defaultHost
    resource_fastly_service_v1_test.go:628: Step 2/2 error: Check failed: Check 2/3 error: fastly_service_v1.foo: Attribute 'default_host' expected "", got "a3w.aws.amazon.com"
--- FAIL: TestAccFastlyServiceV1_defaultHost (74.77s)
```

and did some clean up as well:
https://github.com/fastly/terraform-provider-fastly/pull/434/commits/fa33a24cb5d71e29a6ecccd33818cf25e3789c1c

Integralist:
This looks good to me. Thanks @smaeda-ks !

Issue #430, opened 2021-06-23
fastly_tls_subscription return one challenge


hosht:
### Terraform Version
Terraform v1.0.0
on darwin_amd64
+ provider registry.terraform.io/fastly/fastly v0.32.0

### Affected Resource(s)
- fastly_tls_subscription

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
terraform {
  required_version = "1.0.0"

  required_providers {
    fastly = {
      source  = "fastly/fastly"
      version = "0.32.0"
    }
  }
}

locals {
  tls_domains = [
    "a.example.com",
    "b.example.com"
  ]
}

resource "fastly_tls_subscription" "this" {
  domains               = local.tls_domains
  certificate_authority = "lets-encrypt"
  configuration_id      = "configuration_id"
}
```

### Expected Behavior
when domain argument receives more than two domains, fastly_tls_subscription should return list of managed_dns_challenges and managed_http_challenges for domain validation.

```
"managed_dns_challenges": [
  {
    "record_name": "_acme-challenge.a.example.com",
    "record_type": "CNAME",
    "record_value": "abcde12345.fastly-validations.com"
  },
  {
    "record_name": "_acme-challenge.b.example.com",
    "record_type": "CNAME",
    "record_value": "abcde12345.fastly-validations.com"
  }
]
```

### Actual Behavior
it returns just a.example.com challenges. is it right?

```
"managed_dns_challenge": {
  "record_name": "_acme-challenge.a.example.com",
  "record_type": "CNAME",
  "record_value": "abcde12345.fastly-validations.com"
},
"managed_http_challenges": [
  {
    "record_name": "a.example.com",
    "record_type": "A",
...
```

smaeda-ks:
Thanks, @hosht! This appears to be a bug in our provider.

As this is going to be a breaking change, we will be discussing the fix in https://github.com/fastly/terraform-provider-fastly/pull/435

smaeda-ks:
This has been fixed in v0.33.0:
https://github.com/fastly/terraform-provider-fastly/blob/main/CHANGELOG.md

You can use the new `managed_dns_challenges` attribute instead:
https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/tls_subscription

We appreciate your feedback.

hosht:
Thank you so much, it saves my time !

PR #435, opened 2021-06-27
[bug] fix fastly_tls_subscription resource to properly handle challenges for multi-SAN

smaeda-ks:
fix for: https://github.com/fastly/terraform-provider-fastly/issues/430

Looks like when this resource was originally implemented, multi-SAN cert subscription wasn't taken into account. Specifically, this part...
https://github.com/fastly/terraform-provider-fastly/blob/v0.32.0/fastly/resource_fastly_tls_subscription.go#L176

Therefore, the state file only includes the first domain's challenge data, which is a bug.

Though this is definitely going to be a breaking change.

### DIFF
--- a/docs/resources/tls_subscription.md
+++ b/docs/resources/tls_subscription.md
@@ -78,11 +78,12 @@ data "aws_route53_zone" "demo" {
 
 # Set up DNS record for managed DNS domain validation method
 resource "aws_route53_record" "domain_validation" {
-  name            = fastly_tls_subscription.example.managed_dns_challenges.record_name
-  type            = fastly_tls_subscription.example.managed_dns_challenges.record_type
+  for_each = { for domain in fastly_tls_subscription.example.managed_dns_challenges : domain.record_name => domain }
+  name            = each.value.record_name
+  type            = each.value.record_type
   zone_id         = data.aws_route53_zone.demo.id
   allow_overwrite = true
-  records         = [fastly_tls_subscription.example.managed_dns_challenges.record_value]
+  records         = [each.value.record_value]
   ttl             = 60
 }
 
--- a/docs/resources/tls_subscription.md
+++ b/docs/resources/tls_subscription.md
@@ -161,6 +161,7 @@ $ terraform import fastly_tls_subscription.demo xxxxxxxxxxx
 ### Read-Only
 
 - **created_at** (String) Timestamp (GMT) when the subscription was created.
+- **managed_dns_challenge** (Map of String) (DEPRECATED) The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership.
 - **managed_dns_challenges** (Set of Object) A list of options for configuring DNS to respond to ACME DNS challenge in order to verify domain ownership. (see [below for nested schema](#nestedatt--managed_dns_challenges))
 - **managed_http_challenges** (Set of Object) A list of options for configuring DNS to respond to ACME HTTP challenge in order to verify domain ownership. Best accessed through a `for` expression to filter the relevant record. (see [below for nested schema](#nestedatt--managed_http_challenges))
 - **state** (String) The current state of the subscription. The list of possible states are: `pending`, `processing`, `issued`, and `renewing`.
--- a/docs/resources/tls_subscription.md
+++ b/docs/resources/tls_subscription.md
@@ -12,7 +12,7 @@ Enables TLS on a domain using a certificate managed by Fastly.
 
 DNS records need to be modified on the domain being secured, in order to respond to the ACME domain ownership challenge.
 
-There are two options for doing this: the `managed_dns_challenge`, which is the default method; and the `managed_http_challenges`, which points production traffic to Fastly.
+There are two options for doing this: the `managed_dns_challenges`, which is the default method; and the `managed_http_challenges`, which points production traffic to Fastly.
 
 ~> See the [Fastly documentation](https://docs.fastly.com/en/guides/serving-https-traffic-using-fastly-managed-certificates#verifying-domain-ownership) for more information on verifying domain ownership.
 
@@ -78,11 +78,11 @@ data "aws_route53_zone" "demo" {
 
 # Set up DNS record for managed DNS domain validation method
 resource "aws_route53_record" "domain_validation" {
-  name            = fastly_tls_subscription.example.managed_dns_challenge.record_name
-  type            = fastly_tls_subscription.example.managed_dns_challenge.record_type
+  name            = fastly_tls_subscription.example.managed_dns_challenges.record_name
+  type            = fastly_tls_subscription.example.managed_dns_challenges.record_type
   zone_id         = data.aws_route53_zone.demo.id
   allow_overwrite = true
-  records         = [fastly_tls_subscription.example.managed_dns_challenge.record_value]
+  records         = [fastly_tls_subscription.example.managed_dns_challenges.record_value]
   ttl             = 60
 }
 
@@ -112,14 +112,14 @@ In addition to the arguments listed above, the following attributes are exported
 * `created_at` - Timestamp (GMT) when the subscription was created.
 * `updated_at` - Timestamp (GMT) when the subscription was last updated.
 * `state` - The current state of the subscription. The list of possible states are: `pending`, `processing`, `issued`, and `renewing`.
-* `managed_dns_challenge` - The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership. See Managed DNS Challenge below for details.
+* `managed_dns_challenges` - A list of options for configuring DNS to respond to ACME DNS challenge in order to verify domain ownership. See Managed DNS Challenge below for details.
 * `managed_http_challenges` - A list of options for configuring DNS to respond to ACME HTTP challenge in order to verify domain ownership. See Managed HTTP Challenges below for details.
 
 ### Managed DNS Challenge
 
-The available attributes in the `managed_dns_challenge` block are:
+The available attributes in the `managed_dns_challenges` block are:
 
-* `record_name` - The name of the DNS record to add. For example `_acme-challenge.example.com`. Accessed like this, `fastly_tls_subscription.tls.managed_dns_challenge.record_name`.
+* `record_name` - The name of the DNS record to add. For example `_acme-challenge.example.com`. Accessed like this, `fastly_tls_subscription.tls.managed_dns_challenges.record_name`.
 * `record_type` - The type of DNS record to add, e.g. `A`, or `CNAME`.
 * `record_value` - The value to which the DNS record should point, e.g. `xxxxx.fastly-validations.com`.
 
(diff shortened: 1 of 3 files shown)


templates/resources/tls_subscription.md.tmpl
```diff
@@ -78,11 +78,11 @@ data "aws_route53_zone" "demo" {
...
 # Set up DNS record for managed DNS domain validation method
 resource "aws_route53_record" "domain_validation" {
-  name            = fastly_tls_subscription.example.managed_dns_challenge.record_name
-  type            = fastly_tls_subscription.example.managed_dns_challenge.record_type
+  name            = fastly_tls_subscription.example.managed_dns_challenges.record_name
+  type            = fastly_tls_subscription.example.managed_dns_challenges.record_type
   zone_id         = data.aws_route53_zone.demo.id
   allow_overwrite = true
-  records         = [fastly_tls_subscription.example.managed_dns_challenge.record_value]
+  records         = [fastly_tls_subscription.example.managed_dns_challenges.record_value]
```
smaeda-ks: No longer works as it's now TypeSet rather than TypeMap

Probably better to change them to use `for_each` ...?

```tf
resource "aws_route53_record" "domain_validation" {
  for_each        = toset(fastly_tls_subscription.example.managed_dns_challenge)
  name            = each.record_name
  type            = each.record_type
  zone_id         = data.aws_route53_zone.demo.id
  allow_overwrite = true
  records         = [each.record_value]
  ttl             = 60
}
```

bengesoff: I'd agree that `for_each` would make more sense here - especially as I expect the typical workflow would be to add a record for each DNS challenge in the set. However I think that snippet needs a couple of minor snippets for it to work! 🙂

fastly/resource_fastly_tls_subscription.go
```diff
@@ -68,11 +68,29 @@ func resourceFastlyTLSSubscription() *schema.Resource {
 				Description: "The current state of the subscription. The list of possible states are: `pending`, `processing`, `issued`, and `renewing`.",
 				Computed:    true,
 			},
-			"managed_dns_challenge": {
-				Type:        schema.TypeMap,
-				Description: "The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership.",
+			"managed_dns_challenges": {
```
bengesoff, line 78: As you mentioned, this schema change is a breaking change - to assist in the transition to the new version, it might be good to implement a [`StateUpgrader` function](https://www.terraform.io/docs/extend/resources/state-migration.html) here. The users will still have to update their config files to use the new schema, but at least the statefile can be upgraded automatically

templates/resources/tls_subscription.md.tmpl
```diff
@@ -112,14 +112,14 @@ In addition to the arguments listed above, the following attributes are exported
 * `created_at` - Timestamp (GMT) when the subscription was created.
 * `updated_at` - Timestamp (GMT) when the subscription was last updated.
 * `state` - The current state of the subscription. The list of possible states are: `pending`, `processing`, `issued`, and `renewing`.
-* `managed_dns_challenge` - The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership. See Managed DNS Challenge below for details.
```
bengesoff, line 115: Possibly only tangentially relevant to this PR, but it could be worth removing the "argument reference" and "attribute reference" sections from the docs template, as it gets added to the rendered docs page by `tfplugindocs` already

templates/resources/tls_subscription.md.tmpl
smaeda-ks: @bengesoff Thanks.

```tf
resource "aws_route53_record" "domain_validation" {
  for_each = { for domain in fastly_tls_subscription.example.managed_dns_challenges : domain.record_name => domain }
  name            = each.value.record_name
  type            = each.value.record_type
  zone_id         = data.aws_route53_zone.demo.id
  allow_overwrite = true
  records         = [each.value.record_value]
  ttl             = 60
}
```

maybe something like this would work?

bengesoff: I'd imagine you could use something like the following although I haven't tried it
```hcl
resource "aws_route53_record" "domain_validation" {
  for_each        = fastly_tls_subscription.example.managed_dns_challenges
  name            = each.record_name
  type            = each.record_type
  zone_id         = data.aws_route53_zone.demo.id
  allow_overwrite = true
  records         = [each.record_value]
  ttl             = 60
}
```

smaeda-ks: ok, I'll test

fastly/resource_fastly_tls_subscription.go
smaeda-ks: But `$ terraform refresh` also updates the state file, right? Or, any subsequent apply after the version update would override and correct the state file, is what I was seeing in my tests earlier.

bengesoff: Fair enough if you've tested it! I didn't know if it would correctly overwrite the old state with the new schema, or if you'd have to migrate it. I can see that would be possible in principle as it's a fully computed value, so we wouldn't need to try and preserve their config file, so happy to leave it if it works 👍🏼

templates/resources/tls_subscription.md.tmpl
smaeda-ks: Ok, just tested this. My second suggestion worked fine. Your suggestion gave me the below error:

```tf
| Error: Invalid "each" attribute
│
│   on fastly.tf line 105, in resource "aws_route53_record" "demo":
│  105:   name            = each.record_name
│
│ The "each" object does not have an attribute named "record_name". The supported attributes are each.key and each.value, the current key and value pair of the "for_each" attribute set.
```

so I'll go with the second one. Thanks

bengesoff: Oh you're right, sorry 🤦🏻‍♂️ It was supposed to be `each.value.record_name` but that also doesn't work as `for_each` must be either a map or a set of strings! 

I've just tested it as well and there's a minor error with both:
```
╷
│ Error: Invalid for_each argument
│ 
│   on main.tf line 41, in resource "aws_route53_record" "domain_validation":
│   41:   for_each = { for domain in fastly_tls_subscription.example.managed_dns_challenges : domain.record_name => domain }
│     ├────────────────
│     │ fastly_tls_subscription.example.managed_dns_challenges is a set of object, known only after apply
│ 
│ The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the resources that the for_each depends
│ on.
╵
```

Unfortunately it has to be applied in two stages to get around this - do you think that's worth mentioning somewhere?

templates/resources/tls_subscription.md.tmpl
smaeda-ks: Actually, I think we still have this problem: https://github.com/hashicorp/terraform-plugin-docs/issues/28
so I think we should leave it as-is for the time being.

bengesoff: Ah yes good point

Integralist (approved):
This looks OK to me, but we don't want to merge this until we've discussed a major 1.0.0 release (@smaeda-ks I'll be having discussions with @bengesoff about 1.0.0 as part of our Q3 goals). 

In the mean time I'd also want to get @bengesoff's approval of these changes too.

bengesoff (approved):
Think `make generate-docs` needs rerunning, but otherwise LGTM 👍🏼

smaeda-ks:
pushed doc update

virgofx:
Any chance to get this merged and a release cut? Our workaround is to create separate subscriptions per domain :(

Integralist:
Hi @virgofx the current implementation is considered a breaking change and so would need to be part of a 1.0.0 release (something we're actively working towards at the moment, but unfortunately I couldn't say with certainty as to _when_ exactly it would be ready to release).

In the meantime I'm discussing with the team potential workarounds that don't result in a breaking change.

docs/resources/tls_subscription.md
```diff
@@ -160,11 +161,22 @@ $ terraform import fastly_tls_subscription.demo xxxxxxxxxxx
 ### Read-Only
 
 - **created_at** (String) Timestamp (GMT) when the subscription was created.
-- **managed_dns_challenge** (Map of String) The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership.
+- **managed_dns_challenge** (Map of String) (DEPRECATED) The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership.
```
smaeda-ks: Marked this attribute as DEPRECATED

fastly/resource_fastly_tls_subscription.go
```diff
@@ -172,24 +196,53 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
...
+				})
 			}
+		}
+	}
 
-			managedDNSChallenge = map[string]string{
-				"record_type":  challenge.RecordType,
-				"record_name":  challenge.RecordName,
-				"record_value": challenge.Values[0],
+	// NOTE: This block of code contains a bug where the state file will only include
```
Integralist: I'd prefer to have this as a 'TODO' as I regularly do searches for `TODO:` across our various codebases looking for things to fix up, and although this would be something we actively look for as part of the future 1.0.0 release, I think it's still worth marking as a TODO so we don't accidentally miss it.

```suggestion
	// TODO: This block of code contains a bug where the state file will only include
```

fastly/resource_fastly_tls_subscription.go
```diff
@@ -70,10 +70,34 @@ func resourceFastlyTLSSubscription() *schema.Resource {
 			},
 			"managed_dns_challenge": {
 				Type:        schema.TypeMap,
-				Description: "The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership.",
+				Description: "(DEPRECATED) The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership.",
```
bengesoff: I believe there's `Deprecated` field on the `schema.Schema` type too, which might be worth using. I'm not sure whether that gets picked up by `terraform-plugin-docs` however, so you might still have to keep the description as you have it now

smaeda-ks: Ah, that's good to know. Let me test that.

smaeda-ks: https://www.terraform.io/docs/extend/best-practices/deprecations.html#renaming-a-required-attribute

> Add Deprecated to the schema definition of the existing (now the "old") attribute, noting to use the new attribute in the message.

I thought that would be helpful, but seems the `Deprecated` field doesn't really do anything for `Computed` Schema. I've tested but nothing showed up in my console, which is understandable. So I think we're okay with the current for now. cc: @Integralist

smaeda-ks: Ah, sorry, I think I misunderstood it. It's just used for documentation, I see. I've just pushed a commit to set that field. Thanks

fastly/resource_fastly_tls_subscription.go
```diff
@@ -73,6 +73,31 @@ func resourceFastlyTLSSubscription() *schema.Resource {
 				Description: "The details required to configure DNS to respond to ACME DNS challenge in order to verify domain ownership.",
 				Computed:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
+				Deprecated:  "Use 'managed_http_challenges' attribute instead",
```
bengesoff: I think the message should be "Use 'managed_**dns**_challenges' attribute instead" instead shouldn't it?

smaeda-ks: fixed

smaeda-ks:
Please merge this. Thanks

bengesoff (approved):

virgofx:
Thanks for merging! Can you please cut the 1.0.0 release so consumers can make use of this great work 😃

Integralist:
Hi @virgofx this change will actually go into a `0.33.0` release. We're actively fixing some broken integration tests at the moment, but once we have those fixed I'll be cutting that new release 🙂

PR #439, opened 2021-07-05
replace StateFunc with ValidateDiagFunc

smaeda-ks:
301 from https://github.com/fastly/terraform-provider-fastly/pull/438

Screen Shot 2021-07-05 at 20 23 20

### DIFF
--- a/fastly/validators.go
+++ b/fastly/validators.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 
 	gofastly "github.com/fastly/go-fastly/v3/fastly"
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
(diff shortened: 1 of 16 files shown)


smaeda-ks:
@Integralist @bengesoff I've created a new PR as per the prior discussions in #438 
Thanks

fastly/validators.go
```diff
@@ -179,3 +180,14 @@ func validatePEMBlocks(pemType string) schema.SchemaValidateDiagFunc {
 		return nil, nil
 	})
 }
+
+func validateStringTrimmed() schema.SchemaValidateDiagFunc {
```
bengesoff: When I last had a crack at implementing this, I used the following function:
```go
// validateStringTrimmed is a SchemaValidateDiagFunc which checks whether the string is trimmed.
func validateStringTrimmed(value interface{}, p cty.Path) diag.Diagnostics {
    if value != strings.TrimSpace(value.(string)) {
        return diag.FromErr(p.NewErrorf("string must not contain trailing whitespace"))
    }

    return nil
}
```

It works either way, but for a brand new validation function, it could make sense to use the `schema.SchemaValidateDiagFunc` type directly rather than using the legacy `ToDiagFunc` conversion function like the others do. Other than that, your error message is much better than the one I had!

smaeda-ks: Thanks @bengesoff
I've modified the function a bit.

bengesoff:
@smaeda-ks thanks for the change, I like the updated `validateStringTrimmed` function 👍🏼 looks good overall although you'll also need to look at the tests as some of those will need `trimspace()` adding to them

Integralist:
Thanks @smaeda-ks ! 

This LGTM as an intermediary step towards preventing users losing resources. 

Let's get the tests checked/updated and then @bengesoff and I can give a second review.

smaeda-ks:
@bengesoff @Integralist  Just pushed https://github.com/fastly/terraform-provider-fastly/pull/439/commits/eead702e686efffb7f3f88c82f7cd6073dc08827

Since our test fixtures (ie, `fastly/test_fixtures/fastly_test_*`) don't include trailing spaces, most tests are fine anyhow, I think. The only place I had to fix is forgotten `strings.TrimSpace()` in the `generateKey()` function.

bengesoff (approved):
Nice one, looks good to me 👍🏼  it's really nice to be able to get the "hack" out of there as well

Integralist (approved):

PR #441, opened 2021-07-10
[bug] suppress gzip diff unless fields are explicitly set

smaeda-ks:
We recently made some changes to our [Gzip API](https://developer.fastly.com/reference/api/vcl-services/gzip/) endpoint.

Previously, the endpoint accepted an empty value for `content_types` and `extensions` parameters (ie., `""` is stored in DB as the value). But after the change, API will set their default values if an empty value is sent for these parameters from clients.

This results in unexpected (and potentially infinite) diff as the TF state file contains no data for these parameters but the API returns different data upon refresh.

Since Terraform Plugin SDK currently does not support a way to configure a `Default` for `TypeList/TypeSet`, the only workaround seems to be ignoring these fields in the API response IF those values are not set explicitly in the TF file by the user. Or, we can also make these fields `Required` (+ disallow empty value) even though they're optional in the API spec.

This PR also introduces a downside where remote changes for those particular fields will be ignored unless users set them in TF file.

For example, with this gzip block setting:

```tf
  gzip {
    name = "Test gzip"
  }
```

and if I change `content_types` or `extensions` value on the UI, the subsequent TF runs won't report any diff. But I think that's acceptable.

### DIFF
--- a/fastly/block_fastly_service_v1_gzip.go
+++ b/fastly/block_fastly_service_v1_gzip.go
@@ -127,7 +127,12 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 		// this and so we've updated the below code to convert the type asserted
 		// int into a uint before passing the value to gofastly.Uint().
 		if v, ok := modified["content_types"]; ok {
+			// NOTE: this particular line was added to address a change in the backend API
+			// where it used to accept an empty value but now will use a default value if no value provided.
+			// To allow "resetting" the value on modify (user removed the attribute or set empty value)
+			// we always default to sending an empty string
 			opts.ContentTypes = gofastly.String("")
+
 			set := v.(*schema.Set)
 			if len(set.List()) > 0 {
 				var s []string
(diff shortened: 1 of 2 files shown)


smaeda-ks:
@Integralist This fixes the failing `TestAccFastlyServiceV1_gzips` tests, but needs some discussion...

do some hack vs breaking change (change them to `Required`)

fastly/block_fastly_service_v1_gzip.go
```diff
@@ -127,6 +127,7 @@ func (h *GzipServiceAttributeHandler) Process(d *schema.ResourceData, latestVers
 		// this and so we've updated the below code to convert the type asserted
 		// int into a uint before passing the value to gofastly.Uint().
 		if v, ok := modified["content_types"]; ok {
+			opts.ContentTypes = gofastly.String("")
```
Integralist, line 134: Might be worth adding a comment here (and same for `opts.Extensions = gofastly.String("")`) to state that this line was added to address a change in the backend API where it used to accept an empty value but now will use a default value if no value provided, so to avoid a diff we always default to sending an empty string.

fastly/block_fastly_service_v1_gzip.go
```diff
@@ -173,6 +175,33 @@ func (h *GzipServiceAttributeHandler) Read(d *schema.ResourceData, s *gofastly.S
...
+	// Fastly API will actually set the default value silently when these fields are not sent
+	// or an empty field value is sent. This will cause unexpected diff.
+	// We need to ignore these fields in the API response unless field values are explicitly set.
+	type IgnoreFields struct {
+		Name string
+	}
+	ignoreList := map[string][]IgnoreFields{}
+
+	for _, elem := range d.Get("gzip").(*schema.Set).List() {
+		name := elem.(map[string]interface{})["name"].(string)
```
Integralist: Might be nice to do the type assertion only once rather than three times:

```go
m := elem.(map[string]interface{})
name := m["name"].(string)
if m["content_types"].(*schema.Set).Len() == 0 {
	ignoreList[name] = append(ignoreList[name], IgnoreFields{Name: "content_types"})
}
if m["extensions"].(*schema.Set).Len() == 0 {
	ignoreList[name] = append(ignoreList[name], IgnoreFields{Name: "extensions"})
}
```

fastly/block_fastly_service_v1_gzip.go
```diff
@@ -173,6 +175,33 @@ func (h *GzipServiceAttributeHandler) Read(d *schema.ResourceData, s *gofastly.S
...
+		}
+	}
+
+	for i, g := range gl {
+		if v, ok := ignoreList[g["name"].(string)]; ok && len(v) > 0 {
+			for _, sl := range v {
+				gl[i][sl.Name] = nil
+			}
+		}
+	}
```
Integralist: This might be a good use case for an extra block scope as a delimiter for the specialised behaviour...

```go
// NOTE: Although "content_types" and "extensions" fields are optional in spec...etc
{
	type IgnoreFields struct {
		Name string
	}
	// ...and the rest of the new code here.
}
```

This is a nice _visual_ indicator of the boundary of the new code, and in future if we want to remove the code it's easier to identify where the boundary is for the specialised logic.

fastly/block_fastly_service_v1_gzip.go
```diff
@@ -173,6 +175,33 @@ func (h *GzipServiceAttributeHandler) Read(d *schema.ResourceData, s *gofastly.S
 
 	gl := flattenGzips(gzipsList)
 
+	// Note: Although "content_types" and "extensions" fields are optional in spec,
```
Integralist: For consistency in go projects I typically uppercase these types of comments like `// NOTE: ...`

Integralist (commented):
Thanks @smaeda-ks !

> remote changes for those particular fields will be ignored unless users set them in TF file.

I think this is acceptable considering users should be using TF to manage these resources.

/cc @bengesoff would be good to get your review and thoughts.

bengesoff:
This looks like a good temporary fix although I think we should move to replace it with the breaking change when we can. Other than that, I agree with @Integralist's comments and think it looks good otherwise 👌🏼

smaeda-ks:
Thank you @Integralist and @bengesoff.
I've applied your feedback.

Integralist (approved):

Issue #86, opened 2018-07-08
Feature Request: Remove need for API key when using fastly_ip_ranges data source.


johnwesley:
### Terraform Version
`v0.11.7`

### Affected Provider
- Fastly 

### Expected Behavior
Providing a Fastly API key should be optional when calling `data "fastly_ip_ranges" "fastly" {}`.

### Actual Behavior
Terraform will error out if a key is not provided when calling just the above.
```
Error: Error refreshing state: 1 error(s) occurred:
* provider.fastly: [Err] No API key for Fastly
```

### Steps to Reproduce
1. Comment out `api_key` in the manifest  
```
provider "fastly" {
#  api_key = "96516f75a33f9f255b780e9b8e9eb99b"
}
data "fastly_ip_ranges" "fastly" {}
```
2. `terraform plan`

### Important Factoids
https://api.fastly.com/public-ip-list is public endpoint. If this is the **only** function used in the Fastly provider set, there is no need for an API key. This was the case for me when using `fastly_ip_ranges` to keep our AWS security groups up to date. I didn't want to keep an API key in SCM, so I created a read-only token, but this is still less than ideal.

The conditional in `config.go` puts a hard stop on this:
```
	if c.ApiKey == "" {
		return nil, fmt.Errorf("[Err] No API key for Fastly")
```

ledakis:
It appears the api_key is checked at `terraform init` time, but not used later on to lookup the data source.
I have used dummy text as an api key just to pass the init check you mentioned above @johnwesley.

timoguin:
The GitHub provider handles this same use case by accepting an `anonymous` argument in the provider block:

https://www.terraform.io/docs/providers/github/#anonymous

smaeda-ks:
Hi all, the new `no_auth` provider option is added in v0.33.0:
https://github.com/fastly/terraform-provider-fastly/blob/main/CHANGELOG.md

```tf
provider "fastly" {
  no_auth = true
}
data "fastly_ip_ranges" "test" {}
```

Additionally, you can also use the `alias` option ([TF doc](https://www.terraform.io/docs/language/providers/configuration.html#alias-multiple-provider-configurations)):

```tf
# the default provider configuration
provider "fastly" {
  api_key = "xxxx"
}

provider "fastly" {
  alias   = "noauth"
  no_auth = true
}
data "fastly_ip_ranges" "test" {
  provider = fastly.noauth
}
```

We appreciate your feedback.

PR #444, opened 2021-07-16
Introduce "no_auth" provider option

smaeda-ks:
For: https://github.com/fastly/terraform-provider-fastly/issues/86

### DIFF
--- a/docs/data-sources/ip_ranges.md
+++ b/docs/data-sources/ip_ranges.md
@@ -13,6 +13,10 @@ Use this data source to get the [IP ranges][1] of Fastly edge nodes.
 ## Example Usage
 
 ```hcl
+provider "fastly" {
+  no_auth = true
+}
+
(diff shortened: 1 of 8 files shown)


Integralist (approved):
My only question is do we expect many users to use the Fastly Terraform provider for a single data source that doesn't require authentication?

Either way I think this solves the problem in the only way possible considering we have to construct the client up front before any of the data sources or resources are initialised/called (so it's not like we could avoid having an extra attribute like `no_auth`).

So 👍🏻 from me. Quick ping to @bengesoff for his opinions.

smaeda-ks:
Exactly. There seems to be no way to instantiate the client conditionally for a given data source. We could just create a new instance manually in `data_source_ip_ranges.go` but then we're unable to attach it to `meta` and it won't provide debug logs (the raw request details), so avoiding such special handling.

bengesoff (approved):

Issue #112, opened 2019-01-03
Create VCL as main in one API call


miyagawa:
Hi,

Once https://github.com/sethvargo/go-fastly/pull/93 is merged, this provider should make use of the `Main` attribute of the VCL input so that we can avoid creating a VCL as non-main and then promote it to main in a separate API call.

Here's an untested patch:

```
diff --git a/fastly/resource_fastly_service_v1.go b/fastly/resource_fastly_service_v1.go
index c1a65da..1313bef 100644
--- a/fastly/resource_fastly_service_v1.go
+++ b/fastly/resource_fastly_service_v1.go
@@ -2503,6 +2503,7 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 					Version: latestVersion,
 					Name:    df["name"].(string),
 					Content: df["content"].(string),
+					Main:    df["main"].(bool),
 				}
 
 				log.Printf("[DEBUG] Fastly VCL Addition opts: %#v", opts)
@@ -2510,21 +2511,6 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 				if err != nil {
 					return err
 				}
-
-				// if this new VCL is the main
-				if df["main"].(bool) {
-					opts := gofastly.ActivateVCLInput{
-						Service: d.Id(),
-						Version: latestVersion,
-						Name:    df["name"].(string),
-					}
-					log.Printf("[DEBUG] Fastly VCL activation opts: %#v", opts)
-					_, err := conn.ActivateVCL(&opts)
-					if err != nil {
-						return err
-					}
-
-				}
 			}
 		}
 
```

miyagawa:
the go-fastly PR is merged now.

smaeda-ks:
Added in v0.33.0:
https://github.com/fastly/terraform-provider-fastly/blob/main/CHANGELOG.md

Thank you for the suggestion! @miyagawa

PR #446, opened 2021-07-17
set main field on CreateVCL to reduce redundant subsequent call

smaeda-ks:
For: https://github.com/fastly/terraform-provider-fastly/issues/112


Integralist (approved):
LGTM. 

**Side Note**: The `ActivateVCL` function name in go-fastly is terrible IMHO and makes it hard to know what the function is actually doing (I had to look it up in developer.fastly.com). We should fix that in future.

PR #448, opened 2021-07-21
Simplify Service import logic, and allow selecting a version while importing

bengesoff:
Previously the Service resource import would do a full state refresh twice: once in the `resourceImport` function, and again when Terraform calls `resourceRead` again afterwards. This PR simplifies this, to only optionally set the `cloned_version` in the `resourceImport` function, and delegate the rest of the state refresh to the `resourceRead` called by Terraform automatically.

If no version is specified for the import, then the Read function will detect `cloned_version` is 0, and select either the active version or latest version. This logic hasn't changed, but the `isImport` flag argument is no longer needed.

In simplifying the import logic, it seemed like a good opportunity to add a new feature: allowing `@` and a version number to be appended to the ID whilst importing, to enable the user to choose a version. This is optional, and the previous method of only specifying the ID without the version is still supported.


templates/resources/components/footer.md.tmpl
```diff
@@ -13,4 +13,14 @@ $ terraform import fastly_service_v1.demo xxxxxxxxxxxxxxxxxxxx
 ```{{end}}
 {{ if eq .Data.ServiceType "wasm"}}```
 $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx
+```{{end}}
+
+By default, either the active version will be imported, or the latest version if no version is active.
+Alternatively, a specific version of the service can be selected by appending an `@` followed by the version number to the service ID, e.g.
+
+{{ if eq .Data.ServiceType "vcl"}}```
+$ terraform import fastly_service_v1.demo xxxxxxxxxxxxxxxxxxxx@2
```
Integralist, line 22: Oh interesting. How is `xxxxxxxxxxxxxxxxxxxx` replaced with the mock Service ID?

Integralist (approved):
LGTM. Just one unrelated question @bengesoff

PR #449, opened 2021-07-22
don't use ParallelTest for no_auth data source

smaeda-ks:
... when running full acceptance tests.

https://www.terraform.io/docs/language/providers/configuration.html#default-provider-configurations

It looks like when tests are running in parallel to the IP ranges data source test, sometimes, other tests are getting polluted with an instance of the provider with no API key.

### DIFF
--- a/fastly/data_source_ip_ranges_test.go
+++ b/fastly/data_source_ip_ranges_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 func TestAccFastlyIPRanges(t *testing.T) {
-	resource.ParallelTest(t, resource.TestCase{
+	resource.Test(t, resource.TestCase{
 		ProviderFactories: testAccProviders,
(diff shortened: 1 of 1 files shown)


Integralist (approved):
LGTM. @bengesoff could you please give a 👍🏻  here once the local running of the test suite has finished and is showing as passing. Thanks!

bengesoff:
@Integralist unfortunately a 👎🏼 but having some possible luck with changing `ParallelTest` to `Test`

bengesoff:
The tests look good with the `ParallelTest` replacement - the `alias` stuff doesn't appear to be needed. I'd say we should also add a comment to explain why the test can't be run in parallel. I'm happy to make the update @smaeda-ks if you'd like?

bengesoff (approved):

PR #452, opened 2021-07-29
Make backend `auto_loadbalance` default to false

bengesoff:
Update of #255 to work with latest version of provider.

> The Fastly web interface defaults "Auto load balance" to "No". The most
> common reason for having multiple backends in a single service is to
> route different paths to different backends, rather than load balance
> between different origins. Currently, this provider defaults
> `auto_loadbalance` to true, which is surprising and confusing. This
> commit makes `auto_loadbalance` default to false to reduce confusion.


Integralist (approved):
LGTM. But won't merge until we're ready for v1.0.0

Hiieu:
The changes in this PR have been merged into the v1 branch in the OC fork: https://github.com/opencredo/terraform-provider-fastly/pull/111

Integralist:
Closing as per the above comment. This is because we're consolidating all changes destined for `v1.0.0` of the Fastly Terraform provider into the [opencredo/v1](https://github.com/opencredo/terraform-provider-fastly/tree/v1) branch and will then soon be merged into the Fastly [v1](https://github.com/fastly/terraform-provider-fastly/tree/v1) branch, which will at some near point in the future make its way into `main` for a new release to be cut.

PR #453, opened 2021-07-29
use d.SetNewComputed for TypeSet computed attributes

### LINKED ISSUE #451: Updates to TLS Subscription Doesn't update Managed DNS Challenges Value
The recent change to support multiple domain DNS challenges was great. Thank you.

However, it appears, after the first seed, and when updating the domains from X to X+1, the `managed_dns_challenges` output doesn't increase it's output set to include new values on the corresponding plan/apply. Thus any subsequent dependencies (e.g. Route 53, won't update until a future run).

```
resource "fastly_tls_subscription" "default" {
  domains               = ["test1.example.com"]
  force_update          = true
  force_destroy         = true
}
```

```
resource "aws_route53_record" "domain_validation_records" {
   for_each = { for domain in fastly_tls_subscription.default.managed_dns_challenges : domain.record_name => domain }
   ...

   depends_on = [ fastly_tls_subscription.default ]
}
```

Then change...
```
  domains               = ["test1.example.com", "test2.example.com"]
```

The next plan/apply will update the subscription but **not** any changes for the Route53 records. In order to apply the domain validation record, an additional apply must be run after the current apply completes.

Additionally, once applied, Terraform will return immediately. However, Fastly, takes 1-10 minutes to provision appropriately.

This also causes an issue when going from a long list to small list. It will actually cause Terraform to fail.

----

I just retested the example above an in the documentation ... It appears on the very first run (no subscription, no DNS entries), it also errors also:

```
Error: Invalid for_each argument
  on dns.tf line 65, in resource "aws_route53_record" "domain_validation_records":
  65:   for_each = { for domain in fastly_tls_subscription.default.managed_dns_challenges : domain.record_name => domain }

The "for_each" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the for_each depends on.
```

The above is virtually the same as the documentation. This could be avoided as one could key the above with domains that are specified in the TLS record. However, if you do this, it doesn't work when updating the the domains of a TLS subscription due to the bug listed initially.

### Thus, recommendation:
1) Fix the `managed_dns_challenges` output to return proper values and count on a plan/apply after a change is made to the `domains` field.
2) Update the documentation for domain validation Route53 for_each loop to key using a locally defined array that is passed into the TLS subscription `domains` as that should correspond 1:1:1 all the way through.

### Fastly Provider
`0.33.0`

### Terraform Version
`0.14.6`

### Affected Resource(s)
- `fastly_tls_subscription`

smaeda-ks:
This attempts to fix: https://github.com/fastly/terraform-provider-fastly/issues/451

TODO:
- [x] Update example https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/tls_subscription#example-usage

Note:
I'm not adding acceptance tests for this change because the update action only runs for "issued" state:
https://github.com/fastly/terraform-provider-fastly/blob/v0.33.0/fastly/resource_fastly_tls_subscription.go#L23

If we're about to add tests, that require tests to interact with DNS records as well as waiting for the subscription to be issued, which isn't ideal in my view.

Reference:
https://github.com/hashicorp/terraform-plugin-sdk/issues/195

### DIFF
--- a/fastly/resource_fastly_tls_subscription.go
+++ b/fastly/resource_fastly_tls_subscription.go
@@ -25,6 +25,7 @@ func resourceFastlyTLSSubscription() *schema.Resource {
 			customdiff.ForceNewIf("configuration_id", resourceFastlyTLSSubscriptionStateNotIssued),
 			customdiff.ForceNewIf("domains", resourceFastlyTLSSubscriptionStateNotIssued),
 			customdiff.ForceNewIf("common_name", resourceFastlyTLSSubscriptionStateNotIssued),
+			resourceFastlyTLSSubscriptionSetNewComputed,
 		),
 		Schema: map[string]*schema.Schema{
 			"domains": {
@@ -329,4 +330,18 @@ func resourceFastlyTLSSubscriptionDelete(_ context.Context, d *schema.ResourceDa
 
 func resourceFastlyTLSSubscriptionStateNotIssued(_ context.Context, d *schema.ResourceDiff, _ interface{}) bool {
 	return d.Get("state").(string) != "issued"
+}
+
+func resourceFastlyTLSSubscriptionSetNewComputed(_ context.Context, d *schema.ResourceDiff, meta interface{}) error {
+	// NOTE: This is a workaround for a bug in Terraform core
+	// where TypeSet computed attributes are not being updated with the new values upon applying (in an update action).
+	// This means that they will not be updated until the second "refresh" or "apply" after the first apply.
+	// We should work around this and set the new values immediately upon applying so that other resources
+	// that are dependent on this resource can properly see the diff and trigger updates accordingly upon applying.
+	if d.HasChange("domains") {
+		d.SetNewComputed("managed_dns_challenges")
+		d.SetNewComputed("managed_http_challenges")
+	}
+
+	return nil
 }
--- a/fastly/resource_fastly_tls_subscription.go
+++ b/fastly/resource_fastly_tls_subscription.go
@@ -332,7 +332,7 @@ func resourceFastlyTLSSubscriptionStateNotIssued(_ context.Context, d *schema.Re
 	return d.Get("state").(string) != "issued"
 }
 
-func resourceFastlyTLSSubscriptionSetNewComputed(_ context.Context, d *schema.ResourceDiff, meta interface{}) error {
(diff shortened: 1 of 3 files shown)


fastly/resource_fastly_tls_subscription.go
```diff
@@ -330,3 +331,17 @@ func resourceFastlyTLSSubscriptionDelete(_ context.Context, d *schema.ResourceDa
 func resourceFastlyTLSSubscriptionStateNotIssued(_ context.Context, d *schema.ResourceDiff, _ interface{}) bool {
 	return d.Get("state").(string) != "issued"
 }
+
+func resourceFastlyTLSSubscriptionSetNewComputed(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
+	// NOTE: This is a workaround for a bug in Terraform core
```
Integralist: Can we link to an issue? I presume there's an open issue on the Hashicorp Terraform GitHub repo?
```suggestion
	// NOTE: This is a workaround for a bug in Terraform core (hashicorp/terraform-plugin-sdk#195)
```

smaeda-ks: I added the link in the description above. It's still my _assumption_ so I'm trying to clarify that part.

fastly/resource_fastly_tls_subscription.go
```diff
@@ -25,6 +25,7 @@ func resourceFastlyTLSSubscription() *schema.Resource {
 			customdiff.ForceNewIf("configuration_id", resourceFastlyTLSSubscriptionStateNotIssued),
 			customdiff.ForceNewIf("domains", resourceFastlyTLSSubscriptionStateNotIssued),
 			customdiff.ForceNewIf("common_name", resourceFastlyTLSSubscriptionStateNotIssued),
+			resourceFastlyTLSSubscriptionSetNewComputed,
```
bengesoff, line 28: It might be marginally more readable to use the `customdiff.ComputedIf()` helper method instead of calling `d.SetNewComputed()` directly? I did a similar thing in #388 but up to you

Integralist (approved):
LGTM.

virgofx:
Any chance we can get this merged and a new point release created? Would love to test this in our environment :)

cc: @smaeda-ks @Integralist

Integralist:
Hi @virgofx I'll be cutting a new release on Monday morning. Thanks!

PR #456, opened 2021-07-31
deprecate geo_headers attribute

smaeda-ks:
https://developer.fastly.com/reference/api/vcl-services/request-settings/

We've internally confirmed that the setting `geo_headers` parameter had been exposed without any actual implementation, and hence setting this parameter today basically does nothing. Deprecating this parameter.

Screen Shot 2021-08-01 at 5 55 05


Integralist (approved):

si14:
Sorry @smaeda-ks , I didn't reply to the ticket to thank you in time. Thank you for investigating and quickly fixing it!

Issue #454, opened 2021-07-29
optional backend block


marcb1:
For the resource `fastly_service_v1`, it seems that the `backend` argument is required, this is mentioned in the docs here: https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_v1#required

The backend arg wasn't required in old versions of this plugin (namely `0.26.0`), and we have a few services that define the backend in the actual VCL (as this is allowed by the fastly API).

Is there a workaround to this, or is it possible to make the `backend` arg optional?

Currently getting this error when trying to upgrade to the latest version of the plugin:
```
The argument "backend" is required, but no definition was found.
```

smaeda-ks:
@marcb1 Good point. Thanks for pointing that out. I will look into that.

smaeda-ks:
@marcb1 This will be released in the next version.

PR #457, opened 2021-07-31
make backend block optional

smaeda-ks:
This addresses: https://github.com/fastly/terraform-provider-fastly/issues/454

Note: since this changes it from required to optional, this is NOT considered as a breaking change.

### DIFF
--- a/docs/resources/service_v1.md
+++ b/docs/resources/service_v1.md
@@ -236,7 +236,6 @@ $ terraform import fastly_service_v1.demo xxxxxxxxxxxxxxxxxxxx@2
 
 ### Required
 
-- **backend** (Block Set, Min: 1) (see [below for nested schema](#nestedblock--backend))
 - **domain** (Block Set, Min: 1) A set of Domain names to serve as entry points for your Service (see [below for nested schema](#nestedblock--domain))
 - **name** (String) The unique name for the Service to create
 
@@ -244,6 +243,7 @@ $ terraform import fastly_service_v1.demo xxxxxxxxxxxxxxxxxxxx@2
 
 - **acl** (Block Set) (see [below for nested schema](#nestedblock--acl))
 - **activate** (Boolean) Conditionally prevents the Service from being activated. The apply step will continue to create a new draft version but will not activate it if this is set to `false`. Default `true`
+- **backend** (Block Set) (see [below for nested schema](#nestedblock--backend))
 - **bigquerylogging** (Block Set) (see [below for nested schema](#nestedblock--bigquerylogging))
 - **blobstoragelogging** (Block Set) (see [below for nested schema](#nestedblock--blobstoragelogging))
 - **cache_setting** (Block Set) (see [below for nested schema](#nestedblock--cache_setting))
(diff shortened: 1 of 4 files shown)


fastly/block_fastly_service_v1_backend.go
```diff
@@ -415,7 +420,8 @@ func (h *BackendServiceAttributeHandler) Register(s *schema.Resource) error {
 
 	s.Schema[h.GetKey()] = &schema.Schema{
 		Type:     schema.TypeSet,
-		Required: true,
+		Required: required,
+		Optional: !required,
```
Integralist, line 424: Two quick questions:

1. Do we need to provide _both_ fields, e.g. can't we just replace `Required` with `Optional`?
2. If we can just set `Optional`, then would the logic be simpler (and would it still work) if we reversed it?

e.g. 

```terraform
optional := false

if h.GetServiceMetadata().serviceType == ServiceTypeVCL {
  optional = true
  ...
}

  Optional: optional,
  ...
}
```

fastly/resource_fastly_service_v1_test.go
```diff
@@ -349,6 +349,62 @@ func TestAccFastlyServiceV1_updateInvalidBackend(t *testing.T) {
...
+					testAccCheckFastlyServiceV1Attributes_backends(&service, name, []string{}),
+					resource.TestCheckResourceAttr(
+						"fastly_service_v1.foo", "active_version", "1"),
+					resource.TestCheckResourceAttr(
+						"fastly_service_v1.foo", "backend.#", "0"),
+				),
+			},
+			{
+				Config:      testAccServiceV1Config_staticBackend(name, domain, snippet2),
+				ExpectError: regexp.MustCompile("No backends"),
```
Integralist, line 402: Why do we expect an error here? If a backend is optional with a VCL service, then wouldn't having no backend defined inside the custom VCL not matter?

Integralist (commented):
Just a couple of quick clarifying questions.

fastly/resource_fastly_service_v1_test.go
smaeda-ks: @Integralist The backend "UI" object is optional, to be exact. Our compiler will definitely check if there's at least one `backend {...}` definition and throw an error if not.

fastly/block_fastly_service_v1_backend.go
smaeda-ks: I guess that works too. Let me test

smaeda-ks: @Integralist Looks like TF core requires an explicit set as I did. Otherwise, error:

>        1 error occurred:
>        	* resource fastly_service_v1: backend: One of optional, required, or
>        computed must be set

so I leave that as-is.

Integralist: @smaeda-ks Ah ok, the error is still a bit weird to me because I wouldn't have expected there to have been an issue with the logic per se.

fastly/resource_fastly_service_v1_test.go
Integralist: @smaeda-ks ah ok, so as long as the user has a backend defined within the VCL snippet then the VCL compiler won't complain.

Integralist (approved):

bengesoff:
@Integralist @smaeda-ks I'm running the `TestAccFastlyServiceV1_createServiceWithStaticBackend` test and getting a failure as there doesn't seem to be an error returned in the second step 🤔

smaeda-ks:
Hmm, it should. I'll check tomorrow

smaeda-ks:
https://github.com/fastly/terraform-provider-fastly/pull/462

docs/resources/service_v1.md
```diff
@@ -236,14 +236,14 @@ $ terraform import fastly_service_v1.demo xxxxxxxxxxxxxxxxxxxx@2
...
 
-- **backend** (Block Set, Min: 1) (see [below for nested schema](#nestedblock--backend))
 - **domain** (Block Set, Min: 1) A set of Domain names to serve as entry points for your Service (see [below for nested schema](#nestedblock--domain))
 - **name** (String) The unique name for the Service to create
 
 ### Optional
 
 - **acl** (Block Set) (see [below for nested schema](#nestedblock--acl))
 - **activate** (Boolean) Conditionally prevents the Service from being activated. The apply step will continue to create a new draft version but will not activate it if this is set to `false`. Default `true`
+- **backend** (Block Set) (see [below for nested schema](#nestedblock--backend))
```
chenrui333, line 246: just curious, why backend is optional now? what would be the use case when a service does not have no backends?

smaeda-ks: Well, beacon termination is a good example. They just want logs.

PR #458, opened 2021-08-05
Audit sensitive attributes

Hiieu:
Hi there 👋 ,

in this PR we've updated sensitive attributes.

### DIFF
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -408,8 +408,8 @@ Optional:
 - **pipeline** (String) The ID of the Elasticsearch ingest pipeline to apply pre-process transformations to before indexing
 - **request_max_bytes** (Number) The maximum number of logs sent in one request. Defaults to `0` for unbounded
 - **request_max_entries** (Number) The maximum number of bytes sent in one request. Defaults to `0` for unbounded
-- **tls_ca_cert** (String, Sensitive) A secure certificate to authenticate the server with. Must be in PEM format
-- **tls_client_cert** (String, Sensitive) The client certificate used to make authenticated requests. Must be in PEM format
+- **tls_ca_cert** (String) A secure certificate to authenticate the server with. Must be in PEM format
+- **tls_client_cert** (String) The client certificate used to make authenticated requests. Must be in PEM format
 - **tls_client_key** (String, Sensitive) The client private key used to make authenticated requests. Must be in PEM format
 - **tls_hostname** (String) The hostname used to verify the server's certificate. It can either be the Common Name (CN) or a Subject Alternative Name (SAN)
 - **user** (String) BasicAuth username for Elasticsearch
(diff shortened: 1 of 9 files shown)


Integralist (commented):
Interesting. I wouldn't necessarily have thought the email to be sensitive, but in some ways I guess this makes sense. @smaeda-ks any thoughts on that?

bengesoff:
@Integralist @Hiieu and I were just discussing this and thought we'd raise this for discussion for two reasons: firstly the API docs call this field `user` whereas in the provider we call it `email` (potential breaking change to align these), and also the current provider description field calls it a "secret" whereas, as you said, we wouldn't have necessarily thought it to be sensitive.

Ps I planning to add more context before we marked it "ready for review" but you beat me to it 😉

Integralist:
Thanks @bengesoff 🙂 

So I think the naming should be aligned with the API, so `email` should change to `user` (@smaeda-ks would you agree?)

I would probably follow the API documentation in this case (which has a completely different description 🤦🏻  I always prefer to copy the description rather than write up our own custom description)...

> Your Google Cloud Platform service account email address. The client_email field in your service account authentication JSON. Required.

Nothing about that suggests the user email is 'sensitive'? It's also returned in the API when doing a GET or UPDATE for example (which could well be a mistake on our part if this is supposed to be sensitive).

Reading about GCS 'service accounts' didn't suggest to me that the user email is sensitive either: https://cloud.google.com/iam/docs/service-accounts

smaeda-ks:
@Integralist 

> So I think the naming should be aligned with the API, so email should change to user (@smaeda-ks would you agree?)

Yes

As for the `sensitive` of the email address attribute, I tend to agree that we don't necessary to make this change.

Hiieu:
Thanks @Integralist and @smaeda-ks , I'll change the `email` attribute to `user` and desensitise it.
While I have your attention,  I've removed the `sensistive` attribute from the`tls_ca_cert` and  `tls_client_cert` fields in `fastly/block_fastly_service_v1_httpslogging.go` to keep consistency across the blocks.
But...now I've noticed that: 
1. [block_fastly_service_v1_logging_elasticsearch.go](https://github.com/fastly/terraform-provider-fastly/blob/main/fastly/block_fastly_service_v1_logging_elasticsearch.go#L232)
2. [block_fastly_service_v1_logging_kafka.go](https://github.com/fastly/terraform-provider-fastly/blob/main/fastly/block_fastly_service_v1_logging_kafka.go#L65)

have the same fields set as sensitive, so I wonder if we should desensitise these blocks as well? 🤔

smaeda-ks:
@Hiieu 

> have the same fields set as sensitive, so I wonder if we should desensitise these blocks as well? 🤔

Yes, client CA and cert are generally not considered as sensitive, but client "key". So I'm +1 with desensitise those tls attributes as well. Thanks for flagging that!

Hiieu:
All comments have been addressed:
1. Changed `email` to `user` 
2. Desensitised the ca and client certs

 The PR is ready for review.

Integralist:
Due to the change of `email` to `user` this sounds like a breaking change to me as it'll affect a user's configuration.

Integralist (approved):
LGTM. But because the `email` change to `user` is a breaking change, maybe we move that change to a separate PR otherwise these 'sensitive' audit changes won't be mergeable until we're ready to release a 1.0.0 release. 

Thoughts?

Hiieu:
@Integralist, sounds good, I can move the email change to a separate PR.

Hiieu:
The PR for the email change: https://github.com/fastly/terraform-provider-fastly/pull/461

Integralist:
Thanks @Hiieu let me know if this PR is ready to merge.

bengesoff (approved):

Hiieu:
@Integralist, it's ready to be merged 👍

PR #462, opened 2021-08-12
"no backends" now considered as warning not error in VCL service

smaeda-ks:
This is due to our recent internal change.


smaeda-ks:
cc: @bengesoff

bengesoff (approved):
That makes sense! Thought I was going crazy 😂

Integralist (approved):

PR #463, opened 2021-08-13
Refactor service attribute handlers into CRUD-style functions

bengesoff:
Given that the `fastly_service` resources have such a large number of nested blocks, which are mostly common to both the `fastly_service_v1` VCL service and the `fastly_service_compute` service resources, there is a `ServiceAttributeDefinition` abstraction to allow all of the attributes to be composed into each service resource. This means that each attribute defines a handler which implements the `ServiceAttributeDefinition` interface. The key methods in this interface are `Register`, which adds the attributes to the service's schema; `Read`, which refreshes the state of the attributes; and `Process`, which is responsible for creating, updating and deleting the API resources.

The problem with this, is now there are such a large number of these handlers (~40), it has emerged that there is a very common pattern for majority of them. In particular, as most of them are nested blocks, their type is `schema.TypeSet` and they all have almost identical `Process` functions. Below is an example of the one from the `domain` block:

```go
func (h *DomainServiceAttributeHandler) Process(ctx context.Context, d *schema.ResourceData, latestVersion int, conn *gofastly.Client) error {
	od, nd := d.GetChange(h.GetKey())
	if od == nil {
		od = new(schema.Set)
	}
	if nd == nil {
		nd = new(schema.Set)
	}
	oldSet := od.(*schema.Set)
	newSet := nd.(*schema.Set)
	setDiff := NewSetDiff(func(resource interface{}) (interface{}, error) {
		t, ok := resource.(map[string]interface{})
		if !ok {
			return nil, fmt.Errorf("resource failed to be type asserted: %+v", resource)
		}
		return t["name"], nil
	})
	diffResult, err := setDiff.Diff(oldSet, newSet)
	if err != nil {
		return err
	}
	// DELETE removed resources
	for _, resource := range diffResult.Deleted {
		resource := resource.(map[string]interface{})
		opts := gofastly.DeleteDomainInput{
			ServiceID:      d.Id(),
			Name:           resource["name"].(string),
		}
		log.Printf("[DEBUG] Fastly Domain removal opts: %#v", opts)
		err := conn.DeleteDomain(&opts)
		if errRes, ok := err.(*gofastly.HTTPError); ok {
			if errRes.StatusCode != 404 {
				return err
			}
		} else if err != nil {
			return err
		}
	}
	// CREATE new resources
	for _, resource := range diffResult.Added {
		resource := resource.(map[string]interface{})
		opts := gofastly.CreateDomainInput{
			ServiceID:      d.Id(),
			Name:           resource["name"].(string),
		}
		if v, ok := resource["comment"]; ok {
			opts.Comment = v.(string)
		}
		log.Printf("[DEBUG] Fastly Domain Addition opts: %#v", opts)
		_, err := conn.CreateDomain(&opts)
		if err != nil {
			return err
		}
	}
	// UPDATE modified resources
	//
	// NOTE: although the go-fastly API client enables updating of a resource by
	// its 'name' attribute, this isn't possible within terraform due to
	// constraints in the data model/schema of the resources not having a uid.
	for _, resource := range diffResult.Modified {
		resource := resource.(map[string]interface{})
		opts := gofastly.UpdateDomainInput{
			ServiceID:      d.Id(),
			Name:           resource["name"].(string),
		}
		// only attempt to update attributes that have changed
		modified := setDiff.Filter(resource, oldSet)
		if v, ok := modified["comment"]; ok {
			opts.Comment = gofastly.String(v.(string))
		}
		log.Printf("[DEBUG] Update Domain Opts: %#v", opts)
		_, err := conn.UpdateDomain(&opts)
		if err != nil {
			return err
		}
	}
	return nil
}
```

As you can see, the first part of the function deals with `SetDiff` to make sense of the nested blocks. This uses the `name` attribute to work out which set elements are new, modified, or removed. Then the rest of the function contains `for` loops to iterate through all of the created, updated, and deleted elements and perform the relevant CRUD method on each one. This pattern is exactly the same for all of the following files:

 - `fastly/block_fastly_service_v1_acl.go`
 - `fastly/block_fastly_service_v1_backend.go`
 - `fastly/block_fastly_service_v1_bigquerylogging.go`
 - `fastly/block_fastly_service_v1_blobstoragelogging.go`
 - `fastly/block_fastly_service_v1_cachesetting.go`
 - `fastly/block_fastly_service_v1_condition.go`
 - `fastly/block_fastly_service_v1_dictionary.go`
 - `fastly/block_fastly_service_v1_director.go`
 - `fastly/block_fastly_service_v1_domain.go`
 - `fastly/block_fastly_service_v1_dynamicsnippet.go`
 - `fastly/block_fastly_service_v1_gcslogging.go`
 - `fastly/block_fastly_service_v1_gzip.go`
 - `fastly/block_fastly_service_v1_header.go`
 - `fastly/block_fastly_service_v1_healthcheck.go`
 - `fastly/block_fastly_service_v1_httpslogging.go`
 - `fastly/block_fastly_service_v1_logentries.go`
 - `fastly/block_fastly_service_v1_logging_cloudfiles.go`
 - `fastly/block_fastly_service_v1_logging_datadog.go`
 - `fastly/block_fastly_service_v1_logging_digitalocean.go`
 - `fastly/block_fastly_service_v1_logging_elasticsearch.go`
 - `fastly/block_fastly_service_v1_logging_ftp.go`
 - `fastly/block_fastly_service_v1_logging_googlepubsub.go`
 - `fastly/block_fastly_service_v1_logging_heroku.go`
 - `fastly/block_fastly_service_v1_logging_honeycomb.go`
 - `fastly/block_fastly_service_v1_logging_kafka.go`
 - `fastly/block_fastly_service_v1_logging_kinesis.go`
 - `fastly/block_fastly_service_v1_logging_loggly.go`
 - `fastly/block_fastly_service_v1_logging_logshuttle.go`
 - `fastly/block_fastly_service_v1_logging_newrelic.go`
 - `fastly/block_fastly_service_v1_logging_openstack.go`
 - `fastly/block_fastly_service_v1_logging_scalyr.go`
 - `fastly/block_fastly_service_v1_logging_sftp.go`
 - `fastly/block_fastly_service_v1_papertrail.go`
 - `fastly/block_fastly_service_v1_requestsetting.go`
 - `fastly/block_fastly_service_v1_responseobject.go`
 - `fastly/block_fastly_service_v1_s3logging.go`
 - `fastly/block_fastly_service_v1_snippet.go`
 - `fastly/block_fastly_service_v1_splunk.go`
 - `fastly/block_fastly_service_v1_sumologic.go`
 - `fastly/block_fastly_service_v1_syslog.go`
 - `fastly/block_fastly_service_v1_vcl.go`

Firstly this is not the clearest format that each file could be in, as it is not very obvious where to find things without familiarity with the codebase, and secondly, it results in a lot of repeated code. Therefore this PR introduces a new abstraction for all of these similar files, which is hopefully clearer and reduces a lot of code repetition.

The centrepiece of the new abstraction is the `service_crud_attribute_definition.go` file, with its `ServiceCRUDAttributeDefinition` interface. The handlers listed above now all implement this interface instead of `ServiceAttributeDefinition` described above. Both of these interfaces have also both been moved to their own files instead of sitting inside the `base_fastly_service_v1.go` file.

```go
type ServiceCRUDAttributeDefinition interface {
	Key() string

	GetSchema() *schema.Schema

	Create(ctx context.Context, d *schema.ResourceData, resource map[string]interface{}, serviceVersion int, conn *gofastly.Client) error
	Read(ctx context.Context, d *schema.ResourceData, resource map[string]interface{}, serviceVersion int, conn *gofastly.Client) error
	Update(ctx context.Context, d *schema.ResourceData, resource, modified map[string]interface{}, serviceVersion int, conn *gofastly.Client) error
	Delete(ctx context.Context, d *schema.ResourceData, resource map[string]interface{}, serviceVersion int, conn *gofastly.Client) error
}
```

To make this work with the existing logic, the `ToServiceAttributeDefinition` function implements the `ServiceAttributeDefinition` interface using the same `SetDiff` and `for` loops in its `Process` function as before. The difference now, is that this code only appears once, and it takes care of calling into the handler's CRUD functions from the new interface. Apart from deduplicating the `SetDiff` logic, the other benefit of this is that each file hopefully reads a bit more as you'd expect from a full Terraform resource: the schema defined at the top, with four CRUD functions underneath. An excerpt of the `domain` block's file is shown below. It is similar to the `Process` function shown above, but with the `for` loops extracted into their own functions, and a lot of the boilerplate removed.

```go
type DomainServiceAttributeHandler struct {
	*DefaultServiceAttributeHandler
}

func NewServiceDomain(sa ServiceMetadata) ServiceAttributeDefinition {
	return ToServiceAttributeDefinition(&DomainServiceAttributeHandler{
		&DefaultServiceAttributeHandler{
			key:             "domain",
			serviceMetadata: sa,
		},
	})
}

func (h *DomainServiceAttributeHandler) Key() string { return h.key }

func (h *DomainServiceAttributeHandler) GetSchema() *schema.Schema {
	return &schema.Schema{
		Type:        schema.TypeSet,
		Required:    true,
		Description: "A set of Domain names to serve as entry points for your Service",
		Elem: &schema.Resource{
				"name": {
					Type:        schema.TypeString,
					Required:    true,
					Description: "The domain that this Service will respond to. It is important to note that changing this attribute will delete and recreate the resource.",
				},

				"comment": {
					Type:        schema.TypeString,
					Optional:    true,
					Description: "An optional comment about the Domain.",
				},
			},
		},
	}
}

func (h *DomainServiceAttributeHandler) Create(_ context.Context, d *schema.ResourceData, resource map[string]interface {
}, serviceVersion int, conn *gofastly.Client) error {
	opts := gofastly.CreateDomainInput{
		ServiceID:      d.Id(),
		ServiceVersion: serviceVersion,
		Name:           resource["name"].(string),
	}

	if v, ok := resource["comment"]; ok {
		opts.Comment = v.(string)
	}

	log.Printf("[DEBUG] Fastly Domain Addition opts: %#v", opts)
	_, err := conn.CreateDomain(&opts)
	if err != nil {
		return err
	}
	return nil
}

func (h *DomainServiceAttributeHandler) Read(_ context.Context, d *schema.ResourceData, _ map[string]interface{}, serviceVersion int, conn *gofastly.Client) error {
	log.Printf("[DEBUG] Refreshing Domains for (%s)", d.Id())
	domainList, err := conn.ListDomains(&gofastly.ListDomainsInput{
		ServiceID:      d.Id(),
		ServiceVersion: serviceVersion,
	})

	if err != nil {
		return fmt.Errorf("[ERR] Error looking up Domains for (%s), version (%v): %s", d.Id(), serviceVersion, err)
	}

	// Refresh Domains
	dl := flattenDomains(domainList)

	if err := d.Set(h.GetKey(), dl); err != nil {
		log.Printf("[WARN] Error setting Domains for (%s): %s", d.Id(), err)
	}
	return nil
}

func (h *DomainServiceAttributeHandler) Update(_ context.Context, d *schema.ResourceData, resource, modified map[string]interface {
}, serviceVersion int, conn *gofastly.Client) error {
	opts := gofastly.UpdateDomainInput{
		ServiceID:      d.Id(),
		ServiceVersion: serviceVersion,
		Name:           resource["name"].(string),
	}

	if v, ok := modified["comment"]; ok {
		opts.Comment = gofastly.String(v.(string))
	}

	log.Printf("[DEBUG] Update Domain Opts: %#v", opts)
	_, err := conn.UpdateDomain(&opts)
	if err != nil {
		return err
	}
	return nil
}

func (h *DomainServiceAttributeHandler) Delete(_ context.Context, d *schema.ResourceData, resource map[string]interface {
}, serviceVersion int, conn *gofastly.Client) error {
	opts := gofastly.DeleteDomainInput{
		ServiceID:      d.Id(),
		ServiceVersion: serviceVersion,
		Name:           resource["name"].(string),
	}

	log.Printf("[DEBUG] Fastly Domain removal opts: %#v", opts)
	err := conn.DeleteDomain(&opts)
	if errRes, ok := err.(*gofastly.HTTPError); ok {
		if errRes.StatusCode != 404 {
			return err
		}
	} else if err != nil {
		return err
	}
	return nil
}
```

As you can see, this was a lot of files to refactor and I didn't do all of them by hand. I've deleted the auto-refactoring code in one of the later commits so it doesn't pollute the main tree, but if you're interested in the steps taken to adapt the old code into this new code in the PR, see https://github.com/fastly/terraform-provider-fastly/blob/1e779790af95e899d047d1c2392c98c351f457f5/refactor_ast/main.go.

The aim of this PR is to make the code as clear as possible so any feedback is very welcome, whether it's naming suggestions, tweaking the abstraction, or anything else. Nothing should have changed functionally so there shouldn't be anything in terms of breaking changes. Let me know what you think! 😁


fastly/service_crud_attribute_definition.go
```diff
@@ -0,0 +1,131 @@
...
+
+	oldSet := oldVal.(*schema.Set)
+	newSet := newVal.(*schema.Set)
+
+	setDiff := NewSetDiff(func(resource interface{}) (interface{}, error) {
+		t, ok := resource.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("resource failed to be type asserted: %+v", resource)
+		}
+		return t["name"], nil
```
smaeda-ks, line 93: I actually just added this code:
https://github.com/fastly/terraform-provider-fastly/blob/main/fastly/block_fastly_waf_configuration_v1_active_rule.go#L52-L59

which uses a different attribute name than this known "name" attribute. Well, I guess that will be the only place we do such among resources but I wonder if we can make this value dynamic/settable.

bengesoff: That's very useful to know. I thought that may be a problem but checked and everything did have `name` attribute so I just left if statically defined as you've spotted. I think it could be settable - by defining a `GetUniqueKeyName` method in the interface or something, and having all the other blocks return "name".

Having said this though, this refactor only affects attributes on the service resource, whereas your change was on the `fastly_waf_configuration` resource and should therefore be isolated. As far as I'm aware, the `ServiceAttributeDefinition` interface is only used in the service resource because of the sheer number of nested blocks. 

If you think it might be likely that either another nested block could be added to the service resource without a `name` attribute, or that we might want to at some point expand the existing `ServiceAttributeDefinition` style of splitting up and composing nested blocks to other resources like `waf_configuration`, then I'm definitely happy to make this change.

fastly/service_crud_attribute_definition.go
```diff
@@ -0,0 +1,131 @@
...
+	Create(ctx context.Context, d *schema.ResourceData, resource map[string]interface{}, serviceVersion int, conn *gofastly.Client) error
+	// Read should refresh the state of all of the instance of the nested blocks. See the description of Create for more
+	// details about the arguments.
+	Read(ctx context.Context, d *schema.ResourceData, resource map[string]interface{}, serviceVersion int, conn *gofastly.Client) error
+	// Update should make changes to an existing instance of the nested block. The arguments are as described in the
+	// Create comments, with the exception of modified which will contain only the attributes that have changed.
+	Update(ctx context.Context, d *schema.ResourceData, resource, modified map[string]interface{}, serviceVersion int, conn *gofastly.Client) error
+	// Delete should remove the instance of the nested block. See the description of Create for more details about the
+	// arguments.
+	Delete(ctx context.Context, d *schema.ResourceData, resource map[string]interface{}, serviceVersion int, conn *gofastly.Client) error
```
Integralist: Should probably have some line breaks in-between the function signature and following comment.

fastly/service_crud_attribute_definition.go
```diff
@@ -0,0 +1,131 @@
...
+// convert it to the ServiceAttributeDefinition that the service resources expect.
+// The requirements for a service attribute to be defined in terms of this interface are:
+// - the attribute must be a nested block with a schema type of schema.TypeSet
+// - the nested block must have its own "name" attribute which uniquely defines the nested resource
+// - the block must support all four of the CRUD operations, or at least Create, Read, and Delete if updating the
+//   resource is not supported
+// Some service attributes don't fit into these constraints are better suited to implementing the
+// ServiceAttributeDefinition directly. One example is the "package" block in block_fastly_service_v1_package.go, which
+// only uses an Update operation, and therefore implements ServiceAttributeDefinition directly without
+// ServiceCRUDAttributeDefinition.
```
Integralist: ...can we have some line breaks around this section so we have some breathing space. Otherwise it's a pretty dense wall of text 🙂 

Maybe something like (yeah I'm an 80 column width kinda guy)...

```
// ServiceCRUDAttributeDefinition is an interface for most 
// ServiceAttributeDefinition implementations which can be represented by the 
// four CRUD operations. Most service attributes will fall into this category 
// and should implement this interface instead of ServiceAttributeDefinition 
// directly. 
//
// An attribute that implements ServiceCRUDAttributeDefinition can use the 
// ToServiceAttributeDefinition function defined below in its constructor to 
// convert it to the ServiceAttributeDefinition that the service resources 
// expect.
//
// The requirements for a service attribute to be defined in terms of this 
// interface are:
//
// - the attribute must be a nested block with a schema type of schema.TypeSet
// - the nested block must have its own "name" attribute which uniquely defines 
//   the nested resource
// - the block must support all four of the CRUD operations, or at least Create, 
//   Read, and Delete if updating the resource is not supported
//
// Some service attributes don't fit into these constraints are better suited to 
// implementing the ServiceAttributeDefinition directly. One example is the 
// "package" block in block_fastly_service_v1_package.go, which only uses an 
// Update operation, and therefore implements ServiceAttributeDefinition 
// directly without ServiceCRUDAttributeDefinition.
```

Integralist (approved):
Approving after checking a couple of the initial updated files. It's not practical to review such a large refactoring PR and so we're going to have to rely heavily on the existing test suite. 

@bengesoff can you confirm you have run the test suite and seen a complete pass?

fastly/service_crud_attribute_definition.go
bengesoff: Great point, I've updated 👍🏼 I stuck it as 120 cols as I think that's what the rest of the codebase seems to be but let me know if you want me to cut it down to 80!

Ps something fishy is going on with the build 🤔

Integralist: >  I stuck it as 120 cols as I think that's what the rest of the codebase seems to be

No problem, better for consistency 👍🏻 

Yeah I was going to say the builds appear to be failing. Are the tests passing locally for you then?

bengesoff: Yeah it looks like it's checking out some weird commit before running the tests? I'll see if I can get to the bottom of it

bengesoff:
For the acceptance tests, I'm getting the following failures:
```
TestAccResourceFastlyTLSSubscription
TestAccFastlyServiceV1_createServiceWithStaticBackend
```

I think they're both related to changes on the `main` branch so it might be worth bringing some of those changes in to confirm. The first one is on the subscription resource so unrelated to this PR, and the second one appears to be linked to the backend becoming optional and no longer giving an error.

fastly/service_crud_attribute_definition.go
bengesoff: Ok it's failing on the merged result 🤦🏼

bengesoff: Aha it's due to the changes in #466 getting automerged into these changes incorrectly

bengesoff:
@Integralist I've rebased and integrated the changes from #466 and the tests seem to have been resolved. Re: my last comment, the `TestAccFastlyServiceV1_createServiceWithStaticBackend` has now been fixed, although the `TestAccResourceFastlyTLSSubscription` test is still failing.

bengesoff:
It appears the `TestAccResourceFastlyTLSSubscription` test is failing due to a change in the API to change all the TLS domains to lower case, unrelated to this PR

PR #464, opened 2021-08-14
Fix fastly_service_waf_configuration not updating rule attributes correctly

smaeda-ks:
from code comment:

> // NOTE: Fastly WAF (WAF 2020) API doesn't have a proper batch update endpoint.
> // go-fastly uses the below endpoint for UpsertBatchOperation:
> // "POST /waf/firewalls/:firewall_id/versions/:version_id/active-rules"
> // but this endpoint only updates "status" when it comes to upsert and "revision" field is ignored.
> // Therefore, when one of the rule attributes is changed we must delete it first and create it as a new rule.

### DIFF
--- a/fastly/block_fastly_waf_configuration_v1_active_rule.go
+++ b/fastly/block_fastly_waf_configuration_v1_active_rule.go
@@ -1,6 +1,7 @@
 package fastly
 
 import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


Integralist (approved):
LGTM.

Issue #465, opened 2021-08-17
Version comment in terraform plan/apply taken from wrong version


gtrog:
### Terraform Version
```
Terraform v0.13.7
+ provider registry.terraform.io/fastly/fastly v0.34.0
+ provider registry.terraform.io/hashicorp/vault v2.19.1

Your version of Terraform is out of date! The latest version
is 1.0.4. You can update by downloading from https://www.terraform.io/downloads.html
```
(Upgrading to 1.0.4 is not an easy option for us since it would affect our other terraform projects)

### Affected Resource(s)
Please list the resources as a list, for example:
- `fastly_service_v1`

### Plan Output
```
Terraform will perform the following actions:

  # module.prod.fastly_service_v1.service will be updated in-place
  ~ resource "fastly_service_v1" "service" {
        activate        = true
        active_version  = 76
        cloned_version  = 76
        comment         = "Managed by Terraform"
        default_ttl     = 3600
        id              = "2ZF2ARA4SIl8aAcPEHkwQK"
        name            = "Production"
      - version_comment = "DO NOT ACTIVATE THIS" -> null

        backend {
```

### Expected Behavior
Performing a plan/apply should not pick up `version_comment` from a different version than the current one being cloned.

### Actual Behavior
When performing a plan/apply, the `version_comment` value in the diff is coming from the latest draft version 78:
![image](https://user-images.githubusercontent.com/4908576/129742906-dcd17222-a209-4178-9acd-01c27d0facbc.png)

When I apply these changes, nothing changes to version 76 (since the comment is already empty string), leading to a perpetual diff where subsequent plan/applies still claim it's changing the `version_comment` field.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Have an active version without a version comment
2. Create a draft from this active version and set a version comment
3. Do a terraform plan/apply and verify that the version comment from the draft appears instead of the current active version's

smaeda-ks:
Thanks, I'll take a look.

smaeda-ks:
This has been fixed in: https://github.com/fastly/terraform-provider-fastly/releases/tag/v0.35.0

PR #466, opened 2021-08-24
Correctly update `version_comment`

smaeda-ks:
https://github.com/fastly/terraform-provider-fastly/issues/465

This PR fixes two issues.

1. `version_comment` attribute is ignored on service creation
2. `version_comment` value is wrongly assigned from `s.Version.Comment` as opposed to `s.ActiveVersion.Comment`
`s.Version` contains the last version object regardless of its state (draft/active), so we always should look at `s.ActiveVersion`

### DIFF
--- a/fastly/base_fastly_service_v1.go
+++ b/fastly/base_fastly_service_v1.go
@@ -361,7 +361,7 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta int
 	}
 
 	// Update the cloned version's comment. No new version is required for this.
-	if d.HasChange("version_comment") && !needsChange {
+	if d.HasChange("version_comment") && (!needsChange || isCreate) {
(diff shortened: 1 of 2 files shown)


Integralist (approved):

bengesoff:
A future improvement might be worth looping through the versions to find the one matching `"cloned_version"` and using the comment off that. Using `s.ActiveVersion.Comment` fixes the case when `"activate"` is true, but there could be some disparity still when `"activate"` is false.

smaeda-ks:
@bengesoff could you please elaborate on that more? I'm not following your point... sorry
In which scenario do we expect that to happen?

---

[updated]

I think I get your point. Thanks

bengesoff:
@smaeda-ks no problem - I think your fix is a definite improvement from where it was before, but I just figured it might be another edge case to catch 😄

PR #307, opened 2020-09-01
Changing formatVersion in VCLLoggingAttributes to be a *uint

matpimenta:
As some logging endpoints on go-fastly use uint and some *uint, and
VCLLoggingAttributes needs to cater for both.
The most generic and recommended way to express empty value is using nil
instead of zero.
For the Request objects that are still using uint, the provider
will set the value to 0 (zero) when the VCLLoggingAttributes.formatVersion
pointer is nil. This is compatible with the previous behaviour.

Request objects using *uint which were failing before (as zero was being
sent instead of nill) are now working again.

<details><summary>Acceptance Test results</summary>
<p>

```
=== RUN   TestResourceFastlyFlattenAcl
--- PASS: TestResourceFastlyFlattenAcl (0.00s)
=== RUN   TestResourceFastlyFlattenBigQuery
--- PASS: TestResourceFastlyFlattenBigQuery (0.27s)
=== RUN   TestResourceFastlyFlattenBlobStorage
--- PASS: TestResourceFastlyFlattenBlobStorage (0.00s)
=== RUN   TestAccFastlyServiceV1_bigquerylogging
--- PASS: TestAccFastlyServiceV1_bigquerylogging (101.76s)
=== RUN   TestAccFastlyServiceV1_bigquerylogging_compute
--- PASS: TestAccFastlyServiceV1_bigquerylogging_compute (104.58s)
=== RUN   TestAccFastlyServiceV1_bigquerylogging_env
--- PASS: TestAccFastlyServiceV1_bigquerylogging_env (111.36s)
=== RUN   TestAccFastlyServiceV1_acl
--- PASS: TestAccFastlyServiceV1_acl (123.37s)
=== RUN   TestResourceFastlyFlattenCacheSettings
--- PASS: TestResourceFastlyFlattenCacheSettings (0.00s)
=== RUN   TestAccFastlyServiceV1_blobstoragelogging_basic_compute
--- PASS: TestAccFastlyServiceV1_blobstoragelogging_basic_compute (81.64s)
=== RUN   TestResourceFastlyFlattenConditions
--- PASS: TestResourceFastlyFlattenConditions (0.00s)
=== RUN   TestAccFastlyServiceV1_blobstoragelogging_default
--- PASS: TestAccFastlyServiceV1_blobstoragelogging_default (99.44s)
=== RUN   TestResourceFastlyFlattenDictionary
--- PASS: TestResourceFastlyFlattenDictionary (0.00s)
[...]

PR #304, opened 2020-08-20
Keeping backends in directors when backends are modified

matpimenta:
Fixes #143 

This change is based on the Option 2 described on #197 and some further analysis to determine the best option.

I've made the Diff computation a separate component so that in the future, other parts of the code can re-use it to perform updates instead of delete/create.

I've also fixed some unrelated issues with logging endpoints on the compute resource (see commit cc7bec3) which was causing the acceptance tests to fail.

<details><summary>Acceptance Test results</summary>
<p>

```
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v  -timeout 360m -ldflags="-X=github.com/fastly/terraform-provider-fastly/version.ProviderVersion=acc"
?       github.com/terraform-providers/terraform-provider-fastly        [no test files]
=== RUN   TestResourceFastlyFlattenAcl
--- PASS: TestResourceFastlyFlattenAcl (0.00s)
=== RUN   TestAccFastlyServiceV1_acl
--- PASS: TestAccFastlyServiceV1_acl (76.48s)
=== RUN   TestAccFastlyServiceV1_bigquerylogging
--- PASS: TestAccFastlyServiceV1_bigquerylogging (70.51s)
=== RUN   TestAccFastlyServiceV1_bigquerylogging_compute
--- PASS: TestAccFastlyServiceV1_bigquerylogging_compute (58.07s)
=== RUN   TestAccFastlyServiceV1_bigquerylogging_env
--- PASS: TestAccFastlyServiceV1_bigquerylogging_env (70.25s)
=== RUN   TestResourceFastlyFlattenBigQuery
--- PASS: TestResourceFastlyFlattenBigQuery (0.18s)
=== RUN   TestResourceFastlyFlattenBlobStorage
--- PASS: TestResourceFastlyFlattenBlobStorage (0.00s)
=== RUN   TestAccFastlyServiceV1_blobstoragelogging_basic
--- PASS: TestAccFastlyServiceV1_blobstoragelogging_basic (131.49s)
=== RUN   TestAccFastlyServiceV1_blobstoragelogging_basic_compute
--- PASS: TestAccFastlyServiceV1_blobstoragelogging_basic_compute (51.44s)
=== RUN   TestAccFastlyServiceV1_blobstoragelogging_default
--- PASS: TestAccFastlyServiceV1_blobstoragelogging_default (63.51s)
=== RUN   TestAccFastlyServiceV1_blobstoragelogging_env
[...]

PR #477, opened 2021-09-24
Bump go-fastly to v5

Integralist:
**NOTE**: There were breaking changes in v4 and v5 of go-fastly. I've bumped the go-fastly version and applied the relevant interface changes to the provider.

- Switch from `bool` to `Compatibool` for ACL entries [#307](https://github.com/fastly/go-fastly/pull/307)
- Change Backend `SSLCiphers` field from `[]string` to `string` [#304](https://github.com/fastly/go-fastly/pull/304)


smaeda-ks:
@Integralist Thank you

Can we also update the description for this field?
https://github.com/fastly/terraform-provider-fastly/blob/v0.35.0/fastly/block_fastly_service_v1_backend.go#L122-L127

I believe we support `:` `,` and whitespace as a separator:
https://www.openssl.org/docs/manmaster/man1/openssl-ciphers.html

> The cipher list consists of one or more cipher strings separated by colons. Commas or spaces are also acceptable separators but colons are normally used.

We can basically use this exact same text... :)

ref: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1l/ssl/ssl_ciph.c#L604-L605

smaeda-ks (approved):

PR #479, opened 2021-09-29
Remove TLS subscriptions that 404 from state

bendrucker:
When a TLS subscription is removed outside of Terraform, currently it causes refreshes to fail

```
╷
│ Error: 404 - Not Found:
│ 
│     Title:  Not found
│     Detail: The subscription you requested was not found
```

Instead, the provider should remove these resources from state. This is handled in other resources but the exact implementation differs a bit by resource.

### DIFF
--- a/fastly/resource_fastly_tls_subscription.go
+++ b/fastly/resource_fastly_tls_subscription.go
@@ -3,12 +3,12 @@ package fastly
 import (
 	"context"
 	"fmt"
-	"log"
 	"strings"
 	"time"
 
 	"github.com/fastly/go-fastly/v5/fastly"
 	gofastly "github.com/fastly/go-fastly/v5/fastly"
+	"github.com/hashicorp/go-cty/cty"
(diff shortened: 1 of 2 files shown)


smaeda-ks (changes requested):
Hi @bendrucker, thanks for the PR.

`tls_subscription_validation` is a virtual resource so I don't think it's worth adding this logic in there. Could you please undo that?

bendrucker:
Hey @smaeda-ks I thought it might still help address the following scenario:

1. I apply configuration that creates a `tls_subscription` and uses the validation resource to wait for completion
2. I delete the TLS subscription outside of Terraform
3. I re-apply the configuration


The result should be the same after steps 1 and 3. Without this logic in the validation resource, I expect the same error would occur on refresh during step 3, requiring `state rm` to proceed.

smaeda-ks:
ok, thanks. I'll give it a try and get back to you.

smaeda-ks (approved):
Added a commit to change the way to return the warning message (and removed nesting).

PR #482, opened 2021-10-05
Add `use_tls` attribute for Splunk logging

smaeda-ks:
ref. https://developer.fastly.com/reference/api/logging/splunk/

^^ the default `use_tls` value is `false` so this addition is not considered as breaking change.

Also bumping go-fastly version to v5.1.0 to support this new struct field:
https://github.com/fastly/go-fastly/blob/main/CHANGELOG.md

### DIFF
--- a/fastly/block_fastly_service_v1_splunk.go
+++ b/fastly/block_fastly_service_v1_splunk.go
@@ -73,7 +73,7 @@ func (h *SplunkServiceAttributeHandler) GetSchema() *schema.Schema {
 			Type:        schema.TypeBool,
 			Optional:    true,
 			Default:     false,
-			Description: "Whether to use TLS for secure logging. Defult: `false`",
+			Description: "Whether to use TLS for secure logging. Default: `false`",
 		},
 	}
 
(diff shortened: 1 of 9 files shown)


fastly/block_fastly_service_v1_splunk.go
```diff
@@ -69,6 +69,12 @@ func (h *SplunkServiceAttributeHandler) GetSchema() *schema.Schema {
 			Description: "The client private key used to make authenticated requests. Must be in PEM format.",
 			Sensitive:   true,
 		},
+		"use_tls": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
+			Description: "Whether to use TLS for secure logging. Defult: `false`",
```
Integralist: @smaeda-ks can you regenerate the docs please once applying this edit...

```suggestion
			Description: "Whether to use TLS for secure logging. Default: `false`",
```

Integralist (changes requested):
LGTM. Just one small typo to correct and thus docs to be regenerated.

fastly/block_fastly_service_v1_splunk.go
smaeda-ks: @Integralist Thank you. Fixed the typo.

Integralist (approved):

Issue #484, opened 2021-10-09
[Tips] Resolve `i/o timeout` error when running Terraform on GCP network


smaeda-ks:
### Issue
A symptom is a user gets `dial tcp x.x.x.x:443: i/o timeout` error message when running Terraform (plan/apply) as below:

```bash
Error: Get "https://api.fastly.com/service/xxxx/details": dial tcp 199.232.78.35:443: i/o timeout

Error: [ERR] Error looking up HTTPS logging endpoints for (xxxx), version (13): Get "https://api.fastly.com/service/xxxx/version/13/logging/https": dial tcp 199.232.78.35:443: i/o timeout
```

### Cause

If you’re running Terraform on GCP (Google Cloud) network, specifically on a VM machine under Cloud NAT, each VM can only use 64 ports by default. Therefore, it’s very possible to cause port exhaustion during Terraform execution as it makes many API requests in parallel.

Is Terraform sending THAT amount of parallel requests...?
Not really. The main reason is that GCP NAT has a strict limitation where they enforce "delay" of two-minuets for port reuse: https://cloud.google.com/nat/docs/ports-and-addresses#ports-reuse-tcp

This is why it is so easy to cause port exhaustion issues and hence the timeout error.

### Workaround

You can increase port, or assign a public IP to the VM machine.
https://cloud.google.com/nat/docs/troubleshooting#cloud-nat-and-ports
https://cloud.google.com/nat/docs/ports-and-addresses#ports

See also: [Best practices around running Terraform Cloud agents behind GCP NAT (SNAT)](https://support.hashicorp.com/hc/en-us/articles/4406210135443-Best-practices-around-running-Terraform-Cloud-agents-behind-GCP-NAT-SNAT-)

smaeda-ks:
### Update

We have identified a problem in our provider where the connection keep-alive was being disabled by the underlying library we use, which was contributing to this particular problem.

The new provider version has been released with the fix:
https://github.com/fastly/terraform-provider-fastly/releases/tag/v0.37.0

We believe this fix would reduce the occurrence of this NAT issue with the GCP network.

PR #485, opened 2021-10-09
Override `Transport` to enable keepalive and add new `force_http2` provider option

smaeda-ks:
This is rather a very micro performance tuning.

The main motivation of this change is to (very slightly) help users in this situation:
https://github.com/fastly/terraform-provider-fastly/issues/484

This PR is not a fix nor a complete workaround for that issue at all and it can still happen, but we can at least reduce port consumption to 1/2 if they use more than one Fastly provider resource within the same plan/apply.

For instance, let's say if I have three Fastly resources defined in my .tf file, such as `fastly_service_v1`, `fastly_tls_subscription `, `fastly_ip_ranges`.

If I run `plan`, then this will consume at least 3 source ports because of the TLS handshake messages initiated by each resource:

Screen Shot 2021-10-10 at 3 12 39

Because each resource initializes its own client (@bengesoff correct me if I'm wrong), by default, `http.Transport` will start TLS handshake regardless of the existing established connection (pool) in order to verify if the destination supports HTTP/2 protocol by checking ALPN fields in the Server Hello message so that it can fallback to HTTP/1.x if needed.

But `http2.Transport` doesn't provide such fallback and always assumes the destination supports HTTP/2 so it can skip TLS handshake if there's a connection pool for that destination already. So in the above example, port consumption would be reduced from 3 to 1.

Since `api.fastly.com` supports HTTP/2, it should be fine to force HTTP/2.

---

[update 1]

For the safeness, we could introduce a new provider option something like `force_http2`:
https://github.com/fastly/terraform-provider-fastly/blob/main/fastly/provider.go#L15
(to be clear, HTTP/2 is enabled by default, but this option would disable a fallback to HTTP/1.x)

but my opinion is that we should still enable it by default and give users an option to opt it out if they see any issues dealing with HTTP/2 due to their local network (e.g., security proxies in the corp-network). Because otherwise, no one would notice this new option, I think.

[update 2]

As @bengesoff pointed out in his comment below, we're actually using [cleanhttp](https://github.com/hashicorp/go-cleanhttp) module via go-fastly.
https://github.com/fastly/go-fastly/blob/v5.1.0/fastly/client.go#L161-L163

This means we are effectively disabling keepalive connection as well as `MaxIdleConnsPerHost`:
https://github.com/hashicorp/go-cleanhttp/blob/v0.5.2/cleanhttp.go#L14-L15

This is consuming local ports (source ports) more than necessary that could impact some of the clients under the restricted NAT environment such as https://github.com/fastly/terraform-provider-fastly/issues/484
and it turned out to be the main cause of this port exhaustion issue.

Fortunately, this PR would fix this problem as well.

[update 3]

Since now we know the core problem is keepalive being disabled by cleanhttp, I'm introducing a new `force_http2` provider option but with the default value of `false`.

### DIFF
--- a/fastly/config.go
+++ b/fastly/config.go
@@ -47,8 +47,8 @@ func (c *Config) Client() (*FastlyClient, diag.Diagnostics) {
 	// 2 (minor). while http.Transport supports HTTP/2 by default, it does TLS-ALPN negotiation
 	// in order to support HTTP/1.x fallback. This means each new client connection initiated
 	// by each resource will start TLS handshake regardless of the existing connection pool status.
-	// explicitly assigning http2.Transport so there will be just one TLS-ALPN negotiation happens
-	// amoung all Fastly provider resources against the same api.fastly.com:443 destination.
+	// explicitly assigning http2.Transport so there will be just one TLS-ALPN negotiation happening
+	// (across all Fastly provider resources) against the same api.fastly.com:443 destination.
 	if c.ForceHttp2 {
 		fastlyClient.HTTPClient.Transport = &http2.Transport{}
 	} else {
--- a/fastly/config.go
+++ b/fastly/config.go
@@ -6,6 +6,7 @@ import (
 	gofastly "github.com/fastly/go-fastly/v5/fastly"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/logging"
+	"golang.org/x/net/http2"
 )
 
 type Config struct {
@@ -33,6 +34,11 @@ func (c *Config) Client() (*FastlyClient, diag.Diagnostics) {
 		return nil, diag.FromErr(err)
 	}
 
+	// explicitly assigning http2.Transport so there will be just one TLS-ALPN negotiation happens
+	// amoung all Fastly provider resources against the same api.fastly.com:443 destination.
+	// otherwise, each resource would consume a different source port due to TLS handshake.
+	// see also: https://github.com/fastly/terraform-provider-fastly/issues/484
+	fastlyClient.HTTPClient.Transport = &http2.Transport{}
 	fastlyClient.HTTPClient.Transport = logging.NewTransport("Fastly", fastlyClient.HTTPClient.Transport)
 
 	client.conn = fastlyClient
--- a/fastly/config.go
+++ b/fastly/config.go
@@ -2,6 +2,7 @@ package fastly
 
 import (
 	"fmt"
+	"net/http"
 
 	gofastly "github.com/fastly/go-fastly/v5/fastly"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -10,10 +11,11 @@ import (
 )
 
 type Config struct {
-	ApiKey    string
-	BaseURL   string
-	UserAgent string
-	NoAuth    bool
+	ApiKey     string
+	BaseURL    string
+	UserAgent  string
+	NoAuth     bool
+	ForceHttp2 bool
 }
 
 type FastlyClient struct {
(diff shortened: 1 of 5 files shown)


bengesoff:
Interesting work here @smaeda-ks! 

I _believe_ the client should be shared among resources with the same provider.

I wonder if multiple ports are being opened due to `go-fastly`'s use of the [`clean-http`](https://github.com/hashicorp/go-cleanhttp) library*? In particular I think it disables `KeepAlive`s  and idle connections which may prevent the connection being reused and therefore having another port opened (as well as disabling connection pooling between instances of the client, but as I say, I don't think this is the case here, although I may be wrong).

I'd be curious whether replacing the HTTP transport and overriding the `cleanhttp`'s transport with **any** other transport using default values would fix the problem, or if it is specifically HTTP/2 that is fixing it here. Have you tried replacing it with the standard transport? Of course the HTTP/2 multiplexing could also be helping if the resources are making API requests in parallel.

Either way, looks like you've found something that works, but I'm thinking if `KeepAlive`s and idle connections were the culprit, then you might be able to get away without forcing HTTP/2, and preserve some safety as you mention in your comment update.

*: https://github.com/fastly/go-fastly/blob/main/fastly/client.go#L162 / https://github.com/hashicorp/go-cleanhttp/blob/master/cleanhttp.go#L44

smaeda-ks:
@bengesoff Ahh, I missed we're using cleanhttp... well, that explains A LOT. Yes, it's indeed contributing to that GCP NAT issue, OMG. In fact, rerunning packet capture with the default code does indicate it's consuming so much source ports due to that `DefaultClient` function disabling keepalive as well as MaxIdleConnsPerHost.

I'm honestly not sure if we want to keep using cleanhttp module any longer... it's unclear why we had to choose it at that time. 

So overring it with `http.Transport` will fix that bit. But it still does not solve my original observation point that is related to how net/http makes a connection in HTTP/2 (has minor impact). It's still relevant to use `http2.Transport` instead to fix both problems altogether.

fastly/config.go
```diff
@@ -33,6 +36,25 @@ func (c *Config) Client() (*FastlyClient, diag.Diagnostics) {
...
+	// this consumes local ports (source ports) more than necessary that could impact
+	// some of the clients under restricted NAT environments such as:
+	// https://github.com/fastly/terraform-provider-fastly/issues/484
+	// overriding it with the default (still non-shared transport) so we can enable keepalive
+	//
+	// 2 (minor). while http.Transport supports HTTP/2 by default, it does TLS-ALPN negotiation
+	// in order to support HTTP/1.x fallback. This means each new client connection initiated
+	// by each resource will start TLS handshake regardless of the existing connection pool status.
+	// explicitly assigning http2.Transport so there will be just one TLS-ALPN negotiation happens
+	// amoung all Fastly provider resources against the same api.fastly.com:443 destination.
```
Integralist:
```suggestion
	// explicitly assigning http2.Transport so there will be just one TLS-ALPN negotiation happening
	// (across all Fastly provider resources) against the same api.fastly.com:443 destination.
```

go.mod
```diff
@@ -11,4 +11,5 @@ require (
 	github.com/hashicorp/terraform-plugin-docs v0.5.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.7.1
 	github.com/stretchr/testify v1.7.0
+	golang.org/x/net v0.0.0-20210326060303-6b1517762897
```
Integralist, line 14: Do we not need an updated `go.sum`?

smaeda-ks: @Integralist I ran mod tidy but that didn't change `go.sum`. This hash appears to be in `go.sum` already

fastly/config_test.go
```diff
@@ -15,3 +16,30 @@ func TestUserAgentContainsProviderVersion(t *testing.T) {
...
+	// <http.Transport Value>
+	ts1 := tv1.FieldByName("transport").Elem().Elem().String()
+
+	tv2 := reflect.ValueOf(client2.conn.HTTPClient.Transport).Elem()
+	// <http2.Transport Value>
+	ts2 := tv2.FieldByName("transport").Elem().Elem().String()
+
+	if ts1 == ts2 {
+		t.Errorf("Failed to create client with force_http2: %#v, %#v", ts1, ts2)
+	}
```
smaeda-ks: I haven't come up with a better way of testing this easily given `http2.Transport` is under `logging.transport`, which is a private struct field.

Integralist (approved):

fastly/config_test.go
```diff
@@ -15,3 +16,30 @@ func TestUserAgentContainsProviderVersion(t *testing.T) {
...
+	c2 := Config{
+		ApiKey:     "someapikey",
+		BaseURL:    "http://localhost",
+		ForceHttp2: true,
+	}
+	client2, _ := c2.Client()
+
+	tv1 := reflect.ValueOf(client1.conn.HTTPClient.Transport).Elem()
+	// <http.Transport Value>
+	ts1 := tv1.FieldByName("transport").Elem().Elem().String()
```
Integralist: @smaeda-ks Why the double `.Elem()` chain?

smaeda-ks: @Integralist No strong reason to be honest... the returned string would be `<*http.Transport Value>` (.Elem().String()) or `<http.Transport Value>` (.Elem().Elem().String()), either way works. Do you prefer the former?

PR #487, opened 2021-10-12
Ignore 404 on GetPackage when importing wasm service

smaeda-ks:


smaeda-ks:
@bengesoff This is a low priority, but would like to have your thoughts on this.
Basically, if a user creates a wasm service in the Fastly UI with no package added (e.g., has no active version yet) and try to import this service, `GetPackage` will report 404 and results in error.

fastly/block_fastly_service_v1_package.go
```diff
@@ -76,12 +76,16 @@ func (h *PackageServiceAttributeHandler) Read(ctx context.Context, d *schema.Res
 		ServiceVersion: s.ActiveVersion.Number,
 	})
 
-	if err != nil {
+	isNewImport := false
+	if err, ok := err.(*gofastly.HTTPError); ok && err.IsNotFound() {
+		isNewImport = true
+		log.Printf("[WARN] No wasm Package found for (%s), version (%v): %v", d.Id(), s.ActiveVersion.Number, err)
```
bengesoff: Wouldn't you want to do something like `d.Set(h.GetKey(), nil)` and then `return nil` here, instead of proceeding? I haven't tried it but I imagine the diff would become
```
+ package {
    ...
  }
```

rather than

```
~ package {
  ~ source_code_hash = "" -> "..."
    filename = (unchanged)
  }
```

which to me shows more clearly that there was no package beforehand.

I agree though, we should definitely handle the 404 case better!

smaeda-ks: Ah, that's right. Thanks, I followed that

bengesoff (approved):
Assuming the latest change works as expected, LGTM 👍🏼

fastly/block_fastly_service_v1_package.go
```diff
@@ -76,7 +76,11 @@ func (h *PackageServiceAttributeHandler) Read(ctx context.Context, d *schema.Res
...
 	})
 
-	if err != nil {
+	if err, ok := err.(*gofastly.HTTPError); ok && err.IsNotFound() {
+		log.Printf("[WARN] No wasm Package found for (%s), version (%v): %v", d.Id(), s.ActiveVersion.Number, err)
+		d.Set(h.GetKey(), nil)
+		return nil
+	} else if err != nil {
 		return fmt.Errorf("[ERR] Error looking up Package for (%s), version (%v): %v", d.Id(), s.ActiveVersion.Number, err)
 	}
```
Integralist: Wondering if we could refactor out the `else if` like so...

```suggestion
        if err != nil {
                if err, ok := err.(*gofastly.HTTPError); ok && err.IsNotFound() {
		        log.Printf("[WARN] No wasm Package found for (%s), version (%v): %v", d.Id(), s.ActiveVersion.Number, err)
		        d.Set(h.GetKey(), nil)
				        return nil
                }
                return fmt.Errorf("[ERR] Error looking up Package for (%s), version (%v): %v", d.Id(), s.ActiveVersion.Number, err)
        }
```

PR #492, opened 2021-10-21
Convert index.md to template to inject provider version

smaeda-ks:


.github/workflows/pr.yml
```diff
@@ -48,6 +48,8 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
smaeda-ks, line 52: because `actions/checkout@v2` does shallow clone and can't find the correct tag name:
https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches

Integralist (approved):

PR #496, opened 2021-11-03
Do not send 0 `subnet` value unless explicitly set

smaeda-ks:
After we changed the `subnet` type to `int`:
https://github.com/fastly/go-fastly/pull/288
and updated go-fastly version:
https://github.com/fastly/terraform-provider-fastly/blob/main/CHANGELOG.md#0350-september-15-2021

We started sending `0` subnet value even if the value is not set. This is because `convertSubnetToInt` returns the default value (`0`) when passing an empty string (`""`). Fixing this to only send 0 value when the `subnet` attribute is explicitly set by the user.

Also, `flattenAclEntries` function needs to know if the returned `subnet` value is `null` or any `int` value in order to track them properly in the state file. To do so, we need to make a change in the upstream go-fastly package:
https://github.com/fastly/go-fastly/pull/316

Otherwise, we always see `0` subnet value regardless of the actual returned value (i.e., `null` vs `0` int value) and save it to the state file, which causes a diff loop.


fastly/resource_fastly_service_acl_entries_v1.go
```diff
@@ -314,6 +302,24 @@ func executeBatchACLOperations(conn *gofastly.Client, serviceID, aclID string, b
...
+		Operation: op,
+		ID:        gofastly.String(v["id"].(string)),
+		IP:        gofastly.String(v["ip"].(string)),
+		Negated:   gofastly.CBool(v["negated"].(bool)),
+		Comment:   gofastly.String(v["comment"].(string)),
+	}
+
+	subnet := convertSubnetToInt(v["subnet"].(string))
+	// only set zero subnet if the attribute is explicitly set
+	if v["subnet"].(string) == "0" || subnet != 0 {
```
bengesoff, line 316: Do you think it might be cleaner to move this logic inside `convertSubnetToInt`? Then `convertSubnetToInt` could return `*int`

bengesoff:
Looks like some of the implementation needs fixing but the principle seems reasonable!

fastly/resource_fastly_service_acl_entries_v1.go
smaeda-ks: I think I'm okay with the current

smaeda-ks:
@bengesoff @Integralist 
I've added one last commit to bump the go-fastly version to `v5.1.2`, which is necessary for this PR.

Integralist (approved):
LGTM

PR #501, opened 2021-11-08
allow TF_LOG=debug in the non-deterministic test

Hiieu:
In this PR, we've set TF_LOG=debug only for estAccFastlyServiceV1_s3logging_basic. 
The aim is to have more context on the error that is probably caused by a race condition.
The logs coming from that test will be visible in the acceptance tests suite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ build:
 test:
 	go test -i $(TEST) || exit 1
 	echo $(TEST) | \
-		xargs -t -n4 go test $(TESTARGS) -timeout=30s -parallel=$(TEST_PARALLELISM)
+		xargs -t -n4 go test -v $(TESTARGS) -timeout=90s -parallel=$(TEST_PARALLELISM)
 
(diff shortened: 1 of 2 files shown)


Integralist (commented):
LGTM. One concern, will this log out a Fastly API key? If so can it be redacted?

Hiieu:
Your concern is valid. It looks like the API key is displayed in the logs.  Let me check if we can redact it from the logs, but after a brief look this would require changing the `Transport` [mechanism](https://github.com/fastly/terraform-provider-fastly/blob/main/fastly/config.go#L72).

smaeda-ks:
If this is a one-time debug thing, not sure if we want to merge this change. Sounds like this should be done locally while we debug this issue, then be removed. Not necessarily to merge this to upstream, IMHO.

Integralist:
@smaeda-ks OK, agreed. We should just get used to adding `TF_LOG=debug` _every time_ we run the test suite so that if the non-deterministic S3 'IAM role can't be used with access/secret key' error is triggered then hopefully we'll have more contextual data available to debug what happened.

PR #506, opened 2021-11-10
Expose `certificate_id` read-only attribute from `fastly_tls_subscription` resource

smaeda-ks:
Use case:
https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/tls_activation

`fastly_tls_certificate` resource isn't used for Fastly managed certs, but BYO certs. Therefore, if adding TLS configuration for a given `certificate_id` that is a Fastly managed cert, there isn't a way to obtain the required `certificate_id` for that `fastly_tls_activation` resource unless exposing it via the `fastly_tls_subscription` resource or specifying the id manually.

### DIFF
--- a/fastly/resource_fastly_tls_subscription.go
+++ b/fastly/resource_fastly_tls_subscription.go
@@ -60,6 +60,11 @@ func resourceFastlyTLSSubscription() *schema.Resource {
 				Optional:    true,
 				Computed:    true,
 			},
+			"certificate_id": {
+				Type:        schema.TypeString,
+				Description: "The certificate ID associated with the subscription.",
+				Computed:    true,
+			},
 			"created_at": {
 				Type:        schema.TypeString,
 				Description: "Timestamp (GMT) when the subscription was created.",
@@ -213,6 +218,9 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 		domains = append(domains, domain.ID)
 	}
 
+	// NOTE: there must be only one certificate id included per subscription
+	var certificateId = subscription.Certificates[len(subscription.Certificates)-1].ID
+
 	var managedHTTPChallenges []map[string]interface{}
 	var managedDNSChallenges []map[string]interface{}
 	for _, domain := range subscription.Authorizations {
@@ -269,6 +277,10 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 		return diag.FromErr(err)
 	}
 	err = d.Set("common_name", subscription.CommonName.ID)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	err = d.Set("certificate_id", certificateId)
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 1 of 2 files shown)


smaeda-ks:
cc: @bengesoff

fastly/resource_fastly_tls_subscription.go
```diff
@@ -213,6 +218,9 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 		domains = append(domains, domain.ID)
 	}
 
+	// NOTE: there must be only one certificate id included per subscription
+	var certificateId = subscription.Certificates[0].ID
```
Hiieu: `TestAccResourceFastlyTLSSubscription` is failing on this line with this error:
```
panic: runtime error: index out of range [0] with length 0
```
Should we generate a certificate for that test case?

smaeda-ks: Thanks @Hiieu 
No, I think I need to add some conditions there. Will test and get back to you.

smaeda-ks: @Hiieu pushed another commit.

Hiieu: Nice, looks good to me 👍

Hiieu (approved):

fastly/resource_fastly_tls_subscription.go
```diff
@@ -213,6 +218,13 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 		domains = append(domains, domain.ID)
 	}
 
+	// NOTE: there must be only one certificate id included per subscription
+	// "pending" and "processing" state may not include the id (for new subsctiptions)
```
Integralist:
```suggestion
	// "pending" and "processing" state may not include the id (for new subscriptions)
```

fastly/resource_fastly_tls_subscription.go
```diff
@@ -213,6 +218,13 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 		domains = append(domains, domain.ID)
 	}
 
+	// NOTE: there must be only one certificate id included per subscription
+	// "pending" and "processing" state may not include the id (for new subscriptions)
+	var certificateId = ""
+	if len(subscription.Certificates) > 0 {
+		certificateId = subscription.Certificates[0].ID
+	}
```
Integralist, line 226: What happens if the len is zero (is that possible)? If so I presume we need to stop the logic flow.

Integralist (commented):
LGTM with the exception of one question...

fastly/resource_fastly_tls_subscription.go
smaeda-ks: @Integralist As the comment suggests, it is possible we get empty `data[]` in response. But we don't need to stop the logic, we just pass it through and store `""` as the cert id with those cert state ("pending"/"processing"). We will get the cert id when the cert is actually issued (e.g., in the next TF refresh).

bengesoff (approved):

Integralist (approved):

Issue #517, opened 2021-12-16
Cannot change condition type


cori:
When applying a change to a vcl service with a conditioned header and trying to change the condition and header type the apply fails with a Bad Request: Cannot change type message. I _believe_ this is when changing the Condition type, which also can't be done in the UI.

The below example fails when trying to change the condition type from `CACHE` to `REQUEST` and the header type from `cache` to `request`.

### Terraform Version
Terraform v1.1.1
on darwin_amd64
+ provider registry.terraform.io/fastly/fastly v0.34.0
+ provider registry.terraform.io/hashicorp/aws v3.65.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/template v2.1.2

### Affected Resource(s)
Please list the resources as a list, for example:
- fastly_service_v1 (specifically, I believe, the nested condition record)

### Terraform Configuration Files
main module file

```hcl
terraform {
  required_version = ">= 0.14.3"
  required_providers {
    fastly = {
      source  = "fastly/fastly"
      version = "~> 0.34.0"
    }
  }
}

provider "fastly" {
  api_key = var.api_key
}

resource "fastly_service_v1" "service" {
  name = var.name

  domain {
    name    = var.domain_name
    comment = "${var.name}-domain"
  }

  # currently we only support a single backend per service
  backend {
    address           = var.backend_hostname
    name              = "${var.name}-backend"
    shield            = "iad-va-us"
    ssl_check_cert    = true
    use_ssl           = true
    auto_loadbalance  = false
    ssl_cert_hostname = var.use_service_cert ? var.backend_hostname : var.domain_name
    ssl_sni_hostname  = var.use_service_cert ? var.backend_hostname : var.domain_name
    override_host     = var.use_service_cert ? var.backend_hostname : null
    port              = 443

  }

  dynamic "condition" {
    for_each = var.use_image_optimization ? [1] : []
    content {
      type      = "REQUEST"
      name      = "${var.name}-image-optimization-condition"
      statement = "req.*"
    }
  }

  dynamic "header" {
    for_each = var.use_image_optimization ? [1] : []
    content {
      action          = "set"
      destination     = "http.x-fastly-imageopto-api"
      type            = "request"
      source          = "\"fastly; qp=*\""
      name            = "${var.name}-image-optimization-header"
      cache_condition = "${var.name}-image-optimization-condition"
      priority        = 1 # image optimization headers should be first
    }
  }

[...]

PR #518, opened 2021-12-17
force creating new `condition` if `type` is changed

smaeda-ks:
https://github.com/fastly/terraform-provider-fastly/issues/517

background: Fastly API doens't support updating the `type` field once it's created.

And because `fastly_service_v1` is a mono resource, the usual `ForceNew` approach doesn't work as it will attempt to re-create a `fastly_service_v1` resource as a whole. `CustomizeDiff` is also not an option either as `condition` is a nested block.

So we unfortunately implement it in this way, where we silently call DELETE and CREATE if `type` attribute is changed. Otherwise, call UPDATE.

### DIFF
--- a/fastly/block_fastly_service_v1_condition.go
+++ b/fastly/block_fastly_service_v1_condition.go
@@ -100,7 +100,16 @@ func (h *ConditionServiceAttributeHandler) Read(_ context.Context, d *schema.Res
 
 func (h *ConditionServiceAttributeHandler) Update(_ context.Context, d *schema.ResourceData, resource, modified map[string]interface {
 }, serviceVersion int, conn *gofastly.Client) error {
-	opts := gofastly.UpdateConditionInput{
+	optsCreate := gofastly.CreateConditionInput{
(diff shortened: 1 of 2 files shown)


Integralist (approved):
Thanks @smaeda-ks 

The implementation makes sense to me. 

Of course it would be preferable for the API to support this feature rather than us implementing this in the Terraform provider (as has been mentioned by others this can lead to fragmentation across API clients). 

That said, this feature is currently broken for users. Due to the potential time it would take to get the API to support this, I'm OK seeing this PR merged/released, with the intention that we actively push for the API to be updated and then we can come back and update this logic accordingly.

**Question**: Are there any downsides/side-effects that I'm not considering that would make us what to tread cautiously before merging?

smaeda-ks:
@Integralist 

> Question: Are there any downsides/side-effects that I'm not considering that would make us what to tread cautiously before merging?

No, I don't expect any side-effects by this change. Thanks

Integralist (approved):
LGTM. Thanks for adding the test.

PR #525, opened 2022-01-13
Bump go-fastly to v6.0.0

Integralist:


Integralist:
I've run the complete test case on this code and it's passing.

smaeda-ks (approved):

PR #516, opened 2021-12-15
simplify example in `fastly_tls_subscription` doc

smaeda-ks:
This also addresses some feedback in https://github.com/fastly/terraform-provider-fastly/issues/513

PR #527, opened 2022-01-15
Revamp `fastly_tls_subscription` examples

smaeda-ks:
continuing my previous PR: https://github.com/fastly/terraform-provider-fastly/pull/516

Adding a large note in the example as it's doing fairly complex things.
We updated the example with the simplified version in the previous PR but it might not be the best option due to the limitation I left in the comment in this PR.

### DIFF
--- a/docs/resources/tls_subscription.md
+++ b/docs/resources/tls_subscription.md
@@ -48,14 +48,21 @@ Usage with AWS Route 53:
 
 ```terraform
 locals {
-  domain_name = "example.com"
+  domains = [
+    "example.com",
+    "*.example.com",
+  ]
+  aws_route53_zone_id = "your_route53_zone_id"
 }
 
 resource "fastly_service_v1" "example" {
   name = "example-service"
 
-  domain {
-    name = local.domain_name
+  dynamic domain {
+    for_each = local.domains
+    content {
+      name = domain.value
+    }
   }
 
   backend {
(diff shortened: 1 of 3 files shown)


docs/resources/tls_subscription.md
```diff
@@ -71,24 +78,42 @@ resource "fastly_tls_subscription" "example" {
...
   depends_on = [fastly_tls_subscription.example]
 
-  for_each        = {
-    for challenge in fastly_tls_subscription.example.managed_dns_challenges :
-    trimprefix(challenge.record_name, "_acme-challenge.") => challenge
+  # NOTE: in this example, two domains are added to the cert ("example.com" and "*.example.com").
+  # The "managed_dns_challenges" read-only attribute only includes one object
+  # for "_acme-challenge.example.com" as the challenge record is common in these two domains.
+  # 
+  # In order to process a cert contains wildcard entries, remove wildcard prefix "*." from the key
```
Integralist:
```suggestion
  # In order to process a cert containing wildcard entries, remove wildcard prefix "*." from the key
```

docs/resources/tls_subscription.md
```diff
@@ -71,24 +78,42 @@ resource "fastly_tls_subscription" "example" {
...
+  # 
+  # A simplified version of this complex "for_each" usage would be:
+  # ```
+  # for_each = {
+  #   for challenge in fastly_tls_subscription.example.managed_dns_challenges :
+  #   trimprefix(challenge.record_name, "_acme-challenge.") => challenge
+  # }
+  # ```
+  # but since the "managed_dns_challenges" attribute is only known after apply,
+  # you will need to create this resource separately ("--target" option) and may not be ideal.
```
Integralist: I know that in golang, defined flags will work with double or single hyphens (e.g. `--target` and `-target`) but in [HashiCorp's documentation](https://learn.hashicorp.com/tutorials/terraform/resource-targeting?in=terraform/cli) they use a single dash when describing flags (such as `-target`) and so I wonder if we should be consistent with them here.
```suggestion
  # you will need to create this resource separately ("-target" option) and may not be ideal.
```

Integralist: Is it worth elaborating on the workflow the user would need to take? Unless it's just a case of the user running `terraform plan` to identify the changes and then `terraform apply -target="aws_route53_record.domain_validation"` and then `terraform apply` again if they have other things they want to be applied?

examples/resources/tls_subscription_with_route53.tf
```diff
@@ -22,24 +29,42 @@ resource "fastly_tls_subscription" "example" {
...
   depends_on = [fastly_tls_subscription.example]
 
-  for_each        = {
-    for challenge in fastly_tls_subscription.example.managed_dns_challenges :
-    trimprefix(challenge.record_name, "_acme-challenge.") => challenge
+  # NOTE: in this example, two domains are added to the cert ("example.com" and "*.example.com").
+  # The "managed_dns_challenges" read-only attribute only includes one object
+  # for "_acme-challenge.example.com" as the challenge record is common in these two domains.
+  # 
+  # In order to process a cert contains wildcard entries, remove wildcard prefix "*." from the key
```
Integralist: Same suggestions from above here.

Integralist (approved):
LGTM. Just a couple of minor comments.

docs/resources/tls_subscription.md
smaeda-ks: @Integralist Thanks. I would leave it as is since the exact workflow may be varied depending upon users' config structure, strategy, etc. I think the current note suggests enough as to what they need if they really want to go with that route (also, TF error messages tell the same information nicely anyway).

PR #528, opened 2022-01-17
Support updating TLS subscriptions in pending state #524

### LINKED ISSUE #524: Support updating TLS certs in pending state
With some recent updates in Fastly TLS stack, we apparently allow users to update certs in a pending state. 

We need some research to see if we should support that path:
https://github.com/fastly/terraform-provider-fastly/blob/v0.35.0/fastly/resource_fastly_tls_subscription.go#L25-L33

smaeda-ks:
#524

### DIFF
--- a/fastly/resource_fastly_tls_subscription_test.go
+++ b/fastly/resource_fastly_tls_subscription_test.go
@@ -26,6 +26,7 @@ func TestAccResourceFastlyTLSSubscription(t *testing.T) {
 	domain2_bad := fmt.Sprintf("%sALT.test", name)
 	commonName1 := domain1
 	commonName2 := domain2
+	var subscriptionId string
 
 	resourceName := "fastly_tls_subscription.subject"
 	resource.ParallelTest(t, resource.TestCase{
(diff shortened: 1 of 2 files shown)


fastly/resource_fastly_tls_subscription_test.go
```diff
@@ -43,12 +44,16 @@ func TestAccResourceFastlyTLSSubscription(t *testing.T) {
...
+					testAccResourceFastlyTLSSubscriptionExists(resourceName, &subscriptionId),
 				),
 			},
 			{
 				Config: testAccResourceFastlyTLSSubscriptionConfig(name, domain1, domain2, commonName2),
-				Check:  resource.TestCheckResourceAttr(resourceName, "common_name", domain2),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "common_name", domain2),
+					// subscription is "updated" so the subscription ID should remain the same
+					resource.TestCheckResourceAttrPtr(resourceName, "id", &subscriptionId),
```
smaeda-ks, line 55: ref. https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/resource#TestCheckResourceAttrPtr
We can't use `TestCheckResourceAttr` here because then it's always an empty string (default value of `string`) and not reflecting the changes made in the previous step.

Integralist (approved):

PR #535, opened 2022-01-27
Remove Director `capacity` field as it's unused

Integralist:
https://developer.fastly.com/reference/api/load-balancing/directors/director/

Screenshot 2022-01-27 at 17 45 13

I've run the full test suite and it's passing.


Integralist:
@triblondon this is ready to review/approve as I've run the full test suite and it's passing.

phamann (approved):

PR #474, opened 2021-09-22
Update dependencies

smaeda-ks:
the main diff is: https://github.com/fastly/terraform-provider-fastly/commit/8c38a0b91d52dcee5c1d624784adcec134a76bc5
others are just a result of running go mod vendor

This is to successfully run test suites with Terraform v1 (`terraform-json` version needs to be updated):
https://github.com/hashicorp/terraform-json/commit/7b6375ccc97200ac852431c01fedd0d8a9045aed#diff-11ffedf6a81044546aa19c8009998208f5a7afec3cdff04ca4281117b5ec0da7R11

Full tests passed on both Terraform v1.0.7 and v0.15.

PR #536, opened 2022-01-27
Support Terraform 1.x

Integralist:
I updated my terraform version from 0.15 to 1.1.4 and noticed an error when running the tests...

```
error: unsupported state format version: expected ["0.1" "0.2"], got "1.0"
```

The fix that was suggested from the terratest project was to bump the `terraform-json` dependency (and worked).


Integralist:
I ran the tests using both 0.x and 1.x terraform version and they both now pass.

kailan (approved):

smaeda-ks:
@Integralist Was this really required? I think we fixed this before:https://github.com/fastly/terraform-provider-fastly/pull/474
and I had no issues with v1 binary since then. I might think this was your local env specific problem.

smaeda-ks:
Ah, never mind. I see it now:
https://github.com/hashicorp/terraform-json/commit/1b370c6345872777de33f05853a25e94cebe1f3d

smaeda-ks:
Looking at the changelogs, it probably makes more sense (and safe) to bump the SDK version rather than adding the indirect reference.
https://github.com/hashicorp/terraform-plugin-sdk/blob/main/CHANGELOG.md

```
	github.com/hashicorp/terraform-plugin-sdk/v2 v2.7.1
```

to 

```
	github.com/hashicorp/terraform-plugin-sdk/v2 v2.10.1
```

which should also fix this.

ref. https://github.com/hashicorp/terraform-plugin-sdk/commit/a928a7b22790d6946de295c1e5cd884be2cf1fc7

Integralist:
Cool. I'll look at bumping the sdk version 👍🏻

PR #537, opened 2022-01-28
Add missing docs and attributes in versionless resources

Hiieu:
Changes:
- Add missing `manage_x` attributes in the docs and examples
- Add `manage_snippets` attribute in the `fastly_service_dynamic_snippet_content` resource.


docs/resources/service_acl_entries.md
```diff
@@ -199,6 +199,33 @@ resource "fastly_service_acl_entries" "entries" {
 }
 ```
 
+### Reapplying original entries with `managed_entries` if the state of the entries drifts
```
Integralist, line 174: @Hiieu might be worth checking with @bengesoff but would the above section "Supporting API and UI ACL updates with ignore_changes" not need to be removed? I don't think using `ignore_changes` would work since Ben's new `manage_*` implementation (or at least I'm not sure if we want to avoid publicly promoting that pattern as it might have unexpected side effects maybe)?

Integralist: Same comment essentially for `docs/resources/service_dictionary_items.md`.

examples/resources/service_dynamic_snippet_content_manage_snippets.tf
```diff
@@ -0,0 +1,15 @@
...
+resource "fastly_service_dynamic_snippet_content" "my_dyn_content" {
+  for_each = {
+    for d in fastly_service_vcl.myservice.dynamicsnippet : d.name => d if d.name == "My Dynamic Snippet"
+  }
+  service_id       = fastly_service_vcl.myservice.id
+  snippet_id       = each.value.snippet_id
+  manage_snippets = true
+  content          = "if ( req.url ) {\n set req.http.my-snippet-test-header = \"true\";\n}"
+
+  lifecycle {
```
Integralist: As mentioned above I don't know if `ignore_changes` is expected to work when the new `manifest_*` is in place?

docs/resources/service_acl_entries.md
Hiieu: As discussed, we've removed `ignore_changes` from the docs and examples, also added a note saying that the built-in meta-argument takes precedence over `manage_x` attributes.

Issue #523, opened 2022-01-06
Implement missing Data Sources


Integralist:
I've noticed on a few occasions that within a top-level API section there will be API endpoints we've implemented but omitted the 'list' behaviour. 

We should investigate what's missing across all the implemented API endpoints and add those list behaviours as new data sources. 

> **NOTE**: I think we just need to double check whether the missing verbs should continue to be omitted when it doesn't make sense to display the sort of information they provide (e.g. does it make sense to implement the stats API endpoints as data sources).

PR #540, opened 2022-02-09
Add datacenters data resource

smaeda-ks:
maybe related: https://github.com/fastly/terraform-provider-fastly/issues/523

```sh
$ FASTLY_API_KEY=**** make testacc TESTARGS='-run=TestAccFastlyDataSourceDatacenters'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccFastlyDataSourceDatacenters -parallel=4 -timeout 360m -ldflags="-X=github.com/fastly/terraform-provider-fastly/version.ProviderVersion=acc"
=== RUN   TestAccFastlyDataSourceDatacenters
=== PAUSE TestAccFastlyDataSourceDatacenters
=== CONT  TestAccFastlyDataSourceDatacenters
--- PASS: TestAccFastlyDataSourceDatacenters (1.34s)
PASS
ok  	github.com/fastly/terraform-provider-fastly/fastly	2.193s
?   	github.com/fastly/terraform-provider-fastly/fastly/hashcode	[no test files]
```


examples/data-sources/datacenters.tf
```diff
@@ -0,0 +1,11 @@
+
+data "fastly_datacenters" "fastly" {}
+
+output "fastly_datacenters_all" {
+  value = data.fastly_datacenters.fastly
```
Integralist, line 5: Should this be... ?

```suggestion
  value = data.fastly_datacenters.fastly.pops
```

fastly/data_source_datacenters.go
```diff
@@ -0,0 +1,101 @@
...
+						"group": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "A code representing the general region of the world in which the POP location resides.",
+						},
+						"shield": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "A code representing the shielding name of the POP. The value may be empty if the POP is not available for shielding.",
+						},
```
Integralist, line 45: I appreciate `coordinates` probably isn't that useful in practice, but for completeness sake is it worth including? 🤔 I'm leaning against _not_ including it, but thought I'd just ask quickly out of interest.

fastly/data_source_datacenters.go
```diff
@@ -0,0 +1,101 @@
...
+
+	log.Printf("[DEBUG] Reading datacenters")
+
+	datacenters, err := conn.AllDatacenters()
+
+	if err != nil {
+		return diag.Errorf("error fetching datacenters: %s", err)
+	}
+	hashBase, _ := json.Marshal(datacenters)
+	hashString := strconv.Itoa(hashcode.String(string(hashBase)))
```
Integralist, line 65: The pattern across the code base so far has been to use `fmt.Sprintf("%d", hashcode.String(...))` although that feels a bit ...hacky (for reasons I can't articulate) compared to using `strconv.Itoa`.

Seems others have written about the performance implications of using `fmt.Sprintf` (https://medium.com/swlh/bad-go-frivolous-sprintf-2ad28fedf1a0) and so it's probably worth at some point us updating the code to use `strconv.Itoa`.

fastly/data_source_datacenters.go
```diff
@@ -0,0 +1,101 @@
...
+		return diag.Errorf("error setting datacenters: %s", err)
+	}
+
+	return nil
+
+}
+
+func flattenDatacenters(datacenters []gofastly.Datacenter) []map[string]interface{} {
+
+	pops := make([]map[string]interface{}, len(datacenters))
```
Integralist, line 78: If there is no specific need for a slice, then should we maybe create a sized array instead?

```suggestion
	var pops [len(datacenters)]map[string]interface{}
```

Integralist (commented):
Thanks @smaeda-ks for opening this PR. I had some questions, but otherwise this LGTM.

fastly/data_source_datacenters.go
Integralist: Hmm, maybe not necessary. 

I was wondering about memory allocations when re-sizing the backing array, but I think that doesn't happen when assigning directly using an index (e.g. only when using `append`, at least from some quick testing it doesn't look like using a slice in this example would cause the backing array to be resized).

fastly/data_source_datacenters.go
smaeda-ks: This was my personal decision to not include it as I can't really see the exact use case for that, especially when it comes to the IaC. For now, I don't think we would need to include it. We can always come back to this later if someone asks, and it won't be a breaking change.

examples/data-sources/datacenters.tf
smaeda-ks: in order to represent the whole structure, I didn't include `.pop` and I think that's right.

PR #542, opened 2022-02-15
docs: fix fastly-s3 hyperlinks

chenrui333:
Fix the hyperlinks in the docs page, below is what is looks like right now.
image


Integralist (approved):
Thanks for catching these issues!

chenrui333:
np 👍

PR #534, opened 2022-01-27
Changes for v1.0.0

Integralist:
**This PR consists of the following changes...**

- Consistent resource naming
- Make backend auto_loadbalance default to false
- Use TypeList schema for gzip nested block attributes
- V1 change email to user in gcslogging
- V1 update default format and format version
- V1 versionless deletion protection

To see the individual changes refer to https://github.com/fastly/terraform-provider-fastly/pull/532 which has the changes broken down by commit (i.e. 6 changes, 6 commits).

**NOTE**: All commits were pre-approved PRs from OpenCredo, Shohei and myself. Meaning that once this PR has a confirmed successful test suite (I'll post in comments below once that is the case), then it should be safe to merge and to cut a new 1.0.0 release.

**ADDITIONAL:** 

- ✅ Support Terraform 1.x CLI ([PR](https://github.com/fastly/terraform-provider-fastly/pull/536)).
- ✅ Remove `capacity` from Dictionary resource ([PR](https://github.com/fastly/terraform-provider-fastly/pull/535)).
- ✅ Write a migration guide ([here is the guide](https://gist.github.com/Integralist/b4c3231f0c95d59cf977c2911125a0a4)).

PR #549, opened 2022-02-25
Publish 1.0.0 Migration Guide as part of docs

reidab:
Starting from https://registry.terraform.io/providers/fastly/fastly/latest/docs and trying to find upgrade information, I had to jump to this repo, find the the changelog, follow the link from there to #534, and then find the link there to the migration guide. I'd like to help make that a bit more obvious for folks in the future. 😀

This imports the 1.0.0 migration guide from https://gist.github.com/Integralist/b4c3231f0c95d59cf977c2911125a0a4 into `/docs/guides` for publication on the Terraform registry with the rest of the docs.

(I'm hoping that this is the most up-to-date version of the migration guide, but the fact that the gist is tagged with `#delete` has me questioning that…)

### DIFF
--- a/docs/guides/1.0.0-migration-guide.md
+++ b/docs/guides/1.0.0-migration-guide.md
@@ -1,0 +1,58 @@
+---
+page_title: "Fastly Terraform 1.0.0 Migration Guide"
+---
+
+## Fastly Terraform 1.0.0 Migration Guide
+
+**Resources omit `_v1` suffix**:
+
+All resources now have a consistent naming convention which omit the previous v1 suffix.
+
+- `fastly_service_v1` -> `fastly_service_vcl` (renamed with `_vcl` suffix)
+- `fastly_service_acl_entries_v1` -> `fastly_service_acl_entries`
+- `fastly_service_dictionary_items_v1` -> `fastly_service_dictionary_items`
+- `fastly_service_dynamic_snippet_content_v1` -> `fastly_service_dynamic_snippet_content`
+- `fastly_user_v1` -> `fastly_user`
+
+**Logging resources have consistent naming format**:
+
+All logging resources now have a consistent naming convention with the provider prefixed with `logging_`.
+
+- `bigquerylogging` -> `logging_bigquery`
+- `blobstoragelogging` -> `logging_blobstorage`
+- `gcslogging` -> `logging_gcs`
+- `httpslogging` -> `logging_https`
+- `logentries` -> `logging_logentries`
+- `papertrail` -> `logging_papertrail`
+- `s3logging` -> `logging_s3`
+- `splunk` -> `logging_splunk`
+- `sumologic` -> `logging_sumologic`
+- `syslog` -> `logging_syslog`
+
+**Director `capacity` removed**:
+
+The Fastly API never supported the `capacity` field for a `director` resource (this was added to the Terraform provider by mistake). Load balancing of director backends is managed by the `weight` field on each associated `backend` resource.
+
+**GCS Logging field `email` renamed**:
+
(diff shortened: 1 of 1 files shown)


Integralist:
Hi @reidab thanks for opening this PR.

I'm sorry you had a rough time locating this information. I actually had this information added to the 'release':

https://github.com/fastly/terraform-provider-fastly/releases/tag/v1.0.0

I'll double check today the changes you have here against the release notes to be sure there's no minor changes I made at the last minute 🙂 

Thanks again.

Integralist:
👋🏻 @reidab be sure to run `make generate-docs` and push up the changes as the associated CI job is currently failing.

Thanks.

docs/guides/1.0.0-migration-guide.md
```diff
@@ -0,0 +1,58 @@
+---
+page_title: "Fastly Terraform 1.0.0 Migration Guide"
```
Integralist, line 2: I think we should reverse the name of the file to be:

```
migration-guide-1.0.0.md
```

Also, as per the [HashiCorp docs](https://www.terraform.io/registry/providers/docs#guides-subcategories), I think we should add a `subcategory` to the frontmatter like so:

```suggestion
page_title: "Fastly Terraform 1.0.0 Migration Guide"
subcategory: "Migration"
```

Then it should get rendered as:

```
Guides
    Migration
        Fastly Terraform 1.0.0 Migration Guide
```

I would prefer this structure as it means, in the future, if we have other guides that need to be added, then we can more clearly distinguish them.

docs/guides/1.0.0-migration-guide.md
```diff
@@ -0,0 +1,58 @@
...
+**Resources omit `_v1` suffix**:
+
+All resources now have a consistent naming convention which omit the previous v1 suffix.
+
+- `fastly_service_v1` -> `fastly_service_vcl` (renamed with `_vcl` suffix)
+- `fastly_service_acl_entries_v1` -> `fastly_service_acl_entries`
+- `fastly_service_dictionary_items_v1` -> `fastly_service_dictionary_items`
+- `fastly_service_dynamic_snippet_content_v1` -> `fastly_service_dynamic_snippet_content`
+- `fastly_user_v1` -> `fastly_user`
+
```
Integralist, line 16: **Can you add the following (before the "Logging resources have consistent naming format:" section). This was a later addition to the [release notes migration guide](https://github.com/fastly/terraform-provider-fastly/releases/tag/v1.0.0)...**

When switching to the new 1.0.0 release, some existing projects might see an error such as:

```
Error: no schema available for module.fastly.<...> while reading state; this is a bug in Terraform and should be reported
```

This error is because the Terraform state file will have a different resource name to what's now exposed via the 1.0.0 release. To resolve this issue you'll need to first remove the old resource from the state file and then reimport the data using the new resource name.

As an example, if you had defined an `fastly_service_acl_entries_v1` resource, then you could remove and reimport using the following commands (any brackets `<...>` should be replaced with actual values):

```bash
terraform state rm 'fastly_service_acl_entries_v1.<your_resource_name>["<your_acl_name>"]'
terraform import 'fastly_service_acl_entries.<your_resource_name>["<your_acl_name>"]' <service_id>/<acl_id>

terraform state rm 'fastly_service_v1.<your_resource_name>'
terraform import 'fastly_service_vcl.<your_resource_name>' <service_id>
```

It's also recommended that you keep a backup of your state file so you can use it to reference any resource IDs necessary.

Integralist (changes requested):

Integralist:
Closing in favour of https://github.com/fastly/terraform-provider-fastly/pull/551 (with contribution credit to @reidab)

PR #563, opened 2022-04-15
Make backends not required on compute@edge resource

jmoney:
Currently, the compute@edge resource requires at least one backend to create the resource however this is actually not required of compute@edge services.  Aligning the terraform resource with the compute@edge service.


Integralist (changes requested):
👋🏻 Thanks for opening this PR (and your patience waiting for feedback).

The changes look good to me. 

I've pulled your branch locally and validated the changes manually (all works as expected). 

I've also run the complete integration test suite, which passes as well.

I have only one change request, which is to update the example `examples/resources/service_compute_basic_usage.tf` to remove the `backend` block and to regenerate the docs by running `make generate-docs`.

Thanks!

jmoney:
I updated the example file and regenerated the doc

jmoney:
Let me know if anything else is required for this @Integralist

Integralist (approved):

PR #565, opened 2022-04-25
fix(data_source_fastly_tls_configuration): listTLSConfigurations() pagination

### LINKED ISSUE #566: data source fastly_tls_configurations with default = true can return more than one result
Hi folks, thanks for your great work on this provider. I ran into a bug where accounts with a *single* TLS configuration (not the norm) can cause an unexpected error due to how pagination is handled. I link to a fix below.

### Terraform Version
1.1.7

### Affected Resource(s)
- data.fastly_tls_configuration

This is relevant if you have ONE tls configuration -- the bug is not present if you have between 2 and 100.
```hcl
data "fastly_tls_configuration" "default" {
  default = true
  # bug is also present if you specify the name or any other attribute besides id
}

output "id" {
  value = data.fastly_tls_configuration.default.id
}
```

### Debug Output
I am willing to share this debug output with a Fastly support engineer, but it contains too much sensitive information to be worth redacting for the public internet. The key behaviour is visible when the account has a single TLS configuration: you can see the provider make a request to `/tls/configurations`, receive a page with a single record, then make a request to `/tls/configurations?page[number]=1`, receive the same page with a single record, then make a request to `/tls/configurations?page[number]=2` and stop. The bug fix linked below contains an explanation of the root cause of the bug.

In the case where there are more than 100 records (the default page size of the /tls/configurations endpoint): the provider would make a request to `/tls/configurations`, receive a page with the first 100 records, then request `tls/configurations?page[number]=100` instead of `tls/configurations?page[number]=2` and so on.

### Expected Behavior
The data source should correctly resolve the default TLS configuration in my Fastly account.

### Actual Behavior
Terraform exited with:
```
Error: Your query returned more than one result. Please change try a more specific search criteria and try again.

  with data.fastly_tls_configuration.default,
  on main.tf line 1, in data "fastly_tls_configuration" "default":
  data "fastly_tls_configuration" "default"
```

### Steps to Reproduce
0. Create an account with only one Fastly TLS configuration, set to default.
1. `terraform apply`

### Important Factoids
Remember, this bug is only present if you have ONE TLS configuration (not the norm) or over 100 (the default page size for the `/tls/configurations` endpoint).

### References
- https://github.com/fastly/terraform-provider-fastly/pull/565

oopsbagel:
Fix TLS configuration pagination to iterate over result pages correctly.

Previously, if a list with only one TLS configuration was returned, the
result contained two duplicate configurations due to a pagination error.

When the same page is fetched twice, multiple entries for the same
configurations are added to the result array. This has the effect that
querying with a filter of `default = true` or on a specific unique name
will return multiple copies of the same result, causing an error in
dataSourceFastlyTLSConfigurationRead() where len(configurations) > 1.

The argument of PageNumber: 0 is dropped because of [0].
Page indexes start at 1, so PageNumber: 0 and PageNumber: 1 query the
same page.

cursor should increment one page at a time in order not to skip values,
and should start at 1 instead of 0 in order not to fetch the same page
twice.

Similarly, if /tls/configurations returned a result spanning multiple
pages containing N results (defaults to 100), the next queries would be
to pages N, 2*N, &c instead of pages 2, 3, ...

[0] https://github.com/fastly/go-fastly/blob/f1e3c717bfe88b473c08dd55aa0f9796fc932c71/fastly/custom_tls_configuration.go#L58

Fixes https://github.com/fastly/terraform-provider-fastly/issues/566

### DIFF
--- a/fastly/data_source_fastly_tls_configuration.go
+++ b/fastly/data_source_fastly_tls_configuration.go
@@ -184,7 +184,7 @@ func getTLSConfigurationFilters(d *schema.ResourceData) []func(*fastly.CustomTLS
 
 func listTLSConfigurations(conn *fastly.Client, filters ...func(*fastly.CustomTLSConfiguration) bool) ([]*fastly.CustomTLSConfiguration, error) {
 	var configurations []*fastly.CustomTLSConfiguration
-	cursor := 0
+	cursor := 1
 	for {
 		list, err := conn.ListCustomTLSConfigurations(&fastly.ListCustomTLSConfigurationsInput{
 			PageNumber: cursor,
@@ -196,7 +196,7 @@ func listTLSConfigurations(conn *fastly.Client, filters ...func(*fastly.CustomTL
 		if len(list) == 0 {
 			break
 		}
-		cursor += len(list)
+		cursor += 1
 
 		for _, configuration := range list {
 			if filterTLSConfiguration(configuration, filters) {
--- a/fastly/data_source_fastly_tls_configuration.go
+++ b/fastly/data_source_fastly_tls_configuration.go
@@ -132,7 +132,7 @@ func dataSourceFastlyTLSConfigurationRead(_ context.Context, d *schema.ResourceD
 		}
 
 		if len(configurations) > 1 {
-			return diag.Errorf("Your query returned more than one result. Please change try a more specific search criteria and try again.")
+			return diag.Errorf("Your query returned more than one result. Please use a more specific search criteria and try again.")
 		}
 
 		configuration = configurations[0]


Integralist (commented):
This PR looks good to me. Thank you for digging into this @oopsbagel 

My only request would be for a new test to be created that validates the 'single TLS configuration returned' scenario.

**EDIT**: I appreciate writing a test for this scenario might not be that straight forward, so failing that, a confirmation that the complete acceptance test suite is passing when run against your service (which only has a single TLS configuration) would suffice. I'll pull your code and run it against my own test service and will confirm shortly the results also.

Thanks.

Integralist:
Just to confirm that the full acceptance test suite passes when run against my own test service.

oopsbagel:
Thanks for your attention, @Integralist. The tests also pass for me.

I concur that writing a unit test to confirm this behaviour is a fairly large ask, especially since Terraform does not seem to provide much tooling for mocking out API responses for provider tests. I'd definitely prefer to get this merged without writing a specific test, since an end to end acceptance test against the real API is more comprehensive and less brittle.

PR #567, opened 2022-04-27
Avoid N backend API calls with new data exposed by `Director` type.

Integralist:
This PR relates to the recent go-fastly API client update:
https://github.com/fastly/go-fastly/pull/347

### DIFF
--- a/fastly/block_fastly_service_director.go
+++ b/fastly/block_fastly_service_director.go
@@ -10,10 +10,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+// DirectorServiceAttributeHandler implements ServiceAttributeDefinition.
 type DirectorServiceAttributeHandler struct {
 	*DefaultServiceAttributeHandler
 }
 
+// NewServiceDirector constructs a service attribute.
 func NewServiceDirector(sa ServiceMetadata) ServiceAttributeDefinition {
 	return ToServiceAttributeDefinition(&DirectorServiceAttributeHandler{
 		&DefaultServiceAttributeHandler{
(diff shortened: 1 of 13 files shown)


Integralist:
cc @jdigory who reported this issue.

fastly/block_fastly_service_director.go
```diff
@@ -10,10 +10,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+// DirectorServiceAttributeHandler implements ServiceAttributeDefinition.
```
Integralist, line 13: My linter was complaining about the lack of doc strings, so I've added them where required.

fastly/block_fastly_service_director.go
```diff
@@ -130,45 +134,18 @@ func (h *DirectorServiceAttributeHandler) Create(_ context.Context, d *schema.Re
 	return nil
 }
 
+// Read refreshes the resource state.
 func (h *DirectorServiceAttributeHandler) Read(_ context.Context, d *schema.ResourceData, _ map[string]interface{}, serviceVersion int, conn *gofastly.Client) error {
 	log.Printf("[DEBUG] Refreshing Directors for (%s)", d.Id())
 	directorList, err := conn.ListDirectors(&gofastly.ListDirectorsInput{
```
Integralist, line 140: In earlier versions of go-fastly the `Director` type returned from the `ListDirectors` API call was missing a `Backends` field. This meant this `Read` function had to make an API call to get all the service backends and to figure out which ones were members of the Director.

The following change removes the API call for service backends.

fastly/block_fastly_service_director.go
```diff
@@ -297,11 +272,11 @@ func flattenDirectors(directorList []*gofastly.Director, directorBackendList []*
 			"retries": int(d.Retries),
 		}
 
+		// NOTE: schema.NewSet expects slice of empty interface so we have to build
+		// this from the Dictionary's Backend field.
 		var b []interface{}
-		for _, db := range directorBackendList {
```
Integralist, line 301: Now instead of looping over a list of backends that came from a service backend API call and were subsequently filtered down, we're looping over the backends that were exposed via the Director itself.

Integralist (commented):
Comments to help the reviewer...

kailan (approved):

PR #572, opened 2022-06-16
Service authorization

noseglid:
This change adds support for [Fastly Service Authorizations][1] to be
managed via Terraform.

Other managed users and services can be referenced by their ID and the
permission is controlled with an enumeration of the allowed values by
the API (`full`, `read_only`, `purge_select` or `purge_all`).

[1]: https://developer.fastly.com/reference/api/account/service-authorization/


go.mod
```diff
@@ -12,3 +12,5 @@ require (
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/net v0.0.0-20211015210444-4f30a5c0130f
 )
+
+replace github.com/fastly/go-fastly/v6 => github.com/noseglid/go-fastly/v6 v6.0.0-20220615131558-6b72cc694806
```
noseglid: This needs to be replaced before this PR is merged. It requires https://github.com/fastly/go-fastly/pull/349 to be merged and released first.

examples/resources/service_authorization_basic_usage.tf
```diff
@@ -0,0 +1,11 @@
+resource "fastly_service_vcl" "myservice" {
```
Integralist: Can we add some example fields to each section, as this will make it more consistent with the other example files. For example, the [ACL Entries example](https://github.com/fastly/terraform-provider-fastly/blob/78fe6303e670d18a21884cabc610a93a9272a748/examples/resources/service_acl_entries_basic_usage.tf) provides configuration for all resources defined to make the example as complete as possible.

Integralist: I would also rename `"myservice"` to `"demo"` so the following `examples/resources/service_authorization_import.txt` file will align.

> **NOTE**: Our use of `myservice` and `demo` are a bit inconsistent, but looks like `demo` is used more.

fastly/resource_fastly_service_authorization.go
```diff
@@ -0,0 +1,115 @@
...
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The ID of this service authorization.",
+			},
+			"permission": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "full",
```
Integralist: I'm actually surprised the API _defaults_ to "full". 

I feel it would be safer to default to "read_only" and have users explicitly opt into more permissive values. 

> **NOTE**: I've reached out to the relevant API team about this as I'm interested to know if they agree that we should default to "read_only" (and if so, then we should update the API endpoint as well to reflect that).

fastly/resource_fastly_service_authorization.go
```diff
@@ -0,0 +1,115 @@
...
+	d.Set("user_id", sa.UserID)
+	d.Set("permission", sa.Permission)
+
+	return nil
+}
+
+func resourceServiceAuthorizationUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	conn := meta.(*FastlyClient).conn
+
+	// Update Name and/or Role.
```
Integralist: Not sure what this line refers to? 

I think it's a side-effect of copy/pasting from another file (`fastly/resource_fastly_user.go`?)

Integralist (commented):
Thanks @noseglid for opening this PR.

Other than a couple of minor comments I have (and also the changes you mentioned regarding ensuring we use a new go-fastly release, once your associated PR is merged/released) this PR LGTM.

fastly/resource_fastly_service_authorization.go
noseglid: That's exactly it 😅  I'll sort it.

fastly/resource_fastly_service_authorization.go
noseglid: I agree. `read_only` would absolutely be a better default.

Alternatively, there could be *no default*, and explicitly require users to set the permissions they desire. Better to be explicit in my opinion.

examples/resources/service_authorization_basic_usage.tf
noseglid: Sure, I will change to `demo`.

fastly/resource_fastly_service_authorization.go
Integralist: Agreed. Let's avoid the default and users have to be explicit about what they want. 👍🏻

noseglid:
This PR is now updated using the main branch of go-fastly/fastly. I will update it again once we have a tagged version to use.

Integralist:
@noseglid I've cut a new go-fastly release now https://github.com/fastly/go-fastly/releases/tag/v6.4.0

noseglid:
@Integralist Any chance to get a new release with these changes included?

Integralist:
Hi @noseglid 

Absolutely. I was planning on cutting a release after merging this PR but it was then brought to my attention (via another customer) that the go-fastly change set was missing the list endpoint and that took up my focus for a bit.

So (in case it's of interest to you) I've since added support for listing service authorizations, and also cut that into a new go-fastly release:
https://github.com/fastly/go-fastly/releases/tag/v6.4.1

I can't cut a new Terraform release unless all integration tests are passing, which at the moment there are a couple of failing tests related to changes made in our API layer, so I have a PR open that fixes those, and I'm just waiting a member of my team to review/approve:
https://github.com/fastly/terraform-provider-fastly/pull/573

Once that's done I'll cut a new `v2.1.0` release and ping here to let you know when it's done.

noseglid:
That sounds great. Thank you for the update!

Integralist:
@noseglid Sorry for the delay:
https://github.com/fastly/terraform-provider-fastly/releases/tag/v2.1.0

noseglid:
Fantastic, thank you!

PR #573, opened 2022-06-23
Fix integration tests

Integralist:
**Summary**:
- Replace `geoip.country_code` (deprecated; causes Varnish compiler error) with `client.geo.country_code`.
- Remove check for `shield` property on a POP (not every POP has a shield property).


harmony7 (approved):
looks ok to me

PR #578, opened 2022-07-19
allow service to be reused

### LINKED ISSUE #515: Feature request: Add option to deactivate service as part of terraform destroy
### Terraform Version
1.0.11

### Affected Resource(s)
- fastly_service_v1

### Description
Currently if you have a service with active version, there are 2 options with terraform destroy:
- `force_destroy = true` which deletes the service
- `force_destroy = false` which causes the terraform to fail

We would very much need to be able to deactivate the service with destroy without causing it to fail. This is what I have expected to be a default behavior with terraform destroy being run while having `force_destroy` set to `false` and I'm sorry, but I have to ask: what even is the usecase for having it set to `false` if not deactivating the service? Deleting services that are not active, thus forcing manual deactivation beforehand? If I don't want the service to be affected by the terraform destroy command, I have better options like removing it from the state etc...

Currently the best we can do is to let the terraform destroy fail and deactivate it directly via Fastly REST API, which kind of defeats the purpose of the terraform provider.

Integralist:
Fixes #515

**Summary**: add `reuse` attribute to `fastly_service_compute` and `fastly_service_vcl` that causes `terraform destroy` to not fully delete the service.

See my notes for the full details and my manual validation of this change set, and the various caveats:
https://github.com/fastly/terraform-provider-fastly/issues/515#issuecomment-1172187386

**NOTE**: I've manually tested this feature, and found it to work (see [comment](https://github.com/fastly/terraform-provider-fastly/issues/515#issuecomment-1172187386)) but there are no tests written because the Terraform test framework expects resources to be destroyed at the end of each e2e test, and this PR implements a feature that sits outside that scenario, and so any test we write would fail unless written in a really janky way and additionally we'd need to implement a post-test process to then clean-up the service resource that was set to persist being destroyed.

**Flow Examples**

- `terraform destroy`: service not activated
  - no attributes set == delete service API call (successful)
  - `force_destroy` == no deactivation service API call, delete service API call (successful)
  - `reuse` == no deactivation service API call, no delete service API call (basically a no-op)
- `terraform destroy`: service activated
  - no attributes set == delete service API call (fails -- as I can't delete an active service without deactivating first)
  - `force_destroy` == deactivate service API call (successful), delete service API call (successful)
  - `reuse` == deactivate service API call (successful)

### DIFF
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -105,7 +105,7 @@ $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx@2
 - **logging_splunk** (Block Set) (see [below for nested schema](#nestedblock--logging_splunk))
 - **logging_sumologic** (Block Set) (see [below for nested schema](#nestedblock--logging_sumologic))
 - **logging_syslog** (Block Set) (see [below for nested schema](#nestedblock--logging_syslog))
-- **reuse** (Boolean) Services that are active cannot be destroyed. This will cause the service to be deactivated allowing it to be reused by importing it into another Terraform project. Default `false`
+- **reuse** (Boolean) Services that are active cannot be destroyed. If set to `true` a service Terraform intends to destroy will instead be deactivated (allowing it to be reused by importing it into another Terraform project). If `false`, attempting to destroy an active service will cause an error. Default `false`
 - **version_comment** (String) Description field for the version
 
 ### Read-Only
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -105,6 +105,7 @@ $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx@2
 - **logging_splunk** (Block Set) (see [below for nested schema](#nestedblock--logging_splunk))
 - **logging_sumologic** (Block Set) (see [below for nested schema](#nestedblock--logging_sumologic))
 - **logging_syslog** (Block Set) (see [below for nested schema](#nestedblock--logging_syslog))
+- **reuse** (Boolean) Services that are active cannot be destroyed. This will cause the service to be deactivated allowing it to be reused by importing it into another Terraform project. Default `false`
 - **version_comment** (String) Description field for the version
 
 ### Read-Only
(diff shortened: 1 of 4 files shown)


fastly/base_fastly_service.go
```diff
@@ -544,11 +549,16 @@ func resourceServiceDelete(_ context.Context, d *schema.ResourceData, meta inter
 		}
 	}
 
-	err := conn.DeleteService(&gofastly.DeleteServiceInput{
-		ID: d.Id(),
-	})
-	if err != nil {
-		return diag.FromErr(err)
+	if !d.Get("reuse").(bool) {
```
triblondon, line 551: Would it be clearer to make this a check for force_destroy instead of not reuse?

docs/resources/service_compute.md
```diff
@@ -105,6 +105,7 @@ $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx@2
 - **logging_splunk** (Block Set) (see [below for nested schema](#nestedblock--logging_splunk))
 - **logging_sumologic** (Block Set) (see [below for nested schema](#nestedblock--logging_sumologic))
 - **logging_syslog** (Block Set) (see [below for nested schema](#nestedblock--logging_syslog))
+- **reuse** (Boolean) Services that are active cannot be destroyed. This will cause the service to be deactivated allowing it to be reused by importing it into another Terraform project. Default `false`
```
triblondon: The first time I read this I thought it meant "Services that are active cannot be destroyed *if you set this property to true*".  However now I see the impl, I realise it doesn't mean that.  

```suggestion
- **reuse** (Boolean) Services that are active cannot be destroyed. If set to `true` a service Terraform intends to destroy will instead be deactivated (allowing it to be reused by importing it into another Terraform project). If `false`, attempting to destroy an active service will cause an error. Default `false`
```

triblondon (approved):

fastly/base_fastly_service.go
Integralist: No, because if the user doesn't specify anything then this breaks the current behaviour.

e.g. currently, if the user has a deactivated service and they don't specify `force_destroy`, then the service will successfully be deleted. By changing the condition to be `d.Get("force_destroy").(bool)` we're effectively saying they _have_ to add `force_destroy` to delete their deactivated service. By negating the reuse check we can keep the existing behaviour while supporting the new behaviour.

PR #579, opened 2022-07-20
Splunk token required

### LINKED ISSUE #577: Opaque 400 Error
Hi there,

I'm getting an opaque error. Its likely caused by misconfiguration on my part, but the error message could be more helpful (though this may be coming directly from the Fastly API itself).

### Terraform Version
1.1.7

### Affected Resource(s)
- fastly_service_vcl

```hcl
resource "fastly_service_vcl" "assets_cdn" {
  name = var.fastly_service_name

  domain {
    name    = var.domain_name
    comment = "Assets Domain"
  }

  // Origin
  backend {
    name                  = "Origin"
    address               = "${var.s3_bucket_name}.s3-${var.s3_region}.amazonaws.com"
    override_host         = "${var.s3_bucket_name}.s3-${var.s3_region}.amazonaws.com"
    port                  = 443
    ssl_cert_hostname     = "*.s3-${var.s3_region}.amazonaws.com"
    use_ssl               = true
    auto_loadbalance      = false
    connect_timeout       = 5000
    first_byte_timeout    = 15000
    between_bytes_timeout = 100000
    error_threshold       = 0
  }

  gzip {
    name       = "Compression Policy"
    extensions = [
      "css",
      "js",
      "html",
      "eot",
      "ico",
      "otf",
      "ttf",
      "json",
      "svg"
    ]
    content_types = [
      "text/html",
      "application/x-javascript",
      "text/css",
      "application/javascript",
      "text/javascript",
      "application/json",
      "application/vnd.ms-fontobject",
      "application/x-font-opentype",
      "application/x-font-truetype",
      "application/x-font-ttf",
      "application/xml",
      "font/eot",
      "font/opentype",
      "font/otf",
      "image/svg+xml",
      "image/vnd.microsoft.icon",
      "text/plain",
      "text/xml"
    ]
  }

  logging_splunk {
    name   = "Web logs to splunk"
    url    = "https://${local.splunk_domain}.splunkcloud.com:443/services/collector/event"
    format = file("${local.config_dir}/logging/fastly-web-splunk-format.json")
  }

  vcl {
    name    = "main.vcl"
    content = file("${local.config_dir}/vcl/main.vcl")
    main    = true
  }

  vcl {
    name    = "fastly-boilerplate-begin.vcl"
    content = file("${local.config_dir}/vcl/fastly-boilerplate-begin.vcl")
  }

  vcl {
    name    = "fastly-boilerplate-end.vcl"
    content = file("${local.config_dir}/vcl/fastly-boilerplate-end.vcl")
  }

}
```

### Expected Behavior
The service should deploy.

### Actual Behavior
When doing terraform deploy, changes are deployed seemingly succesfully, however terraform returns the below error. Though it doesn't say what token/key is missing, for what purpose (is it a fastly key missing, s3, splunk?).
```
module.fastly_service.fastly_service_vcl.assets_cdn: Modifying... [id=<my-id>]
╷
│ Error: 400 - Bad Request:
│ 
│     Title:  Bad request
│     Detail: missing required token/key for authentication
│ 
│   with module.fastly_service.fastly_service_vcl.assets_cdn,
│   on ../../modules/fastly_service/service.tf line 5, in resource "fastly_service_vcl" "assets_cdn":
│    5: resource "fastly_service_vcl" "assets_cdn" {
│ 
╵
Releasing state lock. This may take a few moments...

Exited with code exit status 1
```

### Steps to Reproduce
terraform apply

Integralist:
Fixes https://github.com/fastly/terraform-provider-fastly/issues/577

**NOTES**: The customer initially reported the Fastly API was returning an ambiguous error. The error messaging was intentional as the API internals are sharing an implementation across multiple types, which means the error had to be generic in wording. 

To help avoid confusion a new version of go-fastly was released (v6.4.3) with additional checks for the `Token` field. This PR bumps the go-fastly dependency to the new version.

I then also noticed a bug in the Terraform provider where the use of `schema.EnvDefaultFunc` meant that if no environment variable (i.e. `FASTLY_SPLUNK_TOKEN`) had been set, then an empty string would be used. Returning an empty string causes Terraform to use the value as the fallback and sends the empty string to go-fastly. 

Instead of an empty string, if the helper function returned `nil`, then Terraform would enforce the 'required field' behaviour on the `Token` attribute and would prompt the user to add a value to their HCL configuration. Thus avoiding having to even call out to go-fastly (only for go-fastly's new check for a token value to fail).

### DIFF
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -595,6 +595,7 @@ Optional:
 Required:
 
 - **name** (String) A unique name to identify the Splunk endpoint. It is important to note that changing this attribute will delete and recreate the resource
+- **token** (String, Sensitive) The Splunk token to be used for authentication
 - **url** (String) The Splunk URL to stream logs to
 
 Optional:
@@ -603,7 +604,6 @@ Optional:
 - **tls_client_cert** (String) The client certificate used to make authenticated requests. Must be in PEM format.
 - **tls_client_key** (String, Sensitive) The client private key used to make authenticated requests. Must be in PEM format.
 - **tls_hostname** (String) The hostname used to verify the server's certificate. It can either be the Common Name or a Subject Alternative Name (SAN)
-- **token** (String, Sensitive) The Splunk token to be used for authentication
 - **use_tls** (Boolean) Whether to use TLS for secure logging. Default: `false`
 
 
--- a/docs/resources/service_vcl.md
+++ b/docs/resources/service_vcl.md
@@ -1009,6 +1009,7 @@ Optional:
 Required:
 
 - **name** (String) A unique name to identify the Splunk endpoint. It is important to note that changing this attribute will delete and recreate the resource
+- **token** (String, Sensitive) The Splunk token to be used for authentication
 - **url** (String) The Splunk URL to stream logs to
 
 Optional:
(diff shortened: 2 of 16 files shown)


kailan (approved):
Looks good to me!

PR #614, opened 2022-11-14
Bump go-fastly to new v7 major release

Integralist:
This is technically a breaking-change PR because we remove a long deprecated attribute. Also, the underlying go-fastly changes are so substantial that it's probably worth moving to a new major version.


go.mod
```diff
@@ -4,11 +4,11 @@ go 1.18
 
 require (
 	github.com/bflad/tfproviderlint v0.28.1
-	github.com/fastly/go-fastly/v6 v6.8.0
+	github.com/fastly/go-fastly/v7 v7.0.0
 	github.com/google/go-cmp v0.5.9
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/terraform-plugin-docs v0.13.0
-	github.com/hashicorp/terraform-plugin-sdk/v2 v2.10.1
+	github.com/hashicorp/terraform-plugin-sdk/v2 v2.24.0
```
Integralist, line 11: ℹ️ I had to bump this dependency to fix a failing test caused by a bug in the Terraform SDK.

fastly/resource_fastly_service_acl_entries.go
```diff
@@ -232,33 +232,34 @@ func resourceServiceACLEntriesDelete(_ context.Context, d *schema.ResourceData,
...
 func flattenACLEntries(aclEntryList []*gofastly.ACLEntry) []map[string]any {
-	var resultList []map[string]any
-
-	for _, currentACLEntry := range aclEntryList {
-		aes := map[string]any{
-			"id":      currentACLEntry.ID,
-			"ip":      currentACLEntry.IP,
-			"negated": currentACLEntry.Negated,
-			"comment": currentACLEntry.Comment,
+	var result []map[string]any
```
Integralist, line 237: ℹ️ I made a bunch of consistency changes to variable naming.

fastly/block_fastly_service_logging_s3.go
```diff
@@ -76,9 +76,13 @@ func (h *S3LoggingServiceAttributeHandler) GetSchema() *schema.Schema {
 			Default:     "s3.amazonaws.com",
 		},
 		"gzip_level": {
-			Type:        schema.TypeInt,
-			Optional:    true,
-			Default:     0,
+			Type:     schema.TypeInt,
+			Optional: true,
+			// NOTE: The default represents an unset value
```
Integralist, line 81: ℹ️ The changes in go-fastly cause complications for the Terraform provider.

go-fastly v7+ now expects a pointer, but the Terraform provider originally set a 'default' value of zero for the gzip_level attribute (which is a valid API value). This meant the default value would be wrapped in a pointer and consequently was always sent to the API, but this caused the API to error as some combinations of attributes are considered conflicting.  

For example, configuring compression_codec + gzip_level is an invalid API request and was being triggered by our test suite which explicitly defined compression_codec in Terraform configuration while gzip_level was now being added to the API request because of the default value.

We weren't able to determine if the user actually configured gzip_level in their Terraform configuration because zero might well be a value they would set (as it's a valid value for that attribute). This meant we needed to set a default value (i.e. -1) that we could use to determine if the user had set the attribute in their Terraform configuration (see the updated 'flatten' function).

Additionally, some fields shouldn't have an empty string passed as it would cause the Fastly API to return an error. Because of go-fastly's new behaviour it now will not 'omitempty' empty strings due to the pointer field type.

fastly/block_fastly_service_dynamicsnippet.go
```diff
@@ -37,6 +37,12 @@ func (h *DynamicSnippetServiceAttributeHandler) GetSchema() *schema.Schema {
 		Optional: true,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
+				"content": {
```
Integralist, line 40: ℹ️ We never actually defined an attribute for a user to create/update VCL content for a dynamic snippet.

fastly/block_fastly_service_backend.go
```diff
@@ -152,18 +152,11 @@ func (h *BackendServiceAttributeHandler) GetSchema() *schema.Schema {
 			Description: "Client key attached to origin. Used when connecting to the backend",
 			Sensitive:   true,
 		},
-		"ssl_hostname": {
```
Integralist, line 155: This has been deprecated in go-fastly for nearly two years and was removed in v7, which means by removing this attribute here we'll need to cut a new major Terraform release.

Integralist (commented):
NOTES for reviewer...

shawnps (approved):

Integralist:
**NOTE**: I've run the full integration test suite and it's passing.

PR #618, opened 2022-11-22
Bug: Backends send empty string to API

Integralist:
**Problem**:
When creating a backend, if you set the port to 443 the API will expect SSL/TLS attributes to not be sent with an empty string, and that is what was happening in the latest Terraform v3 release. 

**Context** :
The Terraform provider wraps all values in a pointer (as that is what go-fastly v7+ expects), but the zero value for an attribute (e.g. `""` for the string type) is what will be wrapped inside the pointer and consequently sent to the API via go-fastly.

**Solution**:
As with other attributes, once we switched to port 443, we were able to use the integration/acceptance tests to identify which attributes would cause API errors and consequently avoid setting those fields on the go-fastly struct.


Integralist:
I've run the full integration test suite and it's passing so I will go ahead and merge this PR.

PR #620, opened 2022-11-23
Bug: Some logging providers were sending an empty `placement` to the Fastly API

Integralist:
**Problem**:
When creating a logging provider the `placement` attribute is optional and the API will expect it to not be sent with an empty value, but that is what was happening in the latest Terraform v3 release. 

**Context** :
The Terraform provider wraps all values in a pointer (as that is what go-fastly v7+ expects), but the zero value for an attribute (e.g. `""` for the string type) is what will be wrapped inside the pointer and consequently sent to the API via go-fastly. Previously we had used the integration/acceptance tests to identify which attributes would cause API errors but what happened in this instance was that the tests were setting a value for that field and so no error was reported.

**Solution**:
Do as we have already done with the other logging providers and avoid setting the `placement` field on the go-fastly struct.


Integralist:
I've run the integration test suite and it is passing.

PR #622, opened 2022-11-28
Prevent SSL related fields from being sent empty to the Fastly API.

Integralist:
**NOTE**: While updating all the SSL fields for a backend to not be sent empty to the API (unless the user explicitly set them to be empty) I noticed that setting `use_ssl` would cause an API error... 

```
Error: invalid configuration for Fastly Service (<REDACTED>): 
Mandatory SSL cert checks require specifying cert hostname
```

To fix this you have to also specify `ssl_cert_hostname`, so I've updated the tests to reflect this.

### DIFF
--- a/fastly/block_fastly_service_backend_test.go
+++ b/fastly/block_fastly_service_backend_test.go
@@ -171,6 +171,10 @@ func TestAccFastlyServiceVCLBackend_basic(t *testing.T) {
 	backendName := fmt.Sprintf("backend-tf-%s", acctest.RandString(10))
 	backendAddress := "httpbin.org"
 
+	// The following backends are what we expect to exist after all our Terraform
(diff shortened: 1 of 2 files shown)


fastly/block_fastly_service_backend_test.go
```diff
@@ -171,6 +171,10 @@ func TestAccFastlyServiceVCLBackend_basic(t *testing.T) {
 	backendName := fmt.Sprintf("backend-tf-%s", acctest.RandString(10))
 	backendAddress := "httpbin.org"
 
+	// The following backends are what we expect to exist after all our Terraform
```
doramatadora, line 174: 🙌🏼

doramatadora (approved):

PR #630, opened 2023-01-06
fix: force refresh when service version is reverted outside of Terraform

### LINKED ISSUE #629: Attributes aren't refreshed if no attributes of that type exist in local state
### Terraform Version
Terraform v1.3.3
on darwin_arm64

### Affected Resource(s)
- fastly_service_vcl

```terraform
resource "fastly_service_vcl" "cache-setting-test" {
  name = "cache-setting-test"
  domain {
      name = "cache-setting-test.example.com"
  }
  cache_setting {
    action          = "pass"
    cache_condition = "Do not cache non-render responses or non-200"
    name            = "redirect cache setting"
    stale_ttl       = 0
    ttl             = 0
  }
  condition {
    name      = "Do not cache non-render responses or non-200"
    priority  = 100
    statement = "beresp.status != 200 || req.url !~ \"^/render\""
    type      = "CACHE"
  }
}

```

It appears that most of the attribute handlers intentionally avoid refreshing if the local state has no attributes of that type - for example:

https://github.com/fastly/terraform-provider-fastly/blob/main/fastly/block_fastly_service_cachesetting.go#L90-L93

This can cause problems if the local state has become out of sync - e.g. if the service has been rolled back. Is this intentional?

### Expected Behavior

A terraform refresh should update the local state with the complete current state of the Fastly service.

### Actual Behavior

A terraform refresh does not pick up attributes from the service if there are no attributes of that type already existing in the local state.

### Steps to Reproduce
1. Create a service as v1 using the terraform config above
2. Delete the `cache_setting` attribute block and `terraform apply`, creating v2 of the service
3. Rollback the service to v1 via the Fastly UI
4. Run `terraform apply` again - it will now report that no changes are needed, and so will not remove the `cache_setting` attribute. Running `terraform refresh` does not help.

Integralist:
Fixes #629

**NOTE:** I tested this locally by building the provider and following the following steps...

1. Create a new fastly_service_vcl service with two domains.
2. Delete one of the domains and plan/apply the change (creating service version 2).
3. Rollback the service to version 1 via the Fastly UI.
4. Run terraform plan and it should display the same plan as step 2.
5. Once plan is applied, then `terraform show` reveals service is now at version 3 and has the correct resources.

### DIFF
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -111,6 +111,7 @@ $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx@2
 
 - `active_version` (Number) The currently active version of your Fastly Service
 - `cloned_version` (Number) The latest cloned version by the provider
+- `force_refresh` (Boolean) Used internally by the provider to temporarily indicate if all resources should call their associated API to update the local state. This is for scenarios where the service version has been reverted outside of Terraform (e.g. via the Fastly UI) and the provider needs to resync the state for a different active version (this is only if `activate` is `true`).
 - `id` (String) The ID of this resource.
 - `imported` (Boolean) Used internally by the provider to temporarily indicate if the service is being imported, and is reset to false once the import is finished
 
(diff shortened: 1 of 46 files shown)


kailan (approved):
Looks good 👍

jonnangle:
Thanks for this @Integralist!

PR #633, opened 2023-01-18
Update service_dictionary_items.md

emelgar-tek:
Fixed previous note to use the proper markdown for a note. Note is for importing an edge dictionary and ensuring it's empty before importing. If the dictionary resource is not empty  the import fails silently.

### DIFF
--- a/docs/resources/service_dictionary_items.md
+++ b/docs/resources/service_dictionary_items.md
@@ -230,6 +230,8 @@ resource "fastly_service_dictionary_items" "items" {
 
 ## Import
 
+~> **Note:** The dictionary resouce should be empty before importing
+
 This is an example of the import command being applied to the resource named `fastly_service_dictionary_items.items`
 The resource ID is a combined value of the `service_id` and `dictionary_id` separated by a forward slash.
 


docs/resources/service_dictionary_items.md
```diff
@@ -230,6 +230,8 @@ resource "fastly_service_dictionary_items" "items" {
 
 ## Import
 
+~> **Note:** The dictionary resouce should be empty before importing
```
Integralist, line 233:
```suggestion
~> **Note:** The dictionary resource should be empty before importing
```

Integralist (changes requested):
Hi @emelgar-tek 

Thanks for opening this PR.

You've made changes to the _generated_ documentation rather than the source files.

You need to update this file I believe:
https://github.com/fastly/terraform-provider-fastly/blob/main/templates/resources/service_dictionary_items.md.tmpl

Once you've updated that file, run `make generate-docs` and then commit all the changes.

Thanks!

Integralist:
Closing this in favour of https://github.com/fastly/terraform-provider-fastly/pull/634

PR #658, opened 2023-03-07
Enable the declaration of the keepalive_time backend attribute

loxal:


fastly/block_fastly_service_backend.go
```diff
@@ -73,6 +73,12 @@ func (h *BackendServiceAttributeHandler) GetSchema() *schema.Schema {
 			Default:     "",
 			Description: "Name of a defined `healthcheck` to assign to this backend",
 		},
+		"keepalive_time": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Default:     null,
```
Integralist: 👋🏻 @loxal I'll ping you once I cut a new go-fastly release with the new KeepAliveTime field but in the mean time this particular line of code is invalid (Go doesn't have a `null` keyword). You might have been thinking of `nil` but you wouldn't be able to use that as a default value for an integer. I would just leave the `Default` field unset and later on in your code check if the integer is a value above zero (because realistically no one is going to set the value to be zero, which is the default value for an int type).

fastly/block_fastly_service_backend.go
```diff
@@ -276,6 +282,7 @@ func (h *BackendServiceAttributeHandler) buildCreateBackendInput(service string,
 		ErrorThreshold:      gofastly.Int(resource["error_threshold"].(int)),
 		FirstByteTimeout:    gofastly.Int(resource["first_byte_timeout"].(int)),
 		HealthCheck:         gofastly.String(resource["healthcheck"].(string)),
+		KeepAliveTime:       gofastly.Int(resource["keepalive_time"].(int)),
```
Integralist: We should probably only set this field on `opts` if the value provided is greater than zero. I don't think it makes sense to send zero (which is what it will be if the user doesn't define it in their config).

Integralist: So remove the field from here and underneath the `opts` declaration/assignment have something like...

```go
if resource["keepalive_time"].(int) > 0 {
    opts.KeepAliveTime = gofastly.Int(resource["keepalive_time"].(int))
}
```

fastly/block_fastly_service_backend.go
Integralist: @loxal I've now published https://github.com/fastly/go-fastly/releases/tag/v7.4.0

Integralist:
You'll also need to run `make generate-docs` locally and commit the updated documentation.

fastly/block_fastly_service_backend.go
```diff
@@ -37,6 +37,12 @@ func (h *BackendServiceAttributeHandler) GetSchema() *schema.Schema {
 			Required:    true,
 			Description: "An IPv4, hostname, or IPv6 address for the Backend",
 		},
+		"auto_loadbalance": {
```
Integralist: You might need to rebase `main` branch as this was updated recently...
https://github.com/fastly/terraform-provider-fastly/pull/657

fastly/block_fastly_service_backend.go
```diff
@@ -289,6 +295,9 @@ func (h *BackendServiceAttributeHandler) buildCreateBackendInput(service string,
 	// WARNING: The following fields shouldn't have an empty string passed.
 	// As it will cause the Fastly API to return an error.
 	// This is because go-fastly v7+ will not 'omitempty' due to pointer type.
+	if resource["keepalive_time"].(int) > 0 {
+		opts.KeepAliveTime = gofastly.Int(resource["keepalive_time"].(int))
+	}
```
Integralist, line 296: I'd actually move this conditional above the WARNING because although it's doing something similar, it's not for the same reason as the following attributes.

docs/resources/service_compute.md
```diff
@@ -150,11 +150,13 @@ Required:
 
 Optional:
 
+- `auto_loadbalance` (Boolean) Denotes if this Backend should be included in the pool of backends that requests are load balanced against. Default `false`
```
Integralist: 👋🏻 

Just to clarify that 'compute' services should NOT have `auto_loadbalance`.

The following PR was merged recently to _remove_ `auto_loadbalance` from the `fastly_service_compute` resource, so best to rebase `main` branch and ensure those changes aren't undone:
https://github.com/fastly/terraform-provider-fastly/pull/657

This means the `TestResourceFastlyFlattenBackendCompute` test  (line 137) should NOT add the `"auto_loadbalance":      false,`.

Thanks.

Integralist:
```suggestion

```

fastly/block_fastly_service_backend.go
```diff
@@ -37,6 +37,12 @@ func (h *BackendServiceAttributeHandler) GetSchema() *schema.Schema {
 			Required:    true,
 			Description: "An IPv4, hostname, or IPv6 address for the Backend",
 		},
+		"auto_loadbalance": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
+			Description: "Denotes if this Backend should be included in the pool of backends that requests are load balanced against. Default `false`",
+		},
```
Integralist:
```suggestion

```

fastly/block_fastly_service_backend.go
```diff
@@ -170,12 +181,6 @@ func (h *BackendServiceAttributeHandler) GetSchema() *schema.Schema {
 	required := false
 
 	if h.GetServiceMetadata().serviceType == ServiceTypeVCL {
```
Integralist:
```suggestion
	if h.GetServiceMetadata().serviceType == ServiceTypeVCL {
		blockAttributes["auto_loadbalance"] = &schema.Schema{
			Type:        schema.TypeBool,
			Optional:    true,
			Default:     false,
			Description: "Denotes if this Backend should be included in the pool of backends that requests are load balanced against. Default `false`",
		}
```

fastly/block_fastly_service_backend.go
```diff
@@ -270,6 +275,7 @@ func (h *BackendServiceAttributeHandler) createDeleteBackendInput(service string
 func (h *BackendServiceAttributeHandler) buildCreateBackendInput(service string, latestVersion int, resource map[string]any) gofastly.CreateBackendInput {
 	opts := gofastly.CreateBackendInput{
 		Address:             gofastly.String(resource["address"].(string)),
+		AutoLoadbalance:     gofastly.CBool(resource["auto_loadbalance"].(bool)),
```
Integralist:
```suggestion
```

fastly/block_fastly_service_backend.go
```diff
@@ -318,7 +327,6 @@ func (h *BackendServiceAttributeHandler) buildCreateBackendInput(service string,
 	}
 
 	if h.GetServiceMetadata().serviceType == ServiceTypeVCL {
```
Integralist:
```suggestion
	if h.GetServiceMetadata().serviceType == ServiceTypeVCL {
		opts.AutoLoadbalance = gofastly.CBool(resource["auto_loadbalance"].(bool))
```

fastly/block_fastly_service_backend.go
```diff
@@ -417,11 +428,13 @@ func flattenBackend(remoteState []*gofastly.Backend, sa ServiceMetadata) []map[s
 	for _, resource := range remoteState {
 		data := map[string]any{
 			"address":               resource.Address,
+			"auto_loadbalance":      resource.AutoLoadbalance,
```
Integralist:
```suggestion
```

fastly/block_fastly_service_backend.go
```diff
@@ -441,7 +454,6 @@ func flattenBackend(remoteState []*gofastly.Backend, sa ServiceMetadata) []map[s
 		}
 
 		if sa.serviceType == ServiceTypeVCL {
```
Integralist:
```suggestion
		if sa.serviceType == ServiceTypeVCL {
			data["auto_loadbalance"] = resource.AutoLoadbalance
```

fastly/block_fastly_service_backend_test.go
```diff
@@ -131,10 +134,12 @@ func TestResourceFastlyFlattenBackendCompute(t *testing.T) {
 					"address":               "www.notexample.com",
 					"override_host":         "origin.example.com",
 					"port":                  80,
+					"auto_loadbalance":      false,
```
Integralist:
```suggestion
```

Integralist (changes requested):

fastly/block_fastly_service_backend.go
```diff
@@ -289,6 +294,9 @@ func (h *BackendServiceAttributeHandler) buildCreateBackendInput(service string,
 	// WARNING: The following fields shouldn't have an empty string passed.
```
Integralist:
```suggestion
	if resource["keepalive_time"].(int) > 0 {
		opts.KeepAliveTime = gofastly.Int(resource["keepalive_time"].(int))
	}
	
	// WARNING: The following fields shouldn't have an empty string passed.
```

fastly/block_fastly_service_backend.go
```diff
@@ -289,6 +294,9 @@ func (h *BackendServiceAttributeHandler) buildCreateBackendInput(service string,
...
 	// As it will cause the Fastly API to return an error.
 	// This is because go-fastly v7+ will not 'omitempty' due to pointer type.
+	if resource["keepalive_time"].(int) > 0 {
+		opts.KeepAliveTime = gofastly.Int(resource["keepalive_time"].(int))
+	}
```
Integralist, line 296:
```suggestion

```

Integralist (approved):

PR #661, opened 2023-03-09
feat(fastly_service_compute): support new `content` attribute

### LINKED ISSUE #583: Feature request: Better way to provide package for compute services
Feature request to provide the Compute@Edge binary package by other means than a file name.

When deploying a compute@edge service one has to specify a file name in the `fastly_service_compute` > `package` > `filename` setting. Especially when deploying a service using Terraform Cloud triggered by version control this becomes problematic, because it forces one to commit the built binary package to git.

Instead of committing a 5MB or so file on each change, I would kindly request a new feature to improve this. Possible solutions could be

1. Provide the relative path to the source code in the `package` setting and then auto build from that.

```terraform
resource "fastly_service_compute" "test" {
  ...
  package {
    source = "./src/"
  }
}
```

2. Set content instead of filename. 

```terraform
resource "fastly_service_compute" "test" {
  ...
  package {
    content = file("./package.tar.gz")
  }
}
```

This could help, because we could provide the file content by other means, e.g. input variables

```terraform
resource "fastly_service_compute" "test" {
  ...
  package {
    content = base64decode(var.package_content)
  }
}
```

3. Allow URLs in filename. Build and upload the image to the cloud before `terraform apply` and then download the file from the specified URL

```terraform
resource "fastly_service_compute" "test" {
  ...
  package {
    filename = "https://storage.googleapis.com/xxx/package.tar.gz"
  }
}
```

Integralist:
Fixes: https://github.com/fastly/terraform-provider-fastly/issues/583

Please refer to the linked issue above for more context.

**Summary**: Customers have been requesting more flexibility in how a compute package is provided via Terraform.

## Notes

The existing package integration test is still passing:

```
make testacc TESTARGS='-run=TestAccFastlyServiceVCL_package_basic'
```

I've written a test for the new `content` attribute/behaviour and validated it's passing. 

I've also compiled the provider locally and manually validated that the behaviour works as expected.

Screenshot 2023-03-14 at 17 38 58

For the sake of my local testing, I wrote a quick Go script to read the compute package tar.gz file and to return the base64 encoding of the contents (this was to mimic a common solution used by our customers, e.g. `google_storage_bucket_object_content`)...

```go
package main

import (
	"encoding/base64"
	"fmt"
	"os"
)

func main() {
	validPackageContent, _ := os.ReadFile("pkg/testing-compute-package.tar.gz")
	b64Content := base64.StdEncoding.EncodeToString(validPackageContent)
	fmt.Printf("%s", b64Content)
}
```

I then have the following Terraform configuration which I invoke via the Terraform CLI and pass in the `-var` flag so I can set the `package_content` input variable...

```tf
terraform {
  required_providers {
    fastly = {
      source = "fastly/fastly"
      version = "4.0.0"
    }
  }
}

variable "package_content" {
  type = string
}

resource "fastly_service_compute" "testing-compute-package" {
  name = "testing-compute-package"

  domain {
    name = "testing-compute-package.integralist.co.uk"
  }

  package {
    content = var.package_content
  }

  force_destroy = true
}
```

Running `terraform show` after `terraform apply` reveals the following state...

```
# fastly_service_compute.testing-compute-package:
resource "fastly_service_compute" "testing-compute-package" {
    activate       = true
    active_version = 1
    cloned_version = 1
    comment        = "Managed by Terraform"
    force_destroy  = true
    force_refresh  = false
    id             = "W8fGlE1Ww9wLRiCYkZeCT4"
    imported       = false
    name           = "testing-compute-package"

    domain {
        name = "testing-compute-package.integralist.co.uk"
    }

    package {
        content          = "H4sIAAAJbogA/+y9C5xdV3UfvJ/...LOTS_OF_TEXT.../BwAA//8DALk/RCkAjg0A"
        source_code_hash = "659521795434ed7f4476e16ef5a85b01465167179f5c0247827c758b174798e2dac3991278ec07f1b4e581ab1909cb1fe5056480ccc56e1410c84db4898c11f1"
    }
}
```

Running `terraform plan` again shows there are no changes necessary.

If I change my compute project's code and re-run `fastly compute build` this will produce a different package and so subsequently a different base64 encoding value will be part of the next `terraform plan` which should cause an update to be issued and a new `source_code_hash` attribute to be generated...

Screenshot 2023-03-14 at 17 51 11

### DIFF
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -131,12 +131,10 @@ Optional:
 <a id="nestedblock--package"></a>
 ### Nested Schema for `package`
 
-Required:
-
-- `filename` (String) The path to the Wasm deployment package within your local filesystem
-
 Optional:
 
+- `content` (String) The contents of the Wasm deployment package (e.g. could be provided using an input variable or via external data source output variable). Conflicts with `filename`. Exactly one of these two arguments must be specified
+- `filename` (String) The path to the Wasm deployment package within your local filesystem. Conflicts with `content`. Exactly one of these two arguments must be specified
 - `source_code_hash` (String) Used to trigger updates. Must be set to a SHA512 hash of the package file specified with the filename. The usual way to set this is filesha512("package.tar.gz") (Terraform 0.11.12 and later) or filesha512(file("package.tar.gz")) (Terraform 0.11.11 and earlier), where "package.tar.gz" is the local filename of the Wasm deployment package
 
 
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -133,7 +133,7 @@ Optional:
 
 Optional:
 
-- `content` (String) The contents of the Wasm deployment package (e.g. could be provided using an input variable or via external data source output variable). Conflicts with `filename`. Exactly one of these two arguments must be specified
+- `content` (String) The contents of the Wasm deployment package as a base64 encoded string (e.g. could be provided using an input variable or via external data source output variable). Conflicts with `filename`. Exactly one of these two arguments must be specified
 - `filename` (String) The path to the Wasm deployment package within your local filesystem. Conflicts with `content`. Exactly one of these two arguments must be specified
 - `source_code_hash` (String) Used to trigger updates. Must be set to a SHA512 hash of the package file specified with the filename. The usual way to set this is filesha512("package.tar.gz") (Terraform 0.11.12 and later) or filesha512(file("package.tar.gz")) (Terraform 0.11.11 and earlier), where "package.tar.gz" is the local filename of the Wasm deployment package
 
--- a/fastly/block_fastly_service_package.go
+++ b/fastly/block_fastly_service_package.go
@@ -47,7 +47,6 @@ func (h *PackageServiceAttributeHandler) Register(s *schema.Resource) error {
 					Description:   "The path to the Wasm deployment package within your local filesystem. Conflicts with `content`. Exactly one of these two arguments must be specified",
 					ConflictsWith: []string{"package.0.content"},
 				},
-				// sha512 hash of the file
 				"source_code_hash": {
 					Type:          schema.TypeString,
 					Optional:      true,
--- a/fastly/block_fastly_service_package.go
+++ b/fastly/block_fastly_service_package.go
@@ -35,16 +35,14 @@ func (h *PackageServiceAttributeHandler) Register(s *schema.Resource) error {
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"content": {
-					Type:          schema.TypeString,
-					Optional:      true,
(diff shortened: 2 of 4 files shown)


docs/resources/service_compute.md
```diff
@@ -131,12 +131,10 @@ Optional:
 <a id="nestedblock--package"></a>
 ### Nested Schema for `package`
 
-Required:
-
-- `filename` (String) The path to the Wasm deployment package within your local filesystem
-
 Optional:
 
+- `content` (String) The contents of the Wasm deployment package as a base64 encoded string (e.g. could be provided using an input variable or via external data source output variable). Conflicts with `filename`. Exactly one of these two arguments must be specified
```
Integralist, line 136: **NOTE:** This has to be a base64 encoded string because Terraform's [`base64decode`](https://developer.hashicorp.com/terraform/language/functions/base64decode) function fails when decoding the result of reading a package, as the read content produces a `[]byte` that is not valid UTF-8 (see their documentation, which additionally states you should avoid using it on blob type data, see also https://github.com/hashicorp/terraform/issues/21669#issuecomment-500896874).

But this is fine as it fits with the tools our customers are currently using, e.g.

```tf
data "google_storage_bucket_object_content" "this" {
  name   = "component/package.tar.gz.b64"
  bucket = "my-fastly-repository"
}

// This is base64 encoded...
data.google_storage_bucket_object_content.key.content
```

fastly/block_fastly_service_package.go
```diff
@@ -34,17 +35,24 @@ func (h *PackageServiceAttributeHandler) Register(s *schema.Resource) error {
 		MinItems:    1,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
+				"content": {
+					Type:          schema.TypeString,
+					Optional:      true,
+					Description:   "The contents of the Wasm deployment package as a base64 encoded string (e.g. could be provided using an input variable or via external data source output variable). Conflicts with `filename`. Exactly one of these two arguments must be specified",
+					ConflictsWith: []string{"package.0.filename"},
```
razaj92: `ConflictsWith` would pass validation if neither `content` or `filename` are set.. maybe better to go with `ExactlyOneOf`? 

https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/schema.go#LL284C2-L284C2

fastly/block_fastly_service_package.go
```diff
@@ -72,6 +91,14 @@ func (h *PackageServiceAttributeHandler) Process(_ context.Context, d *schema.Re
 	return nil
 }
 
+type PkgType int64
+
+const (
+	PkgUndefined PkgType = iota
```
razaj92: This threw me off for a bit, completely optional suggestion:
```suggestion
	_ PkgType = iota
```

razaj92 (approved):
LGTM

PR #670, opened 2023-03-24
fix(tls/subscriptions): tls configuration id should always be passed

### LINKED ISSUE #669: fastly_tls_subscription doesn't appear to support updates with custom tls_configuration
### Terraform Version
```
v1.3.2
```

### Affected Resource(s)
- fastly_tls_subscription

```hcl
data "fastly_tls_configuration" "config" {
  name = var.fastly_tls_configuration
}

resource "fastly_tls_subscription" "tlssub" {
  domains               = [for domain in fastly_service_vcl.service[0].domain : domain.name]
  configuration_id      = data.fastly_tls_configuration.config[0].id

  force_update          = true
  force_destroy         = true

  common_name           = local.fqdn
}
```

### Expected Behavior
What should have happened?

We should be able to modify tls_subscription resources with custom tls_configurations.

### Actual Behavior
What actually happened?

400 on the update.
```
 Error: 400 - Bad Request:
│
│     Title:  Bad request
│     Detail: Must provide TLS configuration relationship
│
│   with fastly_tls_subscription.tlssub[0],
│   on fastly.tf line 153, in resource "fastly_tls_subscription" "tlssub":
│  153: resource "fastly_tls_subscription" "tlssub" {
```

Looking at the go code here, it looks like there might again be an implicit dependency on the default tls configuration option?

And it looks like the existing code to build up the update doesn't account for the configuration id correctly: [terraform-provider-fastly/fastly/resource_fastly_tls_subscription.go](https://github.com/fastly/terraform-provider-fastly/blob/e0c80d39060082271f382965d21d341e1affaeb0/fastly/resource_fastly_tls_subscription.go#L328-L331)

it only adds that piece in if there's a change:
[terraform-provider-fastly/fastly/resource_fastly_tls_subscription.go](https://github.com/fastly/terraform-provider-fastly/blob/e0c80d39060082271f382965d21d341e1affaeb0/fastly/resource_fastly_tls_subscription.go#L343-L345)

It looks like this is a PATCH so I'm surprised that's erroring out, but it does appear to be doing so.

### Steps to Reproduce
1. build a tls_subscription using a customer tls_configuration
2. attempt to update that tls_subscription without changing the tls_configuration (adding or removing a domain is what i was doing)

Integralist:
Fixes: https://github.com/fastly/terraform-provider-fastly/issues/669

**Problem:** A customer noticed when modifying their TLS Subscription domains that an error was reported saying a TLS Configuration ID was missing. This was caused by the customer not having a default TLS Configuration set, only custom configuration objects (one of which they specify the ID for in their HCL config). But because that ID wasn't changing, the current TF implementation wouldn't send it in the API request, and so the API returned an error because it expected an ID.

**Solution**: We always provide the TLS Configuration object ID. Even if the customer has a default TLS Configuration object set on their account we'll still pass whatever the computed ID was to ensure the API doesn't error.

**NOTE:** I've run the complete end-to-end integration test suite and it's passing.

### DIFF
--- a/fastly/resource_fastly_tls_subscription.go
+++ b/fastly/resource_fastly_tls_subscription.go
@@ -336,7 +336,7 @@ func resourceFastlyTLSSubscriptionUpdate(ctx context.Context, d *schema.Resource
 	// This is why we wrap the API request in the following conditional check.
 	// We then send BOTH "domains" and "common_name" in the API request.
 	// This is because they both will have a pre-existing value.
-	if d.HasChange("domains") || d.HasChange("common_name") {
+	if d.HasChanges("domains", "common_name") {
 		// NOTE: The API doesn't care if the domains are in a different order.
 		// I mention this because if it did, then we'd only want to set the Domains
 		// field on the input struct if there was a change because we otherwise
--- a/fastly/resource_fastly_tls_subscription.go
+++ b/fastly/resource_fastly_tls_subscription.go
@@ -323,32 +323,47 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 }
 
 func resourceFastlyTLSSubscriptionUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	conn := meta.(*APIClient).conn
-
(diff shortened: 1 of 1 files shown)


razaj92 (approved):
LGTM

PR #685, opened 2023-04-26
Bump go-fastly to new v8 major release to add project-id in GCS for logging

bowrocker:
Bump the `go-fastly` version to v8 in support of adding the missing `project-id` field to the GCS provider for logging.


fastly/block_fastly_service_logging_gcs.go
```diff
@@ -38,6 +38,11 @@ func (h *GCSLoggingServiceAttributeHandler) GetSchema() *schema.Schema {
 			DefaultFunc: schema.EnvDefaultFunc("FASTLY_GCS_ACCOUNT_NAME", ""),
 			Description: "The google account name used to obtain temporary credentials (default none). You may optionally provide this via an environment variable, `FASTLY_GCS_ACCOUNT_NAME`.",
 		},
+		"project_id": {
```
Integralist: This new field needs to be placed in alphabetical order (so just before the "secret_key" field). This is what the CI 'lint' job is failing on currently.

Integralist:
To fix the failing "docs" CI job you'll need to run `make generate-docs` locally and then commit/push the changes.

fastly/block_fastly_service_logging_gcs.go
bowrocker: Thank you, fixed in e2e4b46f2acc808712cb6bb700c99457294d420d.

fastly/block_fastly_service_logging_gcs.go
```diff
@@ -151,6 +156,7 @@ func (h *GCSLoggingServiceAttributeHandler) Create(_ context.Context, d *schema.
 		Name:             gofastly.String(resource["name"].(string)),
 		Path:             gofastly.String(resource["path"].(string)),
 		Period:           gofastly.Int(resource["period"].(int)),
+		ProjectID:        gofastly.String(resource["project-id"].(string)),
```
Integralist: I think this is a typo...

```suggestion
		ProjectID:        gofastly.String(resource["project_id"].(string)),
```

Integralist (changes requested):

fastly/block_fastly_service_logging_gcs.go
bowrocker: Thank you for catching that, fixed in 3e942ee27f88d4abcf4a5b16d924ff58bf0aec84.

Integralist (approved):

PR #686, opened 2023-04-28
fix(product_enablement): skip disabling image_optimizer

Integralist:
> **NOTE:** I've run the full test suite and it's passing ✅

## Problem

**Summary:** Users who accidentally end up with a `product_enablement` block in their Terraform config/state (even though they don't have entitlement to enable products) can end up disabling the ImageOptimizer product on their service if it was enabled separately by Fastly Support.

**Explanation:**
If I don't have the entitlement to enable products, then I need to go to Fastly Support to ask them to enable ImageOptimizer. 

If I later decide to add a `product_enablement` block to my Terraform (example below) because I think it's OK for me to add it and I'm not aware of any entitlement restrictions (e.g. I've added it by mistake), then Terraform will try to enable all the products (as I've configured it to):

```terraform
product_enablement {
    brotli_compression = true
    domain_inspector   = true
    image_optimizer    = true
    origin_inspector   = true
    websockets         = true
}
```

When I run `terraform apply` Terraform will return an API error that says "product cannot be self enabled".

But now, internally, Terraform's state will actually hold the following...

```terraform
product_enablement {
    brotli_compression = false
    domain_inspector   = false
    image_optimizer    = true
    origin_inspector   = false
    websockets         = false
}
```

It will have `image_optimizer = true` because when Terraform executes the `Read` method of the `product_enablement` 'block' it went to the Fastly API and discovered that IO was enabled (as per me asking Fastly Support previously to enable IO for me).

Now if I delete the `product_enablement` block from my config, Terraform will run its "Delete" method.

The delete flow will call https://developer.fastly.com/reference/api/products/enablement/#disable-product for each product, and consequently, Terraform will call the 'disable product' endpoint for the ImageOptimizer product and that would (unintuitively) successfully disable the IO product! When really you would expect the API call to fail (which currently Terraform silently ignores the API errors so that a user such as myself can allow the Delete method to complete and thus have Terraform clean-up my internal state to no longer contain a `product_enablement` block). 

> **NOTE:** The reason that particular API call doesn't error is because it's a special case. If a user is given a free trial to ImageOptimizer, when the free trial expires an internal Fastly monitoring tool will call the 'disable product' API endpoint for that user's service.

Now that's **_NOT_** what I want to have happen. I don't want to disable image_optimizer because I need it for my service, and I've now obviously discovered that I can't enable it again using Terraform, so this means I'd have to contact Fastly Support in a panic asking them to enable it again as my service is broken. The fact that Terraform disabled IO is a side-effect of me accidentally adding this `product_enablement` block in the first place and so Terraform's Delete method has to be called to clean things up.

> **NOTE:** Yes, it's possible to _manually_ strip out `product_enablement` from the Terraform state file but realistically that's a lot to ask of a customer to have to deal with.

## Solution

As far as I see it, if the Terraform "Delete" flow is actioned, then we can workaround image_optimzer being accidentally disabled by first making an API request to 'enable' IO and then if that call is successful we know that the user is entitled to 'disable' the same product.

This should mean all three known scenarios are catered for...

### 1. Entitled user

This means, for a user who has `product_enablement` in their config and who also has entitlement to self-enable/disable products, then deleting the `product_enablement` block from their Terraform config should be fine and the 'disable product' calls will all still be made.

### 2. Non-entitled user with IO enabled by Fastly Support

This means, for a user who accidentally adds the `product_enablement` block to their Terraform config but doesn't have entitlement to self-enable/disable products but they _do have_ IO enabled on their service, then they can delete `product_enablement` from their Terraform config, Terraform will run its Delete method and all the API calls to 'disable product' will be made _except for_ the call to disable IO (remember, the call to disable IO won't be called due to our check first to see if it can be enabled, and in this case the user is not entitled). Any API calls that are made to disable a product are going to fail silently anyway and is by design because we want the user to be able to successfully clean up their internal Terraform state. The intention is for this type of user to request IO be disabled via Fastly Support (as they would have done to get it enabled).

### 3. Non-entitled user with no products enabled

This means if a user accidentally adds `product_enablement` and doesn't have entitlement to self-enable/disable products and they _don't_ have IO enabled on their service either, then they can delete `product_enablement` from their Terraform config, Terraform will run its Delete method and all the API calls to 'disable product' that are made (remember, the call to disable IO won't be called due to our check first to see if it can be enabled), will fail silently and the user's internal state will be cleaned-up appropriately.

### DIFF
--- a/fastly/block_fastly_service_product_enablement.go
+++ b/fastly/block_fastly_service_product_enablement.go
@@ -99,7 +99,7 @@ func (h *ProductEnablementServiceAttributeHandler) GetSchema() *schema.Schema {
 }
 
 // Create creates the resource.
-func (h *ProductEnablementServiceAttributeHandler) Create(_ context.Context, d *schema.ResourceData, resource map[string]any, serviceVersion int, conn *gofastly.Client) error {
+func (h *ProductEnablementServiceAttributeHandler) Create(_ context.Context, d *schema.ResourceData, resource map[string]any, _ int, conn *gofastly.Client) error {
 	serviceID := d.Id()
 
 	if h.GetServiceMetadata().serviceType == ServiceTypeCompute {
@@ -176,7 +176,7 @@ func (h *ProductEnablementServiceAttributeHandler) Create(_ context.Context, d *
 }
 
 // Read refreshes the resource.
-func (h *ProductEnablementServiceAttributeHandler) Read(_ context.Context, d *schema.ResourceData, _ map[string]any, serviceVersion int, conn *gofastly.Client) error {
+func (h *ProductEnablementServiceAttributeHandler) Read(_ context.Context, d *schema.ResourceData, _ map[string]any, _ int, conn *gofastly.Client) error {
 	localState := d.Get(h.Key()).(*schema.Set).List()
 
 	if len(localState) > 0 || d.Get("imported").(bool) || d.Get("force_refresh").(bool) {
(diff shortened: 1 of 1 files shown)


razaj92 (approved):

Integralist:
Closing this PR in favour of two separate PRs:

1. https://github.com/fastly/terraform-provider-fastly/pull/687
2. https://github.com/fastly/terraform-provider-fastly/pull/689

PR #698, opened 2023-05-19
breaking(compute): fix package hash bug

### LINKED ISSUE #697: Use new files_hash package metadata property
**Problem**: https://github.com/fastly/cli/issues/743
**Solution**: https://github.com/fastly/cli/issues/743#issuecomment-1547451995

Integralist:
Fixes https://github.com/fastly/terraform-provider-fastly/issues/697 https://github.com/fastly/cli/issues/743

> **NOTE:** ✅ The complete integration test suite has been run and is passing.

## Problem

There is a long-standing issue with how Terraform reacts to the `package.tar.gz` file that the CLI produces. Effectively, hashing the package is inconsistent and causes Terraform to think the code has changed even when it hasn't.

## Solution

To resolve the issue the Package API now returns a new metadata property (`files_hash`) that calculates the hash from a sorted list of the files within the package.

This PR updates the Terraform provider to use this new property instead of the original `hashsum` metadata property and exposes a new `fastly_package_hash` data source that will generate the appropriate value for the `source_code_hash` attribute. 

## Notes

⚠️ Although the public interface has not changed, the underlying implementation changes have meant customers will no longer be able to use the previous approach of using `filesha512` to generate a hash from their package file. So we must consider this PR a breaking change that will result in a new major release.

This does require a slight change to a customer's process, which currently looks like this...

```tf
source_code_hash = filesha512("package.tar.gz")
```

Instead, we'll recommend the use of the `fastly_package_hash` data source...

```tf
data "fastly_package_hash" "example" {
  filename = "./path/to/package.tar.gz"
}

resource "fastly_service_compute" "example" {
  # ...

  package {
    filename         = "./path/to/package.tar.gz"
    source_code_hash = data.fastly_package_hash.example.hash
  }
}
```


razaj92 (approved):
LGTM! 

The recommended approach may not work in all Terraform environments where command substitution is not possible. Some users may have to use a `local_exec` and consume output from there as part of the plan/apply. However this still relies on the fastly cli being available/installed.

Future optimisation could include a `source_code_hash` data_source, to internalise all of this to the provider

Integralist:
Thanks @razaj92 👍🏻 

I like your idea of using a data_source so I'll work on doing that and updating this PR accordingly.

~~The only issue would be if the user tries to provide the package using `content` rather than `filename`. So I'll need to investigate how to handle that scenario.~~ << wasn't an issue, I've written a test that validates the implementation.

PR #707, opened 2023-06-09
feat(fastly_secretstore): implement resource and documentation

Integralist:
This PR implements a new `fastly_secretstore` resource and `fastly_secretstores` data source.

✅ The full end-to-end integration test suite has been run and is passing successfully.


fastly/resource_fastly_secretstore_test.go
```diff
@@ -0,0 +1,182 @@
...
+			ServiceVersion: service.Version.Number,
+		})
+		if err != nil {
+			return fmt.Errorf("error listing all resource links: %s", err)
+		}
+
+		found = false
+		for _, link := range links {
+			if link.Name == linkName {
+				found = true
```
mccurdyc, line 101: Why don't you `break` on this one?

mccurdyc (approved):
One non-blocking question

fastly/resource_fastly_secretstore_test.go
Integralist: Good spot, I should break there yes!

mccurdyc (approved):

fastly/resource_fastly_secretstore.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	log.Printf("[DEBUG] REFRESH: Secret Store input: %#v", input)
+
+	store, err := conn.GetSecretStore(input)
+	if err != nil {
+		log.Printf("[WARN] No Secret Store found '%s'", d.Id())
+		d.SetId("")
+		return diag.FromErr(err)
+	}
```
Integralist, line 65:
```suggestion
	if err != nil {
		if e, ok := err.(*gofastly.HTTPError); ok && e.IsNotFound() {
			log.Printf("[WARN] No Secret Store found '%s'", d.Id())
			d.SetId("")
			return nil
		}
		return diag.FromErr(err)
	}
```

mccurdyc (approved):

jfctfl:
Hi, glad to see this coming! We could really use this for our terraform deployment, are there any plans on when this will be merged and available?

Integralist:
Thanks @jfctfl 

I'm just at this moment chasing up feedback on when I'm able to merge this PR and publish it into a new release.

Unfortunately I don't have a timeline yet. Hopefully not too long.

jfctfl:
Awesome 🚀 
If you have any updates (also on the timeline), would appreciate if you could share those!

PR #711, opened 2023-06-14
feat: add `file_max_bytes` attribute to `logging_s3` resource

kellymclaughlin:


fastly/block_fastly_service_logging_s3.go
```diff
@@ -446,6 +455,7 @@ func (h *S3LoggingServiceAttributeHandler) buildCreate(s3Map any, serviceID stri
 		BucketName:                   gofastly.String(resource["bucket_name"].(string)),
 		CompressionCodec:             gofastly.String(resource["compression_codec"].(string)),
 		Domain:                       gofastly.String(resource["domain"].(string)),
+		FileMaxBytes:                 gofastly.Int(resource["file_max_bytes"].(int)),
```
Integralist, line 458: The zero value for int (`0`) will be wrapped in a pointer and subsequently, zero will be sent to the API. 

Is that OK? e.g. the API will know to ignore zero? (because logically zero doesn't make sense)

Or would it be better to check for `> 0` as we do further down on line 486?

e.g.

```go
if v, ok := resource["file_max_bytes"].(int); v > 0 {
	opts.FileMaxBytes = gofastly.Int(v)
}
```

This way if the value is zero (either because that's what the user set explicitly, or because the value wasn't provided so `0` is the zero value for the `int` type), then we'll end up sending `nil` to go-fastly and it will omit the value from the request body to the API.

kellymclaughlin: Yeah, that's a great idea. I'll make that change.

kellymclaughlin: Actually, after taking another look at the [validation code](https://github.com/fastly/bulleit/blob/develop/sink/impls/amazon/s3/s3.go#L470-L479) that the API uses for the endpoint, the zero value is ok in this case. In the docs we say:
> A value of 0 can be used to indicate there is no limit on the size of uploaded files, otherwise the minimum value is 1048576 bytes (1 MiB.)

So I think in this case it's ok as it is since we do explain the semantics of the zero value.

Integralist: Ah! nice ok. That's fine then 🙂

Integralist (approved):

PR #715, opened 2023-06-22
fix(rate_limiter): lookup new ID before actioning a deletion

rayzorinc:
## Issue
when deleting an existing rate_limiter, we encounter a "422 Unprocessable Entity" error.  

```hcl

  #fastly_service_vcl.service will be updated in-place
  ~ resource "fastly_service_vcl" "service" {
      ~ cloned_version     = 99 -> (known after apply)
        id                 = "XXXX"
        name               = "test"
        # (11 unchanged attributes hidden)



      - rate_limiter {
          - action               = "response" -> null
          - client_key           = "client.ip,req.http.User-Agent" -> null
          - feature_revision     = 1 -> null
          - http_methods         = "GET,PUT,TRACE,POST,HEAD,DELETE,PATCH,OPTIONS" -> null
          - name                 = "rate limit" -> null
          - penalty_box_duration = 2 -> null
          - ratelimiter_id       = "33Fp31Wo4lUKNLmw7g9H2s" -> null
          - rps_limit            = 100 -> null
          - window_size          = 10 -> null

          - response {
              - content      = <<-EOT
                    <html>
                    	<head>
                    		<title>Too Many Requests</title>
                    	</head>
                    	<body>
                    		<p>Too Many Requests</p>
                    	</body>
                    </html>
                EOT -> null
              - content_type = "text/html" -> null
              - status       = 429 -> null
            }
        }



        # (25 unchanged blocks hidden)
    }

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

fastly_service_vcl.service: Modifying... [id=XXX]
╷
│ Error: Plugin did not respond
│
│   with fastly_service_vcl.service,
│   on fastly_service/main.tf line 204, in resource "fastly_service_vcl" "service":
│  204: resource "fastly_service_vcl" "service" {
│
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain more details
```

This is because when we clone a service, it creates a new ratelimit ID associated with the clone.  The fix is try to find the new ratelimit id from the clone and remove it.

### DIFF
--- a/fastly/block_fastly_service_ratelimiter.go
+++ b/fastly/block_fastly_service_ratelimiter.go
@@ -252,10 +252,35 @@ func (h *RateLimiterAttributeHandler) Update(_ context.Context, d *schema.Resour
 
 // Delete deletes the resource.
 func (h *RateLimiterAttributeHandler) Delete(_ context.Context, d *schema.ResourceData, resource map[string]any, serviceVersion int, conn *gofastly.Client) error {
-	input := h.createDeleteERLInput(d.Id(), serviceVersion, resource)
+	var rateLimiterID string
+
+	// IMPORTANT: Cloning a Service will result in new Rate Limiter IDs.
+	//
+	// This means, to update a Rate Limiter we have to first get a list of all
+	// available Rate Limiters on the cloned service version, then identify the
+	// one we need by its 'name' (which the API doesn't treat as unique, so
+	// multiple Rate Limiters in theory can have the same name, but in Terraform
+	// we enforce that the name must be unique otherwise we can't safely determine
+	// the Rate Limiter ID).
+
+	erls, err := conn.ListERLs(&gofastly.ListERLsInput{
+		ServiceID:      d.Id(),
+		ServiceVersion: serviceVersion,
+	})
+	if err != nil {
+		return err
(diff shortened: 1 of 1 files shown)


fastly/block_fastly_service_ratelimiter.go
```diff
@@ -267,9 +292,9 @@ func (h *RateLimiterAttributeHandler) Delete(_ context.Context, d *schema.Resour
 	return nil
 }
 
-func (h *RateLimiterAttributeHandler) createDeleteERLInput(service string, latestVersion int, resource map[string]any) gofastly.DeleteERLInput {
+func (h *RateLimiterAttributeHandler) createDeleteERLInput(rateLimiterID, service string, latestVersion int, resource map[string]any) gofastly.DeleteERLInput {
```
Integralist: I would just suggest editing this `createDeleteERLInput` method so it only accepts `rateLimiterID` as it doesn't appear to require any other args.

Integralist (approved):
LGTM. Thank you @rayzorinc !

PR #733, opened 2023-08-07
feat(package): make `package` block optional

### LINKED ISSUE #728: package required for compute service

- `fastly_service_compute`

### Expected Behavior

In the web UI, package is not a required attribute while creating a new compute service. We are (at least currently) planning to use terraform to manage the creation/destruction of our services and resources, but utilize CI/CD to manage the packaging. As such, we would prefer NOT to be required to provide a package at creation time here.

### Actual Behavior

Package is required, so we needed to create a placeholder and push that along with service creation (as well as ensuring anyone running terraform has this file around if they need to update things). We also needed to add:
```
  lifecycle {
    ignore_changes = [
      package,
    ]
  }
```

### Steps to Reproduce

Create a tf file with a `fastly_service_compute` resource, omitting the package block, and run `terraform apply`

Integralist:
Fixes: https://github.com/fastly/terraform-provider-fastly/issues/728

✅ The full end-to-end integration test suite has been run and is passing successfully.

> **NOTE:** This _isn't_ considered a breaking change for the exposed configuration interface because existing users don't have to make any changes to their Terraform configuration (as we move from the more restrictive `required` behaviour to a more open/flexible `optional` behaviour).

### DIFF
--- a/fastly/block_fastly_service_package.go
+++ b/fastly/block_fastly_service_package.go
@@ -29,10 +29,9 @@ func NewServicePackage(sa ServiceMetadata) ServiceAttributeDefinition {
 func (h *PackageServiceAttributeHandler) Register(s *schema.Resource) error {
 	s.Schema[h.GetKey()] = &schema.Schema{
 		Type:        schema.TypeList,
-		Required:    true,
-		Description: "The `package` block supports uploading or modifying Wasm packages for use in a Fastly Compute@Edge service. See Fastly's documentation on [Compute@Edge](https://developer.fastly.com/learning/compute/)",
+		Optional:    true,
+		Description: "The `package` block supports uploading or modifying Wasm packages for use in a Fastly Compute@Edge service (if omitted, ensure `activate = false` is set on `fastly_service_compute` to avoid service validation errors). See Fastly's documentation on [Compute@Edge](https://developer.fastly.com/learning/compute/)",
 		MaxItems:    1,
-		MinItems:    1,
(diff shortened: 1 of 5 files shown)


fastly/block_fastly_service_package.go
```diff
@@ -29,8 +29,8 @@ func NewServicePackage(sa ServiceMetadata) ServiceAttributeDefinition {
 func (h *PackageServiceAttributeHandler) Register(s *schema.Resource) error {
 	s.Schema[h.GetKey()] = &schema.Schema{
 		Type:        schema.TypeList,
-		Required:    true,
-		Description: "The `package` block supports uploading or modifying Wasm packages for use in a Fastly Compute@Edge service. See Fastly's documentation on [Compute@Edge](https://developer.fastly.com/learning/compute/)",
+		Optional:    true,
+		Description: "The `package` block supports uploading or modifying Wasm packages for use in a Fastly Compute@Edge service (if omitted, ensure `activate = false` is set on `fastly_service_compute` to avoid service validation errors). See Fastly's documentation on [Compute@Edge](https://developer.fastly.com/learning/compute/)",
 		MaxItems:    1,
 		MinItems:    1,
```
razaj92: nit: This is no longer required 🙂

razaj92 (approved):

fastly/block_fastly_service_package.go
Integralist: ah nice spot 👍🏻

PR #786, opened 2023-12-07
Support for New Relic OTLP endpoint

mdenomy:
Adds support for the [New Relic OTLP](https://docs.fastly.com/en/guides/log-streaming-newrelic-otlp) logging endpoint

### DIFF
--- a/fastly/block_fastly_service_logging_newrelicotlp.go
+++ b/fastly/block_fastly_service_logging_newrelicotlp.go
@@ -32,10 +32,28 @@ func (h *NewRelicOTLPServiceAttributeHandler) Key() string {
 // GetSchema returns the resource schema.
 func (h *NewRelicOTLPServiceAttributeHandler) GetSchema() *schema.Schema {
 	blockAttributes := map[string]*schema.Schema{
+		"format": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "Apache style log formatting. Your log must produce valid JSON that New Relic OTLP can ingest.",
+		},
+		"format_version": {
+			Type:             schema.TypeInt,
+			Optional:         true,
+			Default:          2,
+			Description:      "The version of the custom logging format used for the configured endpoint. Can be either `1` or `2`. (default: `2`).",
+			ValidateDiagFunc: validateLoggingFormatVersion(),
+		},
 		"name": {
 			Type:        schema.TypeString,
 			Required:    true,
 			Description: "The unique name of the New Relic OTLP logging endpoint. It is important to note that changing this attribute will delete and recreate the resource",
+		},
+		"placement": {
+			Type:             schema.TypeString,
+			Optional:         true,
+			Description:      "Where in the generated VCL the logging call should be placed.",
+			ValidateDiagFunc: validateLoggingPlacement(),
 		},
 		"region": {
 			Type:        schema.TypeString,
@@ -43,43 +61,22 @@ func (h *NewRelicOTLPServiceAttributeHandler) GetSchema() *schema.Schema {
 			Default:     "US",
 			Description: "The region that log data will be sent to. Default: `US`",
 		},
+		"response_condition": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "The name of the condition to apply.",
+		},
 		"token": {
 			Type:        schema.TypeString,
 			Required:    true,
 			Sensitive:   true,
 			Description: "The Insert API key from the Account page of your New Relic account",
 		},
-	}
-
-	if h.GetServiceMetadata().serviceType == ServiceTypeVCL {
-		blockAttributes["format"] = &schema.Schema{
-			Type:        schema.TypeString,
-			Optional:    true,
-			Description: "Apache style log formatting. Your log must produce valid JSON that New Relic OTLP can ingest.",
-		}
-		blockAttributes["format_version"] = &schema.Schema{
-			Type:             schema.TypeInt,
-			Optional:         true,
-			Default:          2,
-			Description:      "The version of the custom logging format used for the configured endpoint. Can be either `1` or `2`. (default: `2`).",
-			ValidateDiagFunc: validateLoggingFormatVersion(),
-		}
-		blockAttributes["placement"] = &schema.Schema{
-			Type:             schema.TypeString,
-			Optional:         true,
-			Description:      "Where in the generated VCL the logging call should be placed.",
-			ValidateDiagFunc: validateLoggingPlacement(),
-		}
-		blockAttributes["response_condition"] = &schema.Schema{
-			Type:        schema.TypeString,
-			Optional:    true,
-			Description: "The name of the condition to apply.",
-		}
-		blockAttributes["url"] = &schema.Schema{
+		"url": {
 			Type:        schema.TypeString,
 			Optional:    true,
 			Description: "The optional New Relic Trace Observer URL to stream logs to for Infinite Tracing.",
-		}
+		},
 	}
 
 	return &schema.Schema{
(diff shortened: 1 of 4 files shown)


mdenomy:
Ran `make testacc TESTARGS='-run=TestAccFastlyServiceVCL_logging_newrelicotlp_basic'` locally. 

Note that this endpoint should only be used for VCL services, so did not add any support/testing for creating Compute services

fastly/block_fastly_service_logging_newrelicotlp.go
```diff
@@ -0,0 +1,264 @@
...
+		},
+		"token": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Sensitive:   true,
+			Description: "The Insert API key from the Account page of your New Relic account",
+		},
+	}
+
+	if h.GetServiceMetadata().serviceType == ServiceTypeVCL {
```
Integralist: As this is only used for a VCL service, let's remove this `if` statement and inline each of its block attributes into `blockAttributes` above.

fastly/block_fastly_service_logging_newrelicotlp.go
```diff
@@ -0,0 +1,264 @@
...
+		})
+		if err != nil {
+			return fmt.Errorf("error looking up New Relic OTLP logging endpoints for (%s), version (%v): %s", d.Id(), serviceVersion, err)
+		}
+
+		dll := flattenNewRelicOTLP(remoteState)
+
+		for _, element := range dll {
+			h.pruneVCLLoggingAttributes(element)
+		}
```
Integralist: We should remove this as the `pruneVCLLoggingAttribute()` method only works on Compute services...

https://github.com/fastly/terraform-provider-fastly/blob/68c8b8f9aca351d148dfd09fda9c46b2925903aa/fastly/service_attribute_definition.go#L95-L103

fastly/block_fastly_service_logging_newrelicotlp.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+}
+
+// Create creates the resource.
+func (h *NewRelicOTLPServiceAttributeHandler) Create(_ context.Context, d *schema.ResourceData, resource map[string]any, serviceVersion int, conn *gofastly.Client) error {
+	opts := h.buildCreate(resource, d.Id(), serviceVersion)
+
+	log.Printf("[DEBUG] Fastly New Relic OTLP logging addition opts: %#v", opts)
+
+	return createNewRelicOTLP(conn, opts)
```
Integralist: This method is only one line long so we should just inline its implementation here.

Let's also inline `h.buildCreate()` because the redirection is a cognitive flow disruption (IMHO).

fastly/block_fastly_service_logging_newrelicotlp.go
```diff
@@ -0,0 +1,264 @@
...
+	return nil
+}
+
+// Delete deletes the resource.
+func (h *NewRelicOTLPServiceAttributeHandler) Delete(_ context.Context, d *schema.ResourceData, resource map[string]any, serviceVersion int, conn *gofastly.Client) error {
+	opts := h.buildDelete(resource, d.Id(), serviceVersion)
+
+	log.Printf("[DEBUG] Fastly New Relic OTLP logging endpoint removal opts: %#v", opts)
+
+	return deleteNewRelicOTLP(conn, opts)
```
Integralist: This method is only called here so I would also suggest inlining the implementation as the extra redirection makes it harder to identify what the logic does (if it was called in multiple places I could see a justification for being a separate function, but as it stands I think it just adds more mental/cognitive overhead).

**UPDATE:** Same for `h.buildDelete()` I feel, as it's only a few lines long and the redirection for this makes it harder to reason about the code when the flow is interrupted multiple times.

fastly/block_fastly_service_logging_newrelicotlp.go
```diff
@@ -0,0 +1,264 @@
...
+		Region:         gofastly.String(resource["region"].(string)),
+		ServiceID:      serviceID,
+		ServiceVersion: serviceVersion,
+		Token:          gofastly.String(resource["token"].(string)),
+		URL:            gofastly.String(resource["url"].(string)),
+	}
+
+	// WARNING: The following fields shouldn't have an empty string passed.
+	// As it will cause the Fastly API to return an error.
+	// This is because go-fastly v7+ will not 'omitempty' due to pointer type.
```
Integralist, line 234: Is this definitely the case for this API endpoint? I just want to be sure we're not copy/pasting (aka cargo-culting) a pattern that might be relevant only for some other endpoints.

fastly/block_fastly_service_logging_newrelicotlp_test.go
```diff
@@ -0,0 +1,246 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccServiceVCLNewRelicOTLPConfig(name, domain),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckServiceExists("fastly_service_vcl.foo", &service),
+					testAccCheckFastlyServiceVCLNewRelicOTLPAttributes(&service, []*gofastly.NewRelicOTLP{&log1}, ServiceTypeVCL),
+					resource.TestCheckResourceAttr(
+						"fastly_service_vcl.foo", "name", name),
+					resource.TestCheckResourceAttr(
+						"fastly_service_vcl.foo", "logging_newrelicotlp.#", "1"),
```
Integralist: nit:

```suggestion
					resource.TestCheckResourceAttr("fastly_service_vcl.foo", "name", name),
					resource.TestCheckResourceAttr("fastly_service_vcl.foo", "logging_newrelicotlp.#", "1"),
```

fastly/block_fastly_service_logging_newrelicotlp_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+			{
+				Config: testAccServiceVCLNewRelicOTLPConfigUpdate(name, domain),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckServiceExists("fastly_service_vcl.foo", &service),
+					testAccCheckFastlyServiceVCLNewRelicOTLPAttributes(&service, []*gofastly.NewRelicOTLP{&log1AfterUpdate, &log2}, ServiceTypeVCL),
+					resource.TestCheckResourceAttr(
+						"fastly_service_vcl.foo", "name", name),
+					resource.TestCheckResourceAttr(
+						"fastly_service_vcl.foo", "logging_newrelicotlp.#", "2"),
```
Integralist: nit:

```suggestion
					resource.TestCheckResourceAttr("fastly_service_vcl.foo", "name", name),
					resource.TestCheckResourceAttr("fastly_service_vcl.foo", "logging_newrelicotlp.#", "2"),
```

fastly/block_fastly_service_logging_newrelicotlp_test.go
```diff
@@ -0,0 +1,246 @@
...
+					dl.UpdatedAt = nil
+
+					// Ignore VCL attributes for Compute and set to whatever is returned from the API.
+					if serviceType == ServiceTypeCompute {
+						dl.FormatVersion = d.FormatVersion
+						dl.Format = d.Format
+						dl.ResponseCondition = d.ResponseCondition
+						dl.Placement = d.Placement
+					}
+
```
Integralist: This looks like a copy/paste mistake as it's been mentioned that Compute isn't supported for this resource.

```suggestion
```

Integralist (changes requested):
Thanks for re-opening this PR! The CI job passes now which is great 👍🏻 

I have a few requested changes please.

fastly/block_fastly_service_logging_newrelicotlp.go
mdenomy: Yes, I believe this is true. I made a quick test program using gofastly and tried creating an endpoint with empty strings for both `ResponseCondition` and `Placement` and got 400 errors for each

```
error: 400 - Bad Request:

    Title:  Bad request
    Detail: No condition named ''
    RateLimitRemaining: 994
    RateLimitReset:     1702666800
```

```
error: 400 - Bad Request:

    Title:  Bad request
    Detail: Not allowed placement for ''
    RateLimitRemaining: 993
    RateLimitReset:     1702666800

```

Integralist (approved):
Thanks @mdenomy for this PR (and your patience waiting for a final review). All looks good to me 👍🏻

PR #794, opened 2023-12-28
Remove mTLS from state that returns 404

hvtonev:
When TLS Mutual Authentication is removed outside of terraform the following error occurs.

```terraform
╷
│ Error: 404 - Not Found:
│ 
│     Title:  Not found
│     Detail: The mutual_authentication you requested was not found
│ 
│ 
╵
```

Expected is to be removed from the state instead, similar to TLS Subscription

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,4 +1,7 @@
-on: pull_request
+on:
+  pull_request:
+  repository_dispatch:
+    types: [ok-to-test-command] # corresponds to ./ok-to-test.yml `commands:` field
 name: Pull request
 jobs:
   lint:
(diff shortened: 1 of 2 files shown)


Integralist (commented):
👋🏻 Thanks for opening this PR!

I don't think the `if` condition works as expected. Take a look at [this reduced test case](https://play.golang.com/p/Nfr8r2fnc06) and let me know your thoughts.

Thanks.

hvtonev:
> 👋🏻 Thanks for opening this PR!
> 
> I don't think the `if` condition works as expected. Take a look at [this reduced test case](https://play.golang.com/p/Nfr8r2fnc06) and let me know your thoughts.
> 
> Thanks.

Agree. If the type assertion fails `err` will become `nil`. Thanks for pointing this out

Integralist (approved):
Thanks for updating the PR @hvtonev 

This is now approved and I'll look to get it merged shortly.

Integralist:
@hvtonev would it be possible to merge the latest `main` into your fork's PR branch and push back up as I'm trying to validate a new CI workflow that validates the exposed interface but looks like this PR is stuck seeing the old CI job. Thanks!

hvtonev:
@Integralist Done

PR #795, opened 2024-01-02
ci: slash command to trigger tests for forked PRs

Integralist:


Integralist:
The `/ok-to-test` did nothing but I think that's because the new workflow (added by this PR) needs to be merged first before the slash command will trigger a repository dispatch event. For example, I did the same setup in my own public repo and it worked fine: https://github.com/Integralist/actions-testing/pull/2 the only difference was that I merged the new ok-to-test.yml workflow before opening that test PR.

PR #803, opened 2024-01-12
Make Compute Package content sensitve

noseglid:
When providing the content directly to a fastly compute service

```
resource "fastly_service_compute" "my_service" {
  // ...
  package {
    content = data.some_content_provider.content
  }
}
```

It will output the full base64 content in the terminal and other interfaces for terraform (e.g. atlantis) when planning the change.

Because packages typically range in megabytes, that's a lot of (useless) information to have displayed.

While "sensitive" might not be exactly true (it's not a password), the [documentation][schema-doc] for sensitive in the Schema states

> It should be used for password or other values which should be hidden.

This would fall in the category "other values which should be hidden".

It can be worked around locally by passing the contents to the [sensitive][sensitive] function

```
package {
  content = sensitive(data.some_content_provider.content)
}
```

[schema-doc]: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.31.0/helper/schema#Schema.Sensitive
[sensitive]: https://developer.hashicorp.com/terraform/language/functions/sensitive

### DIFF
--- a/docs/resources/service_compute.md
+++ b/docs/resources/service_compute.md
@@ -650,7 +650,7 @@ Optional:
 
 Optional:
 
-- `content` (String) The contents of the Wasm deployment package as a base64 encoded string (e.g. could be provided using an input variable or via external data source output variable). Conflicts with `filename`. Exactly one of these two arguments must be specified
+- `content` (String, Sensitive) The contents of the Wasm deployment package as a base64 encoded string (e.g. could be provided using an input variable or via external data source output variable). Conflicts with `filename`. Exactly one of these two arguments must be specified
 - `filename` (String) The path to the Wasm deployment package within your local filesystem. Conflicts with `content`. Exactly one of these two arguments must be specified
 - `source_code_hash` (String) Used to trigger updates. Must be set to a SHA512 hash of all files (in sorted order) within the package. The usual way to set this is using the fastly_package_hash data source.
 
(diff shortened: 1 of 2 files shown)


Integralist:
Hi @noseglid 

Thanks for opening this PR. We appreciate anyone who takes the time to improve the codebase.

I've discussed it internally with the team and we've decided **_not_** to accept the PR.

This is for a few reasons...

1. Although HashiCorp suggest the Sensitive attribute can be used for this purpose, it nevertheless feels like an overloaded term and semantically doesn't make sense for this particular use case. We're not trying to hide the value because it's "sensitive" but actually because it's just noisy.
2. There is a workaround (as you've mentioned) which is to wrap the value in a call to the [`sensitive()` function](https://developer.hashicorp.com/terraform/language/functions/sensitive) so users do have the choice to opt-in to this behaviour if they feel it makes sense to their use case.
3. It's fine for a user to _opt-in_ to this behaviour but if we make it the default it's impossible for them to _opt-out_.

Again, thank you for taking the time to open this PR we do appreciate it.

noseglid:
What is, in your opinion, ever a scenario where you would want this output?

Even a small package prints 4Mb of base64 data. What possible use can you have for viewing this data?

Integralist:
I don't have a use case in mind. It's not up to me to decide how people will want to consume that data, there may well be a reason that I'm not aware of or privy to. As mentioned, if we were to default to that behaviour and then someone reaches out to say they need that information (for whatever reason) then we're in a position where we can't go back without a breaking change. Considering the solution is a simple wrapper call around the value, that seems sufficient for now.

PR #824, opened 2024-03-27
fix(tls_subscription): ensure configuration_id is current value (not initial)

### LINKED ISSUE #787: Fastly TLS Subscription Read doesn't use the Current Value (Uses original)
When changing a TLS subscription to a new certificate/configuration ID, the resource will flap on successive plan/apply's as it appears the Terraform provider is returning the value associated with the original configuration.

When contacting Fastly support, they suggested that's the intention from the original API call and that in order to retrieve the current configuration ID, one should use the following API call. 

```
https://api.fastly.com/tls/domains?filter[tls_certificates.id]={TLS-ID-HERE}&include=tls_activations
```

Thus, the `fastly_tls_subscription` is inherently broken in the declarative nature of how this resource is intended to be used for those who change TLS configurations thereafter.  

### Terraform Version

```shell
Terraform v1.5.5
```

- `fastly_tls_subscription`

```hcl
resource "fastly_tls_subscription" "app" {
  domains               = "sample-endpoint.com"
  configuration_id      = "MXa6JFWQ4xZpkuRhDOzaPA"
  force_update          = true
  force_destroy         = true
}
```
image

### Expected Behavior

The TLS subscription should be set with the new configuration ID and upon successive plan/apply's not "flap" and should show successful.

### Actual Behavior

Successive plan/applys keep attempting to change it to the new version. The original GET/READ call reads from the **original** value of the TLS subscription (not the current).

```
  # fastly_tls_subscription.media will be updated in-place
  ~ resource "fastly_tls_subscription" "app" {
      ~ configuration_id        = "irMQsUYCMyttO9EPRhm2EA" -> "MXa6JFWQ4xZpkuRhDOzaPA"
        id                      = "XXXXXXXXXXXXXXXXXXXXXXX"
        # (12 unchanged attributes hidden)
    }
```

virgofx:
The TLS configuration ID returned by the API is the initial TLS configuration when first activated. Instead, this value should represent the current TLS configuration id which can be found via the following API call.

https://api.fastly.com/tls/domains?filter[tls_certificates.id]={cert-id}&include=tls_activations

- [x] Tested locally and confirmed working!
- [x] Ran through `go fmt`
- [x] Fixes: #787

### DIFF
--- a/fastly/resource_fastly_tls_subscription.go
+++ b/fastly/resource_fastly_tls_subscription.go
@@ -294,10 +294,37 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 	if err != nil {
 		return diag.FromErr(err)
 	}
+
+	// NOTE: The configuration_id should change depending on scenario.
+	// For example, prior to a subscription renewal the ID is the same.
+	// Once a subscription is renewed, we need to search for the latest ID.
+	// The following API endpoint is used to search for the latest ID.
+	// https://www.fastly.com/documentation/reference/api/tls/custom-certs/domains/#list-tls-domains
 	err = d.Set("configuration_id", subscription.Configuration.ID)
 	if err != nil {
 		return diag.FromErr(err)
 	}
+	var tlsDomains []*gofastly.TLSDomain
+	tlsDomains, err = conn.ListTLSDomains(&gofastly.ListTLSDomainsInput{
+		FilterTLSCertificateID: certificateID,
+		Include:                "tls_activations",
+		Sort:                   "tls_activations.created_at",
+	})
+	for _, tlsDomain := range tlsDomains {
+		// Activations may be empty (omitempty)
+		if tlsDomain.Activations == nil {
+			break
+		}
+		activations := len(tlsDomain.Activations)
+		if activations > 0 {
+			err = d.Set("configuration_id", tlsDomain.Activations[0].Configuration.ID)
+			if err != nil {
+				return diag.FromErr(err)
+			}
+			break
+		}
+	}
+
 	err = d.Set("created_at", subscription.CreatedAt.Format(time.RFC3339))
 	if err != nil {
 		return diag.FromErr(err)


virgofx:
ping @Integralist

virgofx:
@Integralist Any chance we could get this reviewed and hopefully merged in a new release? Thanks!

Integralist:
Thanks @virgofx for your patience.

I'm reviewing this today.

fastly/resource_fastly_tls_subscription.go
```diff
@@ -294,10 +294,35 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
...
+		Include:                "tls_activations",
+	})
+	for _, tlsDomain := range tlsDomains {
+		// Activations may be empty (omitempty)
+		if tlsDomain.Activations == nil {
+			break
+		}
+		len := len(tlsDomain.Activations)
+		if len > 0 {
+			err = d.Set("configuration_id", tlsDomain.Activations[len-1].Configuration.ID)
```
Integralist: I prefer not to shadow a builtin function (i.e. `len()`)...

```suggestion
		activations := len(tlsDomain.Activations)
		if activations > 0 {
			err = d.Set("configuration_id", tlsDomain.Activations[activations-1].Configuration.ID)
```

fastly/resource_fastly_tls_subscription.go
```diff
@@ -294,10 +294,35 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
...
 	}
+	var tlsDomains []*gofastly.TLSDomain
+	tlsDomains, err = conn.ListTLSDomains(&gofastly.ListTLSDomainsInput{
+		FilterTLSCertificateID: certificateID,
+		Include:                "tls_activations",
+	})
+	for _, tlsDomain := range tlsDomains {
+		// Activations may be empty (omitempty)
+		if tlsDomain.Activations == nil {
+			break
```
Integralist, line 316: I _think_ this is correct (i.e. the use of `break` instead of `continue`). My understanding of _why_ is: [JSON-API](https://jsonapi.org/) has required the Fastly TLS team to model their response data in such a way that meant even a single 'activation' object has to be placed inside of an array, and that we expect there to only ever be a single activation object.

fastly/resource_fastly_tls_subscription.go
```diff
@@ -294,10 +294,35 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 	if err != nil {
 		return diag.FromErr(err)
 	}
+
+	// The configuration ID returned by the API is the initial TLS configuration when first
+	// activated. Instead, this value should represent the current TLS configuration id which
+	// can be found via the following API call.
+	// https://api.fastly.com/tls/domains?filter[tls_certificates.id]={cert-id}&include=tls_activations
```
Integralist: I prefer to reference the API documentation...

```suggestion
	// https://www.fastly.com/documentation/reference/api/tls/custom-certs/domains/#list-tls-domains
```

fastly/resource_fastly_tls_subscription.go
```diff
@@ -294,10 +294,35 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
 	if err != nil {
 		return diag.FromErr(err)
 	}
+
+	// The configuration ID returned by the API is the initial TLS configuration when first
+	// activated. Instead, this value should represent the current TLS configuration id which
+	// can be found via the following API call.
```
Integralist:
```suggestion
	// NOTE: The configuration_id should change depending on scenario.
	// For example, prior to a subscription renewal the ID is the same.
	// Once a subscription is renewed, we need to search for the latest ID.
	// The following API endpoint is used to search for the latest ID.
```

Integralist (approved):
LGTM. I'm not going to merge immediately, as I would like a member of the Fastly TLS team to take a pass over my review just to sanity check some of my understanding around activation objects.

virgofx:
Thanks for the review. All suggestions look great, updating...

fastly/resource_fastly_tls_subscription.go
Integralist: @virgofx UPDATE!

I've spoken with the TLS team and they recommended sorting the `tlsDomain.Activations` by `CreatedAt` as this will ensure the `[activations-1]` will return the relevant action object.

fastly/resource_fastly_tls_subscription.go
virgofx: Yeah, that was my understanding as well. Additionally, the thought was if for some reason this isn't ever defined, this is augmenting the "configuration_id" by updating it correctly so it will fallback to original behavior to make this  backwards compatible.

virgofx:
Latest changes incorporated. Additionally, re-tested locally and confirmed the sort was specified correctly.

fastly/resource_fastly_tls_subscription.go
```diff
@@ -294,10 +294,37 @@ func resourceFastlyTLSSubscriptionRead(_ context.Context, d *schema.ResourceData
...
+		Sort:                   "tls_activations.created_at",
+	})
+	for _, tlsDomain := range tlsDomains {
+		// Activations may be empty (omitempty)
+		if tlsDomain.Activations == nil {
+			break
+		}
+		activations := len(tlsDomain.Activations)
+		if activations > 0 {
+			err = d.Set("configuration_id", tlsDomain.Activations[activations-1].Configuration.ID)
```
virgofx: Since we're sorting by created at and guarded against length, should we change this to `tlsDomain.Activations[0]` ?

Effectively, this list is always of length 1 as noted above in the comments so it's somewhat not necessary to grab te last value; however, might make more sense to pluck the first?

Integralist: ~~I made a similar suggestion and the TLS team suggested that there could be (in _very_ specific cases) be more than one entries and so it's safer to sort and grab appropriate end~~

Ignore me, I realized your sort is putting latest at the start so [0] should be fine

virgofx: Updated! 

Also, built and ran it through a couple local terraform resources that map to this PR and confirmed working.

Integralist (approved):

Integralist:
Thanks @virgofx for your work on this PR.

I'm not going to publish a new release just yet as I have a couple of other TF tasks I'm looking into at the moment. But hopefully very soon.

PR #827, opened 2024-04-08
fix: update the default timeout value for healthcheck consistency with Fastly App (UI)

### LINKED ISSUE #825: Default origin healthcheck timeout seconds is too short
### Terraform Version

```shell
Terraform v1.7.1
on darwin_arm64
```

- `fastly_service_vcl.healthcheck`

```hcl
healthcheck {
    name   = "example_healthcheck"
    host   = "example.com"
    path   = "/"
}
```

### Expected Behavior

What should have happened?

When I create `healthcheck` setting without the `timeout` field, it's set as 500ms as default.
I think 500ms is too short to respond successfully between Fastly's edge and some regional origins.

On creating healthcheck setting via the managed console, the default timeout is 5s as default.
So I propose that this field also be 5000ms or bit extend as default in terraform.

### Actual Behavior

What actually happened?

500ms is too short and some regional healthcheck may timeout and then Deliver service respond with 503 Backend Unhealthy.

### Debug Output (optional)

Please provide a link to a GitHub Gist containing the complete debug output (see [Terraform docs](https://www.terraform.io/docs/internals/debugging.html)). Ensure any sensitive data is `<REDACTED>`.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids (optional)

Is there anything atypical about your account or set up that we should know?

I'm using Fastly in Japan, and setting up a backend with Cloud Run as Asia regional.
When I accessed from Osaka Data Center and then Cloud Run will accept request in a Hongkong region, then it took over 500ms and timed out the HTTP request.

remore:
Currently timeout value for healthcheck is inconsistent with Fastly App (it's by default 5000ms in Fastly App [source](https://docs.fastly.com/img/guides/create-new-health-check.png), while with terraform it's set to 500). This PR addresses this issue. Although It's not a burning issue (ie there is a workaround for this - just by explicitly setting timeout value user can work around), given it's pretty much common to implicitly set value upon managing resources through Terraform, this could be easily a pitfall for developers who are familiar with Fastly/VCL. Also, with just a 500ms timeuout, it's possible that healthcheck request configured with this default timeout value will just fail depending on where and how this healthcheck will be done.

Fixes #825.

### DIFF
--- a/docs/resources/service_vcl.md
+++ b/docs/resources/service_vcl.md
@@ -506,7 +506,7 @@ Optional:
 - `initial` (Number) When loading a config, the initial number of probes to be seen as OK. Default `3`
 - `method` (String) Which HTTP method to use. Default `HEAD`
 - `threshold` (Number) How many Healthchecks must succeed to be considered healthy. Default `3`
-- `timeout` (Number) Timeout in milliseconds. Default `500`
+- `timeout` (Number) Timeout in milliseconds. Default `5000`
 - `window` (Number) The number of most recent Healthcheck queries to keep for this Healthcheck. Default `5`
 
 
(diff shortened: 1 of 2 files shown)


Integralist:
Thanks @remore

remore:
Thanks for merging and fixing this! 🙌 🙇

PR #828, opened 2024-04-08
fix(rate_limiter): persist uri_dictionary_name to state

Integralist:
**Problem:** For some unknown reason the TF provider was not setting `uri_dictionary_name` on the state READ operation.
**Solution:** Add the missing attribute and add a test to validate the fix.


Integralist:
✅ I've run the end-to-end test suite and all is passing

PR #829, opened 2024-04-09
fix(tls_mutual_authentication): update activation after mtls creation

Integralist:
Fix `fastly_tls_mutual_authentication`. 

There are two steps to setting up mTLS:

1. `POST  /tls/mutual_authentications`
2. `PATCH /tls/activations/tls_activation_id`

The original implementation was missing the second part.


docs/resources/tls_mutual_authentication.md
```diff
@@ -1,28 +1,124 @@
...
+  backend {
+    address = "httpbin.org"
+    name    = "httpbin"
+  }
+  force_destroy = true
+}
+
+resource "fastly_tls_subscription" "www" {
+  domains               = [for domain in fastly_service_vcl.example.domain : domain.name if domain.name == "www.${var.zone}"]
+  certificate_authority = "lets-encrypt"
```
bryangm: Would you be able to change this to "certainly"? We're trying to dogfood it at every chance we get

examples/resources/tls_mutual_authentication_basic_usage.tf
```diff
@@ -0,0 +1,84 @@
...
+  backend {
+    address = "httpbin.org"
+    name    = "httpbin"
+  }
+  force_destroy = true
+}
+
+resource "fastly_tls_subscription" "www" {
+  domains               = [for domain in fastly_service_vcl.example.domain : domain.name if domain.name == "www.${var.zone}"]
+  certificate_authority = "lets-encrypt"
```
bryangm: Same here for "certainly"

bryangm:
From a TLS perspective, the examples and basic usage all made sense. I'm unfamiliar with TF but got the gist of including the PATCH activation call within the mutual_authentication resource

fastly/resource_fastly_tls_mutual_authentication.go
```diff
@@ -218,6 +220,21 @@ func resourceFastlyTLSMutualAuthenticationUpdate(_ context.Context, d *schema.Re
 func resourceFastlyTLSMutualAuthenticationDelete(_ context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	conn := meta.(*APIClient).conn
 
+	// IMPORTANT: You can't delete mTLS with active domains.
```
bryangm, line 228: Since multiple activations/domains can be associated with a mutual_authentication, will this iterate over the list?

razaj92 (approved):

bryangm (approved):

fastly/resource_fastly_tls_mutual_authentication.go
Integralist: This is deleting the activations associated with this specific mTLS resource.

For example, if a user has the following config...

```hcl
resource "fastly_tls_mutual_authentication" "example_a" {
  activation_ids = data.fastly_tls_activation_ids.example_a.ids
  cert_bundle    = "-----BEGIN CERTIFICATE-----\n<REDACTED>\n-----END CERTIFICATE-----"
  enforced       = true
}

resource "fastly_tls_mutual_authentication" "example_b" {
  activation_ids = data.fastly_tls_activation_ids.example_b.ids
  cert_bundle    = "-----BEGIN CERTIFICATE-----\n<REDACTED>\n-----END CERTIFICATE-----"
  enforced       = true
}
```

...then we're creating two separate mTLS resources, and so when deleting this config, each resource (as part of its internal DELETE process) will call `UpdateTLSActivation()` passing an empty string for the `mutual_authentication` API field so as to unset mTLS for each activation ID being passed into the resource (via the `activation_ids` attribute).

PR #834, opened 2024-04-17
Add support for specifying location when creating KV stores

hrmsk66:
👋 This PR adds the ability to specify the location when creating a new KV store using Terraform.

Changes made:

- Update go-fastly to 9.3.0
- Add a new `location` field to the `fastly_kvstore` resource.
- Add a new acceptance test `TestAccFastlyKVStore_WithLocation_validate` and associated helper functions.
- Updated the document with the additional field.

### DIFF
--- a/fastly/resource_fastly_kvstore.go
+++ b/fastly/resource_fastly_kvstore.go
@@ -9,6 +9,7 @@ import (
 	gofastly "github.com/fastly/go-fastly/v9/fastly"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
 func resourceFastlyKVStore() *schema.Resource {
@@ -29,9 +30,12 @@ func resourceFastlyKVStore() *schema.Resource {
 			},
 			"location": {
 				Type:        schema.TypeString,
-				Default:     "",
 				Optional:    true,
 				Description: "The regional location of the KV Store. Valid values are `US`, `EU`, `ASIA`, and `AUS`.",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(
+					[]string{"US", "EU", "ASIA", "AUS"},
+					false,
+				)),
 			},
 			"name": {
 				Type:        schema.TypeString,
@@ -47,8 +51,11 @@ func resourceFastlyKVStoreCreate(_ context.Context, d *schema.ResourceData, meta
 	conn := meta.(*APIClient).conn
 
 	input := &gofastly.CreateKVStoreInput{
-		Name:     d.Get("name").(string),
-		Location: d.Get("location").(string),
+		Name: d.Get("name").(string),
+	}
+
+	if v, ok := d.GetOk("location"); ok {
+		input.Location = v.(string)
 	}
 
 	log.Printf("[DEBUG] CREATE: KV Store input: %#v", input)
--- a/fastly/resource_fastly_kvstore.go
+++ b/fastly/resource_fastly_kvstore.go
@@ -31,7 +31,7 @@ func resourceFastlyKVStore() *schema.Resource {
 				Type:        schema.TypeString,
 				Default:     "",
 				Optional:    true,
-				Description: "The regional location of the KV Store. Valid values are US, EU, ASIA, and AUS.",
+				Description: "The regional location of the KV Store. Valid values are `US`, `EU`, `ASIA`, and `AUS`.",
 			},
 			"name": {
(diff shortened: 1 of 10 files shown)


fastly/resource_fastly_kvstore.go
```diff
@@ -27,6 +27,12 @@ func resourceFastlyKVStore() *schema.Resource {
 				Optional:    true,
 				Description: "Allow the KV Store to be deleted, even if it contains entries. Defaults to false.",
 			},
+			"location": {
+				Type:        schema.TypeString,
+				Default:     "",
+				Optional:    true,
+				Description: "The regional location of the KV Store. Valid values are `US`, `EU`, `ASIA`, and `AUS`.",
+			},
```
Integralist: Let's validate the inputs using `ValidateDiagFunc` like this:
https://github.com/fastly/terraform-provider-fastly/blob/086f3dec206e4e7f8a1bd6aa030ae307901f0738/fastly/block_fastly_service_logging_s3.go#L39-L59

**EDIT:** I've just realised we didn't set up constants, so the values will need to be hard coded.

fastly/resource_fastly_kvstore.go
```diff
@@ -27,6 +27,12 @@ func resourceFastlyKVStore() *schema.Resource {
 				Optional:    true,
 				Description: "Allow the KV Store to be deleted, even if it contains entries. Defaults to false.",
 			},
+			"location": {
+				Type:        schema.TypeString,
+				Default:     "",
```
Integralist: Let's avoid setting `Default: ""` as I think Terraform will probably more correctly handle the null/nil case.

fastly/resource_fastly_kvstore.go
```diff
@@ -41,7 +47,8 @@ func resourceFastlyKVStoreCreate(_ context.Context, d *schema.ResourceData, meta
 	conn := meta.(*APIClient).conn
 
 	input := &gofastly.CreateKVStoreInput{
-		Name: d.Get("name").(string),
+		Name:     d.Get("name").(string),
+		Location: d.Get("location").(string),
```
Integralist: Let's move this to a separate check like...

```go
input := &gofastly.CreateKVStoreInput{
	Name: d.Get("name").(string),
}

if v, ok := d.GetOk("location"); ok {
	input.Location = v.(string)
}
```

fastly/resource_fastly_kvstore_test.go
```diff
@@ -56,6 +56,54 @@ func TestAccFastlyKVStore_validate(t *testing.T) {
...
+func TestAccFastlyKVStore_WithLocation_validate(t *testing.T) {
+	var (
+		kvStore gofastly.KVStore
+		service gofastly.ServiceDetail
+	)
+	kvStoreName := fmt.Sprintf("KV Store %s", acctest.RandString(10))
+	serviceName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	domainName := fmt.Sprintf("fastly-test.tf-%s.com", acctest.RandString(10))
+	linkName := fmt.Sprintf("resource_link_%s", acctest.RandString(10))
+	location := "EU"
```
Integralist, line 68: Let's use the constants from go-fastly instead of hard coding the strings.]

**EDIT:** I've just realised we didn't set up constants 😬

Integralist (commented):
Just a couple of minor comments, otherwise looks good 👍🏻

fastly/resource_fastly_kvstore_test.go
hrmsk66: @Integralist Does this approach look correct?
https://github.com/fastly/go-fastly/pull/521

Integralist: Yeah that's the general idea 👍🏻

hrmsk66: ~~Thank you for checking it. I have opened https://github.com/fastly/go-fastly/pull/521 🙏~~
~~Once the change is merged and released, I will update this PR.~~

**EDIT:** I have updated the PR based on your feedback. If you prefer to add constants to go-fastly first and then modify this PR accordingly, please feel free to hold off on merging this PR. There's no rush on my end.

fastly/resource_fastly_kvstore.go
```diff
@@ -27,6 +28,15 @@ func resourceFastlyKVStore() *schema.Resource {
 				Optional:    true,
 				Description: "Allow the KV Store to be deleted, even if it contains entries. Defaults to false.",
 			},
+			"location": {
```
Integralist, line 32:
```suggestion
			// TODO: Move values to constants inside of go-fastly.
			"location": {
```

Integralist (approved):
LGTM. I added a `TODO:` note into the code for future reference.

hrmsk66:
Thank you 🙇‍♂️

PR #845, opened 2024-05-06
feat(fastly_alerts): add percentage alerts

jsonroy-fastly:
This pull requests add tests and documentation about how to use a percentage alerts.

### DIFF
--- a/fastly/resource_fastly_alert.go
+++ b/fastly/resource_fastly_alert.go
@@ -60,7 +60,7 @@ func resourceFastlyAlert() *schema.Resource {
 						"ignore_below": {
 							Type:        schema.TypeFloat,
 							Optional:    true,
-							Description: "Floor noise that can be configured to ignore data points that are below this threshold.",
+							Description: "Threshold for the denominator value used in evaluations that calculate a rate or ratio. Usually used to filter out noise.",
 						},
 						"period": {
 							Type:        schema.TypeString,
@@ -296,25 +296,19 @@ func buildDimensions(data map[string][]string, v map[string]any) map[string][]st
 }
 
 func buildEvaluationStrategy(v map[string]any) map[string]any {
-	evaluationStrategy := map[string]any{}
-
-	if value, ok := v["type"]; ok {
-		evaluationStrategy["type"] = value.(string)
+	// Required attributes
+	m := map[string]any{
+		"type":      v["type"].(string),
+		"period":    v["period"].(string),
+		"threshold": v["threshold"].(float64),
 	}
 
-	if value, ok := v["period"]; ok {
-		evaluationStrategy["period"] = value.(string)
-	}
-
-	if value, ok := v["threshold"]; ok {
-		evaluationStrategy["threshold"] = value.(float64)
-	}
-
+	// Optional attributes
 	if value, ok := v["ignore_below"]; ok {
-		evaluationStrategy["ignore_below"] = value.(float64)
+		m["ignore_below"] = value.(float64)
 	}
 
-	return evaluationStrategy
+	return m
 }
 
 func buildStringSlice(s *schema.Set) []string {
--- a/fastly/resource_fastly_alert.go
+++ b/fastly/resource_fastly_alert.go
@@ -57,6 +57,11 @@ func resourceFastlyAlert() *schema.Resource {
 				MaxItems:    1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"ignore_below": {
+							Type:        schema.TypeFloat,
+							Optional:    true,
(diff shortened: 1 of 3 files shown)


fastly/resource_fastly_alert.go
```diff
@@ -57,6 +57,11 @@ func resourceFastlyAlert() *schema.Resource {
 				MaxItems:    1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"ignore_below": {
+							Type:        schema.TypeFloat,
+							Optional:    true,
+							Description: "Floor noise that can be configured to ignore data points that are below this threshold.",
```
Integralist: Could we rewrite this description to be more like the API docs? I personally find the API description makes more sense:
https://www.fastly.com/documentation/reference/api/observability/alerts/definition/#evaluation-strategy-data-model

> Threshold for the denominator value used in evaluations that calculate a rate or ratio. Usually used to filter out noise.

fastly/resource_fastly_alert.go
```diff
@@ -291,11 +296,25 @@ func buildDimensions(data map[string][]string, v map[string]any) map[string][]st
...
+		evaluationStrategy["type"] = value.(string)
+	}
+
+	if value, ok := v["period"]; ok {
+		evaluationStrategy["period"] = value.(string)
 	}
+
+	if value, ok := v["threshold"]; ok {
+		evaluationStrategy["threshold"] = value.(float64)
+	}
```
Integralist: The key lookups in the map are redundant here as we specify those fields as `required` in the schema.

So I'd rework this like so:

```go
// Required attributes
m := map[string]any{
	"type":      v["type"].(string),
	"period":    v["period"].(string),
	"threshold": v["threshold"].(float64),
}
// Optional attributes
if value, ok := v["ignore_below"]; ok {
	evaluationStrategy["ignore_below"] = value.(float64)
}
return m
```

fastly/resource_fastly_alert_test.go
```diff
@@ -193,6 +193,69 @@ func TestAccFastlyAlert_basic_stats_aggregate(t *testing.T) {
 	})
 }
 
+func TestAccFastlyAlert_basic_stats_aggregate_percent(t *testing.T) {
```
Integralist: We should really change all these test functions to remove the underscores...

e.g.
```suggestion
func TestAccFastlyAlertBasicStatsAggregatePercent(t *testing.T) {
```

Integralist (changes requested):
LGTM. But there are some minor requested changes...

Integralist:
/ok-to-test sha=330024a

**UPDATE:** The `validate-interface-fork` workflow (trigged by the above slash command) passed without error.

fastly/resource_fastly_alert.go
jsonroy-fastly: Done

fastly/resource_fastly_alert.go
jsonroy-fastly: Reworked as suggested

fastly/resource_fastly_alert_test.go
jsonroy-fastly: I changed four of them to removed the `_`

Integralist:
Sigh. Terraform.

So yeah it seems that Terraform has a thing where it expects one underscore between `TestWhatever_TheRestOfTheName` which is something to do with test name ambiguity.

Screenshot 2024-05-07 at 14 53 21

Integralist:
https://pkg.go.dev/github.com/bflad/tfproviderlint@v0.29.0/passes/AT003

So I think we can just add a single underscore like so `TestAccFastlyAlert_TitleCaseTheRest`

fastly/resource_fastly_alert_test.go
```diff
@@ -193,6 +193,69 @@ func TestAccFastlyAlert_basic_stats_aggregate(t *testing.T) {
...
+		Name:      gofastly.ToPointer(""),
+		ServiceID: gofastly.ToPointer(""),
+	}
+
+	createAlert := gofastly.AlertDefinition{
+		Description: "Terraform percent test",
+		Dimensions:  map[string][]string{},
+		// 25 percent increase
+		EvaluationStrategy: map[string]any{
+			"type":         "percent_increase",
```
darin-nee, line 207: Nice, I see we're adding tests for the new evaluation strategies, but the docs string for `type` doesn't call out these new options.

jsonroy-fastly: True, I added the missing ones, and the missing evaluation periods.

Integralist (approved):
LGTM. Did you have any other changes or is this ready to merge?

jsonroy-fastly:
Ready to merge!

PR #851, opened 2024-05-24
fix(alerts): make `service_id` attribute optional

jsonroy-fastly:
The PR aims to rectify the terraform definition of an alert when we configure an account aggregate alert.
The `service_id` on alerts is now optional when we configure an alerts on the `stats` source.


Integralist:
👋🏻 looks like you need to run `make generate-docs` and then commit those changes to your PR.

jsonroy-fastly:
@Integralist Turns out this branch https://github.com/fastly/terraform-provider-fastly/pull/852 was the one that I wanted to merge.

PR #852, opened 2024-05-24
fix(alerts): make service_id attribute optional

jsonroy-fastly:
The PR aims to rectify the terraform definition of an alert when we configure an account aggregate alert.
The `service_id` on alerts is now optional when we configure an alerts on the `stats` source.

### DIFF
--- a/fastly/resource_fastly_alert.go
+++ b/fastly/resource_fastly_alert.go
@@ -102,9 +102,9 @@ func resourceFastlyAlert() *schema.Resource {
 
 			"service_id": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				ForceNew:    true,
-				Description: "The service which the alert monitors.",
+				Description: "The service which the alert monitors. Optional when using `stats` as the `source`.",
 			},
 
 			"source": {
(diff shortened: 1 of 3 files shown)


fastly/resource_fastly_alert.go
```diff
@@ -102,9 +102,9 @@ func resourceFastlyAlert() *schema.Resource {
 
 			"service_id": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				ForceNew:    true,
-				Description: "The service which the alert monitors.",
+				Description: "The service which the alert monitors. Optional when using `stats` as the `source`.",
```
Integralist, line 110: We should have some code added to the provider to check if `source` is set to `domains` or `origins` and if so, and the `service_id` is not provided, then we error. Because my understanding is that `service_id` is _required_ if either of those other two values are set for `source`.

And have a test around that behaviour also.

jsonroy-fastly: Sounds good

jsonroy-fastly: And yes, you are correct.

Integralist (approved):
LGTM

PR #869, opened 2024-06-25
fix(alerts): ensure that alert creation works properly when 'ignore_below' is not specified

### LINKED ISSUE #868: Acceptance tests for Alerts do not pass
(CC @Integralist @jsonroy-fastly)

https://github.com/fastly/terraform-provider-fastly/blob/4ce2c73d65dfdf161e34f2bf8913a29134f2cc48/fastly/resource_fastly_alert.go#L65

While attempting to run the acceptance tests, I've found a couple which reliably fail. In particular the `BasicStats` tests [here](https://github.com/fastly/terraform-provider-fastly/blob/4ce2c73d65dfdf161e34f2bf8913a29134f2cc48/fastly/resource_fastly_alert_test.go#L79). When these tests run, the API returns a '400 Bad Request' indicating that body parameters were invalid.

After much debugging, I've determined that `ignore_below` is being supplied in the JSON payload with a value of `0` (zero), even though it was not specified in the test configurations. Adding various print statements shows that the test configuration, when rendered as text, does *not* include `ignore_below`, but when the configuration is passed to the function to create the resource [here](https://github.com/fastly/terraform-provider-fastly/blob/4ce2c73d65dfdf161e34f2bf8913a29134f2cc48/fastly/resource_fastly_alert.go#L123) it is present in the `ResourceData`, with a value of zero.

My belief is that the root cause here is that `ignore_below` is specified in the schema as an Optional 'float'. This means that it is not required to appear in the customer's TF configuration, but the item still has a default value since it is a primitive numeric type, and without any other default being specified the default value is zero. Since the API will not accept this value, these tests fail, and presumably any customer configurations which attempt to create alerts this way will also fail.

I'm not sure what to propose here as a solution: if TF will not *exclude* `ignore_below` from the resourceData when it was not specified in the TF configuration, we may have to put some special-case code in go-fastly to exclude it from the rendered JSON.

kpfleming:
If 'ignore_below' is not specified in the 'evaluation_strategy' for an alert, Terraform gives it a value of zero, which the API will not accept. Since zero is not a valid value in the configuration, this change causes it to be ignored when the API operations are invoked.

Closes #868.

### DIFF
--- a/fastly/resource_fastly_alert.go
+++ b/fastly/resource_fastly_alert.go
@@ -318,7 +318,9 @@ func buildEvaluationStrategy(v map[string]any) map[string]any {
 
 	// Optional attributes
 	if value, ok := v["ignore_below"]; ok {
-		m["ignore_below"] = value.(float64)
+		if value.(float64) > 0 {
+			m["ignore_below"] = value.(float64)
+		}
 	}
 
 	return m
--- a/fastly/resource_fastly_alert.go
+++ b/fastly/resource_fastly_alert.go
@@ -318,8 +318,8 @@ func buildEvaluationStrategy(v map[string]any) map[string]any {
 
 	// Optional attributes
 	if value, ok := v["ignore_below"]; ok {
-		if value.(float64) > 0 {
-			m["ignore_below"] = value.(float64)
+		if v := value.(float64); v > 0 {
+			m["ignore_below"] = v
 		}
 	}
 


fastly/resource_fastly_alert.go
```diff
@@ -318,7 +318,9 @@ func buildEvaluationStrategy(v map[string]any) map[string]any {
 
 	// Optional attributes
 	if value, ok := v["ignore_below"]; ok {
-		m["ignore_below"] = value.(float64)
+		if value.(float64) > 0 {
```
Integralist: Yeah we've had to work around these sort of 'quirks' before, for example:
https://github.com/fastly/terraform-provider-fastly/blob/ef22e2c5b2143ba5c5d48a92eaaec79f713af21a/fastly/block_fastly_service_logging_cloudfiles.go#L387-L403

Integralist: Also, rather than doing the assertion twice try this...

```go
if v := value.(float64); v > 0 {
    m["ignore_below"] = v
}
```

Issue #873, opened 2024-07-15
mutual_authentication_id in fastly_tls_activation never set if included when resource is created


landrew57:
### Terraform Version

```shell
$ terraform -v
Terraform v1.8.5
on darwin_arm64
+ provider registry.terraform.io/fastly/fastly v5.11.0
+ provider registry.terraform.io/hashicorp/tls v4.0.5
```

### Affected Fastly Terraform Resource(s)

Please list the affected resources, for example:

- `fastly_tls_activation`

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    fastly = {
      source  = "fastly/fastly"
      version = "5.11.0"
    }
    tls = {
      source  = "hashicorp/tls"
      version = "4.0.5"
    }
  }
}

provider "tls" {
}

provider "fastly" {
}

### Certificate Setup Start

resource "tls_private_key" "key" {
  algorithm = "RSA"
}

resource "tls_self_signed_cert" "this" {
  private_key_pem = tls_private_key.key.private_key_pem

  subject {
    common_name = "test.example.com"
  }

  is_ca_certificate     = true
  validity_period_hours = 360

  allowed_uses = [
    "server_auth",
  ]

  dns_names = ["test.example.com"]
}

resource "tls_self_signed_cert" "issuing" {
  private_key_pem = tls_private_key.key.private_key_pem

  subject {
    common_name = "issuing.test.example.com"
  }

  is_ca_certificate     = true
  validity_period_hours = 360

  allowed_uses = [
    "cert_signing",
    "server_auth",
  ]

  dns_names = ["issuing.test.example.com"]
}

### Certificate Setup End

### Fastly service setup start

resource "fastly_service_vcl" "test" {
  name = "example.com"

  domain {
    name = "test.example.com"
  }

  backend {
    address = "127.0.0.1"
    name    = "localhost"
  }

  force_destroy = true
}

### Fastly service setup end 

### Fastly TLS configuration start

resource "fastly_tls_private_key" "key" {
  key_pem = tls_private_key.key.private_key_pem
  name    = "example_example"
}

resource "fastly_tls_certificate" "this" {
  certificate_body = tls_self_signed_cert.this.cert_pem
  name             = "test.example.com"
  depends_on       = [fastly_tls_private_key.key]
}

resource "fastly_tls_activation" "this" {
  certificate_id           = fastly_tls_certificate.this.id
  domain                   = "test.example.com"
[...]

PR #875, opened 2024-07-19
feat(tls_activation): support mutual_authentication_id in domain activation at creation

landrew57:
- add test case for activation with mTLS

This fails on `main` because the `mutual_authentication_id` is not set on an activation when it is created
but it is set in the terraform state.

- `mutual_authentication_id` cannot be set when activation is created

`Update` the domain activation once it has been created to add the `mutual_authentication_id` to the configuration.

Related to fastly/terraform-provider-fastly/issues/870 and https://github.com/fastly/terraform-provider-fastly/issues/873


kpfleming:
This looks good so far, I'll pull it down to my machine and run the acceptance tests.

kpfleming (approved):
Acceptance test passes.

kpfleming:
Thank you for the contribution!

landrew57:
Hey @kpfleming do you know when a new release will be available with this change?

kpfleming:
We haven't scheduled one yet, but if possible I'll get one made early next week.

kpfleming:
Version 5.12.0 was just released.

PR #890, opened 2024-10-22
fix(tls_mutual_authentication): Ensure that 'enforced' property does not revert to default during changes.

kpfleming:
Ensure that 'enforced' property does not revert to default during changes.

If the 'enforced' property was already set to true in the configuration, and the name (or certificate bundle) of an MTLS authentication resource was changed, the 'enforced' property would revert to 'false'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
-TEST?=$$(go list ./... |grep -v 'vendor')
+GO_BIN ?= go ## Allows overriding go executable.
+
+TEST?=$$($(GO_BIN) list ./... |grep -v 'vendor')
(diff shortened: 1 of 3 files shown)


kpfleming:
`make testacc TESTARGS='-run=TestAccFastlyMTLS'` passes with these changes (it failed after the new test case was added but before the code was changed to always specify the `enforced` value).

kpfleming:
There are two commits in this PR, one which just makes it easier to build and test with 'named' Go binary versions. Review will be easier if the commits are reviewed separately.

philippschulte (approved):
Thank you!

PR #892, opened 2024-11-01
refactor(kvstore): improve key deletion performance

Integralist:
WIP.

I don't have capacity to see this work through to completion but hopefully it's enough to get you started.

Needs integration tests to be validated (e.g. create an actual kv store with entries and make sure the keys are deleted successfully).

### DIFF
--- a/fastly/resource_fastly_kvstore.go
+++ b/fastly/resource_fastly_kvstore.go
@@ -4,8 +4,9 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"sort"
+	"sync"
 
+	"github.com/fastly/go-fastly/v9/fastly"
 	gofastly "github.com/fastly/go-fastly/v9/fastly"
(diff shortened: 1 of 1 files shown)


kpfleming:
I had a thought this morning: rather than duplicating this functionality here and in the CLI, we could just put it into `go-fastly`.

Integralist:
Yup. Great point 👍🏻 Agreed.

Integralist:
That said, there has been concerns mentioned before about custom logic like this going into go-fastly if the intention is to have something like fastly-go be the primary API client (as we'd then have to build this behaviour into someone else's templates as part of the code generation process).

kpfleming:
Understood, although given the state of `fastly-go` and the other generated clients today, that won't be a concern for quite some time.

This could also be some leverage for us to push back on the KV Store team to just add this an actual API endpoint :-)

Integralist:
@kpfleming I'm going to close this PR for now. If it's important to you/users, then let's have a ticket created internally for it.

PR #909, opened 2024-12-19
fix(fastly_package_hash): unnecessary source_code_hash conflict

### LINKED ISSUE #908: `fastly_service_compute.package` does not update if package changes remotely and content is specified without filename

- `fastly_service_compute`

```hcl
resource "fastly_service_compute" "my_serivce" {
  name = "my-service"
  package {
    content = sensitive(module.pkg.content)
  }
}
```

### Expected Behavior

Terraform resets the package to whatever is configured (value of `module.pkg.content`) if it has been altered outside of terraform 

### Actual Behavior

Terraform doesn't identify the drift and does nothing, even if the content of `module.pkg.content` doesn't match what is currently deployed.

### Details

Using the above configuration, when the remote package is changed (e.g. through manual upload in UI), terraform doesn't detect the drift and does nothing.

There is an option `source_code_hash` which seems to invoke the behavior I want, but it [conflicts with providing `content`](https://github.com/fastly/terraform-provider-fastly/blob/main/fastly/block_fastly_service_package.go#L54). I tried building the provider locally and removing this conflict and providing `source_code_hash`. **This seems to create the behavior I'm seeking**.

```diff
diff --git a/fastly/block_fastly_service_package.go b/fastly/block_fastly_service_package.go
index 69d38480a3cc77fd070698be1cd00c5a4849d0d6..ff3baca51f394cf48afb8a7a649662bf9b20d4e3 100644
--- a/fastly/block_fastly_service_package.go
+++ b/fastly/block_fastly_service_package.go
@@ -51,7 +51,6 @@ func (h *PackageServiceAttributeHandler) Register(s *schema.Resource) error {
                                        Optional:      true,
                                        Computed:      true,
                                        Description:   "Used to trigger updates. Must be set to a SHA512 hash of all files (in sorted order) within the package. The usual way to set this is using the fastly_package_hash data source.",
-                                       ConflictsWith: []string{"package.0.content"},
                                },
                        },
                },
```

I'm not aware of why this `ConflictsWith` is there. It stands to reason that you should be able to provide the file hash, regardless of where the bytes of the package comes from.

I've created a PR for my changes if I'm not misunderstanding the use for `source_code_hash`: https://github.com/fastly/terraform-provider-fastly/pull/909

noseglid:
This removes the configuration that `source_code_hash` conflicts with
`content`. Regardless of how bytes are provided for the package, it
stands to reason it should be possible to also provide the files hash (as
computed by `fastly compute hash-files`) so the provider can correctly
detect drift.

Fixes https://github.com/fastly/terraform-provider-fastly/issues/908

### DIFF
--- a/fastly/block_fastly_service_package.go
+++ b/fastly/block_fastly_service_package.go
@@ -51,7 +51,6 @@ func (h *PackageServiceAttributeHandler) Register(s *schema.Resource) error {
 					Optional:      true,
 					Computed:      true,
 					Description:   "Used to trigger updates. Must be set to a SHA512 hash of all files (in sorted order) within the package. The usual way to set this is using the fastly_package_hash data source.",
-					ConflictsWith: []string{"package.0.content"},
 				},
 			},
 		},


Integralist:
@kpfleming 

re: the following comment in the linked issue...

> I'm not aware of why this ConflictsWith is there.

...I had a look back at the original PR https://github.com/fastly/terraform-provider-fastly/pull/661 and I'm also not sure why I added the `ConflictsWith` coupling between `source_code_hash` and `content`. I'm guessing it was just a case of presuming that if someone was providing `content`, then the hash would be calculated for them. But it looks like we neglected to validate changes from outside of Terraform at the time of that PR, so it would seem like this might be OK to remove here.

Have a look back over the original PR and let me know your thoughts.

I'll leave this PR in your hands, but I wanted to at least comment on this PR as I was responsible for adding the original modifications.

Thanks!

noseglid:
Anything I can do to help move this along?

kpfleming:
I don't believe there is anything you can do right now, unfortunately. Our development resources are stretched fairly thin at the moment, so it's taking us longer than we'd like to address issues in the provider. That situation will be changing for the better soon, but in the meantime we appreciate your patience; we will be working on this as soon as we can.

kpfleming:
Thanks for all the detail; this appears to be the correct fix to me. There is a risk that if a user specifies a hash value which does not match the content, Terraform will plan to update the content every time it is executed (because the provided hash won't match the one that the API returns for the current package), but that's clearly a user error.

PR #914, opened 2025-01-11
feat(alert): append 'Managed by Terraform' to descriptions

darin-nee:
This change will inform users of the Fastly app of alerts that are being managed in Terraform.

The managed by string is appended to the description field because it commonly also holds information such as runbook links and instructions provided in the alert notifications, so applying the defaulting behavior that is used for the `comment` field on a Fastly service isn't adequate in this case.

### DIFF
--- a/fastly/resource_fastly_alert.go
+++ b/fastly/resource_fastly_alert.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"log"
+	"strings"
 
 	gofastly "github.com/fastly/go-fastly/v9/fastly"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -11,6 +12,7 @@ import (
 )
 
 const badAlertSourceServiceIdConfig = "empty `service_id` is only supported for `stats` as a source"
+const managedByTerraform = "Managed by Terraform"
 
 func resourceFastlyAlert() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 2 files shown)


fastly/resource_fastly_alert.go
```diff
@@ -4,13 +4,15 @@ import (
...
 	"log"
+	"strings"
 
 	gofastly "github.com/fastly/go-fastly/v9/fastly"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 const badAlertSourceServiceIdConfig = "empty `service_id` is only supported for `stats` as a source"
+const managedByTerraform = "Managed by Terraform"
```
kpfleming: This should be exported here (or in another file) so that the test can use the same string.

kpfleming: This also opens the door to a much larger discussion... are alerts so special that they deserve this sort of tagging, or should we consider doing this more broadly across the provider?

kpfleming (changes requested):

fastly/resource_fastly_alert.go
darin-nee: I don't think there's anything special about alerts. Though decorating a textual field with this info broadly across the provider may need to be undertaken per resource since the Fastly app hasn't adopted a standard approach to surfacing it.

kpfleming (approved):

jsonroy-fastly (approved):

PR #915, opened 2025-01-16
Typo fix

case-fastly:


docs/resources/tls_subscription.md
```diff
@@ -22,7 +22,7 @@ The examples below demonstrate usage with AWS Route53 to configure DNS, and the
 
 **Basic usage:**
 
-The following example demonstrates how to configured two subdomains (e.g. `a.example.com`, `b.example.com`).
```
Integralist, line 25: 👋🏻 

Thanks for taking the time to correct this typo.

This file is the "generated" documentation file, and so you actually need to edit:
https://github.com/fastly/terraform-provider-fastly/blob/main/templates/resources/tls_subscription.md.tmpl#L25

Then run `make generate-docs` to generate this `.md` file.

But I guess as you've already manually edited this generated file you might as well just make the same edit to the template file and then there's no need to run that Makefile target 🙂

Integralist (changes requested):

case-fastly:
Closing this, per the comment above.

PR #899, opened 2024-11-12
build(deps): Upgrade to Go 1.22 for building provider.

kpfleming:
Also includes updates to the latest versions of terraform-plugin-sdk and terraform-plugin-docs.

PR #917, opened 2025-01-27
feat(domain-v1): support v1 functionality

Integralist:


fastly/resource_fastly_domain_v1.go
```diff
@@ -0,0 +1,122 @@
...
+			"domain_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The Domain Identifier (UUID).",
+			},
+			"fqdn": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The fully-qualified domain name for your domain. Can be created, but not updated.",
```
tundal45: Generally, FQDNs consist the trailing `.` so it will be `www.example.com.` instead of `www.example.com`. We have at least used FQDN to mean the latter, without the trailing `.`. It made me wonder if that has caused confusion in the past. I think UDM uses the same domain normalization that other parts of spotless uses so we likely strip it off even if the customer used proper FQDNs but it did made me wonder.

fastly/resource_fastly_domain_v1_test.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+func TestAccFastlyDomainV1_Basic(t *testing.T) {
+	domainName := fmt.Sprintf("fastly-test.tf-%s.com", acctest.RandString(10))
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+		},
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccCheckDomainV1Destroy,
```
tundal45, line 21: Does this automatically get triggered when the test destroys the created object?

tundal45 (approved):
One comment and one non-blocking question. Looks good!

fastly/resource_fastly_domain_v1.go
kpfleming: Yeah, in most cases we should not require the trailing '.', but be willing to accept it if is present. However, that brings up an important point for the TF provider: if the user specifies the FQDN with the trailing '.', it *must* be preserved by the API and returned exactly the way the user entered it. Otherwise the provider will think that the resource has been changed since it was created, and with `ForceNew: true` it will attempt to destroy the resource and recreate it.

In other words: the API must silently accept and return names in both formats, even if it only uses one format internally, or it must noisily reject names in the non-preferred format.

kpfleming (approved):

fastly/resource_fastly_domain_v1_test.go
Integralist: Correct.

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.35.0/helper/resource#TestCase

> // CheckDestroy is called after the resource is finally destroyed
> // to allow the tester to test that the resource is truly gone.
> CheckDestroy [TestCheckFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.35.0/helper/resource#TestCheckFunc)

fastly/resource_fastly_domain_v1.go
Integralist: This is a great point @kpfleming. Thanks.

If the user gives `www.example.com.` and the API happily accepts it but internally normalises it (e.g. strips the trailing dot or adds it if it's missing), then when we _READ_ from the API to check the Terraform state against the user's Terraform config file, then there will always be a diff shown to the user as the state file will have `www.example.com` (as an example) and not `www.example.com.` like in the user's config and it'll presume a change occurred, when in practice it isn't actually a different behaviour.

So one of the following two options has to happen, either... 

1. The TF provider documentation explicitly states what it expects (trailing dot or no dot), and that _must_ correlate to what the API actually returns (which I think is "no trailing dot" but I could be wrong).
2. Or, the user enters whatever they like (trailing dot or not dot, doesn't matter) but the API _must_ return that same value and not a normalised value where it's added or removed.

The latter of those two options I would prefer because it means there's less chance for a user to make a mistake.

@tundal45 @bryangm @gschorkopf can y'all advise. Thanks!

bryangm: @Integralist I would tend to lean to option 1 at this time.

While we only recently exposed `/domains/v1` publicly as part of the UDM project, our `/tls/subscription` and `/tls/certificates` endpoints have used Spotless's domain logic internally to manage domains that appear as SANs on certificates Fastly uses for quite some time. 

While I don't disagree that option 2 would be the _best_ option, the effort and prioritization of the work would take some time and option 1 could clear up the issue considerably sooner.

gschorkopf: On a functional note, as perhaps pointed out already: Spotless always gets rid of the trailing `.` in the `domains.fqdn` DB column. We append it in some cases when interacting with APIs that require it (GlobalSign APIs for example).

Integralist:
@philippschulte might need some help with this PR as I honestly don't know what's going on with this PR and the failing tests.

For example, `make vet` passes for me locally but in the CI we're getting...

```
golang.org/x/mod/internal/lazyregexp: cannot compile Go 1.22 code
golang.org/x/mod/semver: cannot compile Go 1.22 code

Vet found suspicious constructs. Please check the reported constructs
and fix them if necessary before submitting the code for review.
make: *** [Makefile:[7](https://github.com/fastly/terraform-provider-fastly/actions/runs/13060722351/job/36442779461?pr=917#step:6:8)0: vet] Error 1
```

Another example, `make test` passes for me locally but in the CI we're getting...

```
go  test $(go  list ./... |grep -v 'vendor') || exit 1
golang.org/x/mod/internal/lazyregexp: cannot compile Go 1.22 code
golang.org/x/mod/semver: cannot compile Go 1.22 code
?   	github.com/fastly/terraform-provider-fastly/fastly/hashcode	[no test files]
FAIL	github.com/fastly/terraform-provider-fastly/fastly [build failed]
FAIL
make: *** [Makefile:41: test] Error 1
```

Integralist:
@philippschulte I'm not entirely sure but it _looks_ like the issue with this TF PR failing... 

```
golang.org/x/mod/internal/lazyregexp: cannot compile Go 1.22 code
golang.org/x/mod/semver: cannot compile Go 1.22 code
```

...is stemming from a recent update to go-fastly (https://github.com/fastly/go-fastly/pull/580/files). 

In that go-fastly PR there appears to be an indirect dependency bump:

```diff
- golang.org/x/mod v0.14.0 // indirect
+ golang.org/x/mod v0.22.0 // indirect
```

The `0.22.0` version of `x/mod` ([link](https://cs.opensource.google/go/x/mod/+/refs/tags/v0.22.0:go.mod)) contains:

```go.mod
go 1.22.0
```

We have to go back to version [`0.20.0`](https://cs.opensource.google/go/x/mod/+/refs/tags/v0.20.0:go.mod) for it to be compatible:

```go.mod
go 1.18
```

This means, when we vendor the dependencies in TF we see in the `go.sum`:

```go.sum
# golang.org/x/mod v0.22.0
## explicit; go 1.22
golang.org/x/mod/internal/lazyregexp
golang.org/x/mod/modfile
golang.org/x/mod/module
golang.org/x/mod/semver
```

I think the latest go-fastly release is going to force us to have to bump the go toolchain version for TF up to `1.22.0`.

We can't revert that change as we've published a release ([`9.13.0`](https://github.com/fastly/go-fastly/releases/tag/v9.13.0)) since then.

I don't think that should be an issue per-se, as the TF provider is a compiled binary (much like the Fastly CLI) and so it isn't _embedded_ into user code like go-fastly is.

Thoughts?

cc @kpfleming

Integralist:
OK. I've gone ahead and updated this project to use Go `1.22`. 
This also required an update to goreleaser.

See commit for details: https://github.com/fastly/terraform-provider-fastly/pull/917/commits/76b111ce7594309959c2b6d32678e39197425a94

After making the changes in https://github.com/fastly/terraform-provider-fastly/pull/917/commits/76b111ce7594309959c2b6d32678e39197425a94 I ran the goreleaser target to validate...

```shell
$ make goreleaser GORELEASER_ARGS="--skip=validate --clean"

go  install github.com/goreleaser/goreleaser/v2@latest
go: github.com/goreleaser/goreleaser/v2@v2.6.1 requires go >= 1.23.4; switching to go1.23.5
  • skipping validate...
  • cleaning distribution directory
  • loading environment variables
  • getting and validating git state
    • git state                                      commit=14ad9f4844aece0cd277340325b1d9bedc8a39fb branch=integralist/domains-v1 current_tag=v5.15.0 previous_tag=v5.14.0 dirty=true
    • pipe skipped                                   reason=validation is disabled
  • parsing tag
  • setting defaults
  • running before hooks
    • running                                        hook=go mod tidy
  • ensuring distribution directory
  • setting up metadata
  • writing release metadata
  • loading go mod information
  • build prerequisites
  • building binaries
    • building                                       binary=dist/terraform-provider-fastly_freebsd_386_sse2/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_windows_arm64_v8.0/terraform-provider-fastly_v5.15.0.exe
    • building                                       binary=dist/terraform-provider-fastly_windows_arm_6/terraform-provider-fastly_v5.15.0.exe
    • building                                       binary=dist/terraform-provider-fastly_windows_amd64_v1/terraform-provider-fastly_v5.15.0.exe
    • building                                       binary=dist/terraform-provider-fastly_freebsd_arm64_v8.0/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_freebsd_arm_6/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_windows_386_sse2/terraform-provider-fastly_v5.15.0.exe
    • building                                       binary=dist/terraform-provider-fastly_freebsd_amd64_v1/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_linux_amd64_v1/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_linux_386_sse2/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_linux_arm_6/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_linux_arm64_v8.0/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_darwin_amd64_v1/terraform-provider-fastly_v5.15.0
    • building                                       binary=dist/terraform-provider-fastly_darwin_arm64_v8.0/terraform-provider-fastly_v5.15.0
    • took: 1m31s
  • writing artifacts metadata
  • build succeeded after 1m31s
  • thanks for using GoReleaser!
```

philippschulte (changes requested):
>I think the latest go-fastly release is going to force us to have to bump the go toolchain version for TF up to 1.22.0.

>We can't revert that change as we've published a release ([9.13.0](https://github.com/fastly/go-fastly/releases/tag/v9.13.0)) since then.

>I don't think that should be an issue per-se, as the TF provider is a compiled binary (much like the Fastly CLI) and so it isn't embedded into user code like go-fastly is.

I agree and to my knowledge Kevin wanted to bump Go to 1.22 (#899) anyway.

Please just update the `Requirements` and `Developing the Provider` sections in the README.md.

I am planing to cut a new release today. Thank you!

Integralist:
Thanks @philippschulte I've pushed a commit to update the README now 👍🏻

philippschulte (approved):
Thank you!

PR #918, opened 2025-01-30
fix: put an error message for deleting wo dictionaries

skygus:
I ran into an issue trying to delete a write-only dictionary, and it was difficult to understand what was happening. It returns an error: `Bad request: Not allowed to read contents of write_only dictionary`, which doesn't make any sense, since I'm trying to delete it.

I see there's a check for deleting a dictionary which lists the contents, but this won't work for write only dictionaries, so we should have a check to see if it's a write only dictionary first, and return an error to say you should have the `force_destroy` flag.

TBH, I didn't understand how to add a test for this, so this PR is basically informational.


kpfleming:
Thanks for the PR and the detailed analysis! We'll get someone assigned to this in the near future, and we'll take care of ensuring that a test is in place for the change.

anthony-gomez-fastly (approved):
LGTM! thank you!

kpfleming (changes requested):
This will need a changelog entry, and a test case.

anthony-gomez-fastly:
Closing, have updated branch [here](https://github.com/fastly/terraform-provider-fastly/pull/959)

PR #920, opened 2025-02-06
docs(CHANGELOG.md): add missing entry to v5.16.0

philippschulte:
I forgot to list this change in #919. The change should have been in its own PR but it has been made in #917.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ENHANCEMENTS:
 
+- Upgrade to Go 1.22 for building the provider [#917](https://github.com/fastly/terraform-provider-fastly/pull/917)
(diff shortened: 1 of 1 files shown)


kpfleming:
I'm not sure this needs to be in the changelog, since it should not result in any user-visible behavior differences. I don't have a strong opinion either way, but if we do include it then let's use something like "build: Use Golang 1.22 to build the provider" so the format is consistent.

philippschulte:
I don't have a strong opinion either. I just opened the PR because we have stated it before: https://github.com/fastly/terraform-provider-fastly/blob/main/CHANGELOG.md?plain=1#L795

Please feel free to close the PR otherwise let me know and I will make the formatting changes you have proposed. Thanks

PR #924, opened 2025-02-13
docs(index.md): update docs to reflect new fastly version

anthony-gomez-fastly:


philippschulte (approved):
This needs to be updated for every release! We need to update the release docs. Thanks!

PR #925, opened 2025-02-13
ci(.github/): update ci configs to match go-fastly and cli

anthony-gomez-fastly:
All Submissions:

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -13,13 +13,6 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: 1.22.x
-    - name: Restore cache
-      uses: actions/cache@v3
-      with:
-        path: ~/go/pkg/mod
-        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
-        restore-keys: |
-          ${{ runner.os }}-go-mod-
     - name: Fmt
       run: make fmt
     - name: Vet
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -39,15 +32,14 @@ jobs:
...
-    - name: Restore cache
-      uses: actions/cache@v3
-      with:
-        path: ~/go/pkg/mod
-        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
-        restore-keys: |
-          ${{ runner.os }}-go-mod-
     - name: Test
       run: make test
+
```
philippschulte: can you please remove the new line?

.github/workflows/pr.yml
```diff
@@ -39,15 +32,14 @@ jobs:
...
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: dangoslen/changelog-enforcer@v3
+
```
philippschulte: here as well please

.github/workflows/pr.yml
anthony-gomez-fastly: so if i do it here i'm going to do it everywhere, i was matching the rest of the file. thoughts?

philippschulte (dismissed):
Thank you!

.github/workflows/pr.yml
philippschulte: don't worry about it! Thanks

anthony-gomez-fastly: i went ahead and did it because i generally like it that way 😂

philippschulte (approved):
Looks better without the new lines. Thanks!

PR #933, opened 2025-02-25
Add support for 'staging' of draft service-versions.

kpfleming:
This adds a 'stage' attribute to both `fastly_service_vcl` and
`fastly_service_compute` resources, with a default of 'false' (no change
to existing behavior).

If 'stage' is set to 'true', then 'plan' operations which make changes
to service versions will also 'stage' those service versions, whether
they are also activated or not. The expected usage is to have 'stage =
true' and 'activate = false', so that the provider will clone when
necessary to make draft versions, apply changes to those draft
versions and stage them, but otherwise leave them alone. When 'stage =
true' an existing draft version will *not* be cloned in order to apply
additional changes to it, they will be applied to that draft version.

Additional implementation notes:

* Added a new function in `fastly_test.go` to render templates (from the
  text/template module), which can be used by acceptance tests for
  generating HCL. The acceptance tests for the staging feature use
  templates.

* Moved a function from `resource_fastly_service_vcl_test.go` to
  `fastly_test.go`, refactored it, and fixed bugs in it. This function
  is used in acceptance tests to compare the list of expected backends
  for a service to the list of actual backends for the service.

* Added `golang-set` as a new dependency, which is used in the function
  mentioned in the last bullet.

 All Submissions:

### New Feature Submissions:

* [X] Does your submission pass tests?

### Changes to Core Features:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?
* [X] Have you written new tests for your core changes, as applicable?

### User Impact

* [X] What is the user impact of this change? - it adds new functionality


fastly/base_fastly_service.go
```diff
@@ -299,6 +320,7 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta any
 	conn := meta.(*APIClient).conn
 
 	shouldActivate := d.Get("activate").(bool)
+	shouldStage := d.Get("stage").(bool)
```
philippschulte: nit pick; i would move this closer down to where it is actually used.

docs/resources/service_compute.md
```diff
@@ -116,6 +116,7 @@ $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx@2
 - `product_enablement` (Block Set, Max: 1) (see [below for nested schema](#nestedblock--product_enablement))
 - `resource_link` (Block Set) A resource link represents a link between a shared resource (such as an KV Store or Config Store) and a service version. (see [below for nested schema](#nestedblock--resource_link))
 - `reuse` (Boolean) Services that are active cannot be destroyed. If set to `true` a service Terraform intends to destroy will instead be deactivated (allowing it to be reused by importing it into another Terraform project). If `false`, attempting to destroy an active service will cause an error. Default `false`
+- `stage` (Boolean) Conditionally enables new service-versions to be staged. The apply step will create a new draft version but will not stage it if this is set to `false`. Default `false`
```
philippschulte: > The apply step will create a new draft version ....

nit pick; only if the latest version is locked (activated)

docs/resources/service_vcl.md
```diff
@@ -295,6 +295,7 @@ $ terraform import fastly_service_vcl.demo xxxxxxxxxxxxxxxxxxxx@2
 - `response_object` (Block Set) (see [below for nested schema](#nestedblock--response_object))
 - `reuse` (Boolean) Services that are active cannot be destroyed. If set to `true` a service Terraform intends to destroy will instead be deactivated (allowing it to be reused by importing it into another Terraform project). If `false`, attempting to destroy an active service will cause an error. Default `false`
 - `snippet` (Block Set) (see [below for nested schema](#nestedblock--snippet))
+- `stage` (Boolean) Conditionally enables new service-versions to be staged. The apply step will create a new draft version but will not stage it if this is set to `false`. Default `false`
```
philippschulte: Same as I mentioned in `docs/resources/service_compute.md`.

philippschulte (dismissed):
This took me a while to wrap my head around. The changes look good and make sense now. 

I think it's worth to mention somewhere in the provider docs that the following doesn't make sense:

```
resource "fastly_service_vcl" "test" {
  ...
  
  activate = true
  stage = true
  
  ...
  }
  ```

Specially because  `activate = true` is the default. Why would you activate staging and production at the same time. Since we don't check this during the `plan` I think it would be good to  to have a notice in the docs which points out that when you use `stage` to set `activate` to `false`. Maybe the `Guides` section would be the right place or even under the `stage` field description itself.

File: docs/resources/service_compute.md (hunk 2 above)
Comment (kpfleming) [reply docs/resources/service_compute.md]: This is correct as it is written; if `stage = false` then the existing behavior will be followed, which is to always create a draft version even if `cloned_version` is not locked.

The only time that the `locked` status is checked is when `stage = true`.

However... since this is causing confusion, I'll rewrite this to not incorporate multiple negative conditions.

File: docs/resources/service_vcl.md (hunk 3 above)
Comment (kpfleming) [reply docs/resources/service_vcl.md]: This is generated from the source code so will be changed along with the item above.

File: docs/resources/service_compute.md (hunk 2 above)
Comment (kpfleming) [reply docs/resources/service_compute.md]: And now that I'm writing proper docs, I see that you are correct!

The existing behavior for `activate = true` has not changed, but the behavior for `activate = false` has changed; previously `activate = false` would result in cloning even if `cloned_version` was not locked, but now that cloning will not happen unless `cloned_version` is locked. I'm adding documentation to describe this, I don't expect that any customers would be relying on the 'extra' cloning behavior, the new behavior is more logical for when `activate = false`.

**Comment by kpfleming** [comment]
A new 'Activation and Staging' section has been added to the docs for both Delivery and Compute services.

**Review by anthony-gomez-fastly — DISMISSED** [review]
LGTM

**Review by philippschulte — DISMISSED** [review]
Thank you for adding detailed docs!

File: docs/resources/service_compute.md [hunk 4]
```diff
@@ -79,7 +103,7 @@ $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx@2
 
 ### Optional
 
-- `activate` (Boolean) Conditionally prevents the Service from being activated. The apply step will continue to create a new draft version but will not activate it if this is set to `false`. Default `true`
+- `activate` (Boolean) Conditionally prevents new service-versions from being activated. The apply step will create a new draft version but will not activate it if this is set to `false`. Default `true`
```
mattcone:
```suggestion
- `activate` (Boolean) Conditionally prevents new service versions from being activated. The apply step will create a new draft version but will not activate it if this is set to `false`. Default `true`
```

docs/resources/service_compute.md
```diff
@@ -116,6 +140,7 @@ $ terraform import fastly_service_compute.demo xxxxxxxxxxxxxxxxxxxx@2
 - `product_enablement` (Block Set, Max: 1) (see [below for nested schema](#nestedblock--product_enablement))
 - `resource_link` (Block Set) A resource link represents a link between a shared resource (such as an KV Store or Config Store) and a service version. (see [below for nested schema](#nestedblock--resource_link))
 - `reuse` (Boolean) Services that are active cannot be destroyed. If set to `true` a service Terraform intends to destroy will instead be deactivated (allowing it to be reused by importing it into another Terraform project). If `false`, attempting to destroy an active service will cause an error. Default `false`
+- `stage` (Boolean) Conditionally enables new service-versions to be staged. If `set` to true, all changes made by an `apply` step will be staged, even if `apply` did not create a new draft version. Default `false`
```
mattcone:
```suggestion
- `stage` (Boolean) Conditionally enables new service versions to be staged. If `set` to true, all changes made by an `apply` step will be staged, even if `apply` did not create a new draft version. Default `false`
```

docs/resources/service_vcl.md
```diff
@@ -245,7 +269,7 @@ $ terraform import fastly_service_vcl.demo xxxxxxxxxxxxxxxxxxxx@2
 ### Optional
 
 - `acl` (Block Set) (see [below for nested schema](#nestedblock--acl))
-- `activate` (Boolean) Conditionally prevents the Service from being activated. The apply step will continue to create a new draft version but will not activate it if this is set to `false`. Default `true`
+- `activate` (Boolean) Conditionally prevents new service-versions from being activated. The apply step will create a new draft version but will not activate it if this is set to `false`. Default `true`
```
mattcone:
```suggestion
- `activate` (Boolean) Conditionally prevents new service versions from being activated. The apply step will create a new draft version but will not activate it if this is set to `false`. Default `true`
```

docs/resources/service_vcl.md
```diff
@@ -295,6 +319,7 @@ $ terraform import fastly_service_vcl.demo xxxxxxxxxxxxxxxxxxxx@2
 - `response_object` (Block Set) (see [below for nested schema](#nestedblock--response_object))
 - `reuse` (Boolean) Services that are active cannot be destroyed. If set to `true` a service Terraform intends to destroy will instead be deactivated (allowing it to be reused by importing it into another Terraform project). If `false`, attempting to destroy an active service will cause an error. Default `false`
 - `snippet` (Block Set) (see [below for nested schema](#nestedblock--snippet))
+- `stage` (Boolean) Conditionally enables new service-versions to be staged. If `set` to true, all changes made by an `apply` step will be staged, even if `apply` did not create a new draft version. Default `false`
```
mattcone:
```suggestion
- `stage` (Boolean) Conditionally enables new service versions to be staged. If `set` to true, all changes made by an `apply` step will be staged, even if `apply` did not create a new draft version. Default `false`
```

fastly/base_fastly_service.go
```diff
@@ -77,20 +77,25 @@ func resourceService(serviceDef ServiceDefinition) *schema.Resource {
...
+			}),
 			validateUniqueNames("backend"),
 			validateUniqueNames("rate_limiter"),
 			validateUniqueNames("snippet"),
 		),
 		Schema: map[string]*schema.Schema{
 			"activate": {
 				Type:        schema.TypeBool,
-				Description: "Conditionally prevents the Service from being activated. The apply step will continue to create a new draft version but will not activate it if this is set to `false`. Default `true`",
+				Description: "Conditionally prevents new service-versions from being activated. The apply step will create a new draft version but will not activate it if this is set to `false`. Default `true`",
```
mattcone:
```suggestion
				Description: "Conditionally prevents new service versions from being activated. The apply step will create a new draft version but will not activate it if this is set to `false`. Default `true`",
```

fastly/base_fastly_service.go
```diff
@@ -140,6 +147,20 @@ func resourceService(serviceDef ServiceDefinition) *schema.Resource {
 				Description:   "Services that are active cannot be destroyed. If set to `true` a service Terraform intends to destroy will instead be deactivated (allowing it to be reused by importing it into another Terraform project). If `false`, attempting to destroy an active service will cause an error. Default `false`",
 				ConflictsWith: []string{"force_destroy"},
 			},
+			"stage": {
+				Type:        schema.TypeBool,
+				Description: "Conditionally enables new service-versions to be staged. If `set` to true, all changes made by an `apply` step will be staged, even if `apply` did not create a new draft version. Default `false`",
```
mattcone:
```suggestion
				Description: "Conditionally enables new service versions to be staged. If `set` to true, all changes made by an `apply` step will be staged, even if `apply` did not create a new draft version. Default `false`",
```

mattcone (approved):
LGTM! I've left some comments regarding some minor formatting issues.

docs/resources/service_compute.md
kpfleming: This is a generated file, when the `.go` file is changed this will be fixed.

docs/resources/service_compute.md
kpfleming: This is a generated file, when the `.go` file is changed this will be fixed.

docs/resources/service_vcl.md
kpfleming: This is a generated file, when the `.go` file is changed this will be fixed.

docs/resources/service_vcl.md
kpfleming: This is a generated file, when the `.go` file is changed this will be fixed.

kpfleming:
OK, this needs final check-offs from @philippschulte and @anthony-gomez-fastly and it's ready to be merged. Thanks to everyone for the feedback!

philippschulte (approved):
Thank you!

anthony-gomez-fastly (approved):

PR #934, opened 2025-02-25
ci(.github): bring this repo in line with other fastly-devtools repos for CI

anthony-gomez-fastly:
All Submissions:


.github/workflows/pr.yml
```diff
@@ -4,6 +4,26 @@ on:
...
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - uses: dangoslen/dependabot-changelog-helper@v3
+        with:
+          activationLabels: dependencies
+          changelogPath: './CHANGELOG.md'
+          entryPrefix: 'build(deps): '
```
philippschulte, line 18: Two things:
1. Maybe we don't even need to specify the `entryPRefix` because the format in the README looks like exactly what we need: https://github.com/dangoslen/dependabot-changelog-helper?tab=readme-ov-file#entry-format
2. If we do need to specify it are we missing a `-` or is the action setting it by default?

philippschulte (commented):
Thanks for doing this!

.github/workflows/pr.yml
anthony-gomez-fastly: 1. i thought we wanted to maintain the style so this will do that like the example in https://github.com/fastly/go-fastly/pull/617/files
2. it will add it

philippschulte: Awesome, thanks!

philippschulte (approved):

PR #942, opened 2025-03-10
Always validate VCL services after applying changes.

kpfleming:
In PR #733 the validation logic was changed to only attempt validation of services if they would also be activated, in order to avoid failing validation of Compute services which do not have WASM packages attached to them. Unfortunately this breaks customer workflows for VCL services, as they no longer get feedback about validation failures until they attempt to activate the service outside of Terraform.

This patch changes the logic so that VCL services are always validated, even if they will not be activated or staged.

 All Submissions:

### New Feature Submissions:

* [ ] Does your submission pass tests?

### Changes to Core Features:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your core changes, as applicable?

### User Impact

* [ ] What is the user impact of this change?

### DIFF
--- a/fastly/base_fastly_service.go
+++ b/fastly/base_fastly_service.go
@@ -457,23 +457,18 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta any
 			}
 		}
 
-		// Only validate the service if `activate = true` or `stage = true`.
-		// This is primarily for compute services with no package defined.
-		// The user needs to set `activate = false` and `stage = false` to prevent errors.
-		// As they can't activate or stage a service without a package.
-		// There's no value showing validation errors to users in 'draft' mode.
-		mustValidate := false
-		if i := d.Get("activate"); i != nil {
-			if i.(bool) {
-				mustValidate = true
-			}
-		}
-		if i := d.Get("stage"); i != nil {
-			if i.(bool) {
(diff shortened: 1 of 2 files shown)


fastly/base_fastly_service.go
```diff
@@ -457,23 +457,18 @@ func resourceServiceUpdate(ctx context.Context, d *schema.ResourceData, meta any
...
+		// activated or staged.
+		//
+		// Compute (WASM) services should only be validated if
+		// they will be activated or staged, since the service
+		// may not have a WASM package attached while it is
+		// 'draft' and validation would fail. If the user is
+		// expecting to activate or stage the service, though,
+		// it must have a WASM package attached and pass
+		// validation.
+		if (serviceDef.GetType() == ServiceTypeVCL) || d.Get("activate").(bool) || d.Get("stage").(bool) {
```
anthony-gomez-fastly, line 471: is there a test case covering this?

kpfleming: Unfortunately there are no existing test cases for this logic. We can certainly try to craft some but given that this was a regression for existing customers I'd like to at least correct the regression and put 'create test cases' into a backlog issue.

anthony-gomez-fastly (approved):

PR #985, opened 2025-04-24
[CDTOOL-1047] Add an environment variable allowing users to override the sensitive attribute

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

### User Impact

* [x] What is the user impact of this change?
 This change allows users to override the `Sensitive` variable on terraform resources that block the output from appearing after plans and applies.

In order to have this change work, you must set `FASTLY_TF_DISPLAY_SENSITIVE_FIELDS` in your go env to `"true"`


fastly/provider.go
```diff
@@ -30,6 +30,12 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("FASTLY_API_URL", gofastly.DefaultEndpoint),
 				Description: "Fastly API URL",
 			},
+			"display_sensitive_fields": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
```
philippschulte: I think this should default to `false`.

philippschulte (dismissed):
From my local testing this is not yet working as expected. I am not able to get rid of:

```
+ logging_blobstorage {
    # At least one attribute in this block is (or was) sensitive,
    # so its contents will not be displayed.
}
```

I don't see how we override the global variable of `DisplaySensitiveFields` with the config field `display_sensitive_fields`. I don't see the link between the two of them. I think we should also have an acceptance test for this new behavior. Thanks.

fastly/provider.go
anthony-gomez-fastly: you're right, i kept flipping it based on lexical context and landed on the wrong one 😓

anthony-gomez-fastly: fixed in `81dfc032`

fastly/provider.go
```diff
@@ -96,11 +102,12 @@ func Provider() *schema.Provider {
...
 	provider.ConfigureContextFunc = func(_ context.Context, d *schema.ResourceData) (any, diag.Diagnostics) {
 		config := Config{
-			APIKey:     d.Get("api_key").(string),
-			BaseURL:    d.Get("base_url").(string),
-			NoAuth:     d.Get("no_auth").(bool),
-			ForceHTTP2: d.Get("force_http2").(bool),
-			UserAgent:  provider.UserAgent(TerraformProviderProductUserAgent, version.ProviderVersion),
+			APIKey:                 d.Get("api_key").(string),
+			BaseURL:                d.Get("base_url").(string),
+			DisplaySensitiveFields: d.Get("display_sensitive_fields").(bool),
```
anthony-gomez-fastly: @philippschulte this should be where the linkage is

anthony-gomez-fastly:
Screenshot 2025-04-29 at 4 18 58 PM

anthony-gomez-fastly:
Screenshot 2025-04-29 at 4 20 03 PM
Screenshot 2025-04-29 at 4 20 42 PM

fastly/provider.go
```diff
@@ -14,8 +15,14 @@ import (
 // any API requests made by the provider.
 const TerraformProviderProductUserAgent = "terraform-provider-fastly"
 
+// This value can be set to allow terraform output to display sensitive info
+var DisplaySensitiveFields bool
+
 // Provider returns a *schema.Provider.
 func Provider() *schema.Provider {
+
+	DisplaySensitiveFields = os.Getenv("FASTLY_TERRAFORM_PROVIDER_DISPLAY_SENSITIVE_FIELDS") == "true"
```
philippschulte: I would recommend to change this to: `FASTLY_TF_DISPLAY_SENSITIVE_FIELDS` to keep it a little bit shorter but I'll keep this up to you and Kevin.

philippschulte (approved):
Thank you!

fastly/provider.go
anthony-gomez-fastly: addressed in `e9317837`

PR #986, opened 2025-04-25
feat(configstore): Add configstore_entry resource

jmoney:
All Submissions:

### New Feature Submissions:

* [X] Does your submission pass tests? 

### Changes to Core Features:

Adding a configstore_entry resource.  This resource manages a single entry into the configstore as opposed to the configstore_entries resource which manages a batch of them.  Use case is we have single configstore where we register config from a lot of different sources and having multiple resource configstore_entries all trying to manage the configstore was conflicting state and items were being deleted randomly if configured wrong.


jmoney:
@anthony-gomez-fastly you think you have time to review this?

anthony-gomez-fastly:
Hi @jmoney I think this problem has probably been encountered by other users and I don't think we want to make this change as is due to the ripple effects to other resources. I would advise posting on the [Fastly Community Forum](https://community.fastly.com/) with examples. Thank you!

jmoney:
can you point to a specific example to how others have gotten around this issue?  My use case is across terraform workspaces I need to register a bit of config for that service.  This was simple to do in the edge dictionaries world but without having a config store per workspace to register this data per service, it makes it hard when you can only manage all the items or none of the items.

anthony-gomez-fastly:
I don't have one, that was why I suggested to post on the forum.

PR #997, opened 2025-05-13
ci(go.mod): add golangci-lint to repo and remove manual linting, also move tools into separate modfile

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:


fastly/data_source_package_hash.go
```diff
@@ -48,7 +48,7 @@ func dataSourceFastlyPackageHash() *schema.Resource {
 func dataSourceFastlyPackageHashRead(_ context.Context, d *schema.ResourceData, _ any) diag.Diagnostics {
 	log.Printf("[DEBUG] Generating Package hash")
 
-	pkg := fmt.Sprintf("temp-%s-package.tar.gz", randStringBytes(10))
+	pkg := fmt.Sprintf("temp-%s-package.tar.gz", rand.Text())
```
anthony-gomez-fastly: I think this does the same thing

fastly/data_source_package_hash.go
```diff
@@ -61,7 +61,7 @@ func dataSourceFastlyPackageHashRead(_ context.Context, d *schema.ResourceData,
 			return diag.Errorf("failed to decode base64 content: %s", err)
 		}
 
-		err = os.WriteFile(pkg, data, 0o644)
+		err = os.WriteFile(pkg, data, 0o600)
```
anthony-gomez-fastly, line 64: are these perms enough?

fastly/tls_test.go
```diff
@@ -106,7 +105,10 @@ func buildPrivateKey() (*rsa.PrivateKey, string, error) {
 
 func buildCertificate(privateKey *rsa.PrivateKey, domains ...string) (string, error) {
 	now := time.Now()
-	serialNumber := new(big.Int).SetInt64(rnd.Int63())
+	serialNumber, err := rand.Int(rand.Reader, big.NewInt(1000))
```
anthony-gomez-fastly, line 109: i think these are the correct way to fix this lint issue, it doesn't like using `math/rand`

.golangci.yml
```diff
@@ -1,25 +1,18 @@
...
+  modules-download-mode: readonly
 linters:
-  disable-all: true
   enable:
-    - bodyclose
     - durationcheck
     - errcheck
     - exhaustive
-    - exportloopref
-    - forcetypeassert
```
anthony-gomez-fastly, line 16: having this one specifically in is a nightmare because i think we have 1k+ "violations"

fastly/block_fastly_service_acl_test.go
```diff
@@ -140,8 +141,7 @@ func testAccCheckFastlyServiceVCLAttributesACL(service *gofastly.ServiceDetail,
 	}
 }
 
-// testAccAddACLEntries doesn't technically check for anything despite returning a TestCheckFunc. Instead it is used for
-// its side effect of adding an ACL Entry
```
Integralist, line 144: Looks like line 143 was accidentally deleted?

fastly/block_fastly_service_dictionary_test.go
```diff
@@ -187,8 +188,7 @@ func testAccCheckFastlyServiceVCLAttributesDictionary(service *gofastly.ServiceD
 	}
 }
 
-// testAccAddDictionaryItems doesn't technically check for anything despite returning a TestCheckFunc. Instead it is
-// used for its side effect of adding a Dictionary Item
```
Integralist, line 191: Looks like line 190 was accidentally deleted

fastly/data_source_package_hash.go
Integralist: I would think limiting read/write to the "owner" user, and no perms for "group"/"other" users is fine.

fastly/block_fastly_service_acl_test.go
anthony-gomez-fastly: fixed in `571fcd48`

fastly/block_fastly_service_dictionary_test.go
anthony-gomez-fastly: fixed in `511d66ca`

Integralist (approved):

fastly/data_source_package_hash.go
cee-dub: You will want to limit the size of it, perhaps? Like `.Text()[:10]` (the docs say how long of a string it returns).

fastly/data_source_package_hash.go
cee-dub: What are we writing here? GH phone app doesn't let me expand the context. :(

fastly/tls_test.go
cee-dub: Except this now limits the generated values to the range 0-1000 instead of the int64 space as before.

fastly/data_source_package_hash.go
anthony-gomez-fastly: added in `9d2d0faf`

fastly/tls_test.go
cee-dub: Replace 1000 with `math.MaxInt64`

anthony-gomez-fastly: fixed in `abcde7b3`

fastly/data_source_package_hash.go
anthony-gomez-fastly: making a temporary version of the provider if i read correctly

kpfleming: This is taking a Base64-encoded WASM package present in the `content` attribute of the data source block and writing it to disk so that it can be read in using the same codepath as an existing on-disk package.

kpfleming (approved):
Excellent improvements all around.

fastly/data_source_package_hash.go
cee-dub: Permissions seem fine. No real need to restrict the group or world but also doesn't hurt anything.

PR #1007, opened 2025-06-13
docs(tls/subscriptions): remove duplicate schema information

Integralist:
This is a documentation update only.

It removes a bunch of information that is essentially duplicated by the dynamically generated schema documentation.

The documentation removed was stale and didn't (for example) include updated values like `certainly` as an available Certificate Authority and so it was possible a user might see the old information and not scroll down to the proper schema data (which is dynamically generated from the actually provider code) to see there were more options available.

### DIFF
--- a/docs/resources/tls_subscription.md
+++ b/docs/resources/tls_subscription.md
@@ -283,47 +283,6 @@ resource "aws_route53_record" "subdomain" {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


tundal45 (approved):

philippschulte (approved):
Thank you!

PR #1024, opened 2025-07-02
Add Fastly Object Storage guide to documentation.

kpfleming:
Note: `docs/guides/fastly_object_storage.md` is a copy of `templates/guides/fastly_object_storage.md`, and can be ignored during the review process.

All Submissions:


rcaril (approved):
Looks solid to me!

anthony-gomez-fastly (approved):

templates/guides/fastly_object_storage.md
```diff
@@ -0,0 +1,117 @@
...
+  description = "FOS Key"
+  permission  = "read-write-admin"
+}
+
+module "fos" {
+  source = "./fos"
+  access_key_id = fastly_object_storage_access_keys.main.id
+  secret_key    = fastly_object_storage_access_keys.main.secret_key
+}
+```
```
gaylin, line 36: It feels like there's something missing here, either a missing word (which requires correcting punctuation), or some sort of followup statement to complete the second sentence (which requires text after the code block). The `fos/main.tf` code block below this states that it need to be placed and named specifically. Perhaps that was the plan for `main.tf` as well?

templates/guides/fastly_object_storage.md
```diff
@@ -0,0 +1,117 @@
...
+
+The second will need to be placed in a directory named `fos`, and
+named `fos/main.tf`:
```
gaylin: The comma here is unnecessary:

```suggestion
The second will need to be placed in a directory named `fos` and
named `fos/main.tf`:
```

templates/guides/fastly_object_storage.md
```diff
@@ -0,0 +1,117 @@
...
+  lifecycle {
+    ignore_changes = [region]
+  }
+}
+
+```
+
+Note: This example uses the `us-east` region of Fastly Object Storage;
+if you wish to use a different region, ensure that the proper region
+code is included in the `endpoints` block above.
```
gaylin: From a preferred words and punctuation standpoint, we suggest updating this as follows:

```suggestion
Note: This example uses the `us-east` region of Fastly Object Storage.
To use a different region, ensure that the proper region
code is included in the `endpoints` block above.
```

templates/guides/fastly_object_storage.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Getting Started
+
+With the example files in place, you'll need to initialize Terraform
+and obtain Fastly Object Storage credentials.
```
gaylin: This is correct but could benefit from a direct call to action for the reader to do something. Consider this as an option:

```suggestion
With the example files in place, you'll need to initialize Terraform
and obtain Fastly Object Storage credentials. To do this, run the following
commands:
```

templates/guides/fastly_object_storage.md
```diff
@@ -0,0 +1,117 @@
...
+
+```bash
+export FASTLY_API_KEY=<your Fastly API key here>
+terraform init
+terraform apply -target=fastly_object_storage_access_keys.main
+```
+
+Note: Terraform will issue a warning because the `-target` option is
+used. This usage of that option is safe.
```
gaylin: Not a blocker. We tend to use "use" instead of "usage" to sound more friendly.

```suggestion
Note: Terraform will issue a warning because the `-target` option is
used. This use of that option is safe.
```

templates/guides/fastly_object_storage.md
```diff
@@ -0,0 +1,117 @@
...
+
+This step will connect to the Fastly API using the Fastly Terraform
+provider and obtain a set of Fastly Object Storage credentials. Those
+credentials will be stored in the Terraform state files (or other
+state storage), they will not be displayed.
```
gaylin: There's a small punctuation or rewrite fix that needs to be made here to correct the run-on sentence. Here's one option:

```suggestion
This step will connect to the Fastly API using the Fastly Terraform
provider and obtain a set of Fastly Object Storage credentials. Those
credentials will be stored in the Terraform state files (or other
state storage) and will not be displayed.
```

templates/guides/fastly_object_storage.md
```diff
@@ -0,0 +1,117 @@
...
+
+This initial step is necessary because the credentials are required by
+the AWS Terraform provider, and if Terraform attempts to apply the
+entire configuration the AWS provider will report an error because the
+credentials are missing.
```
gaylin: Not a blocker, but to shorten this for clarity, you could split this into two separate sentences to make it easier to read (especially for non-English speakers).

```suggestion
This initial step is necessary because the credentials are required by
the AWS Terraform provider. If Terraform attempts to apply the
entire configuration without them, the AWS provider will report an error because the
credentials are missing.
```

templates/guides/fastly_object_storage.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Completing the Process
+
+With the credentials obtained, a normal Terraform `apply` step can be
+used to create the remaining infrastructure; in this case a bucket in
+Fastly Object Storage named `my-test-bucket-123`. Terraform will pass
+the Fastly Object Storage credentials to the AWS Terraform provider so
+that it can use them to authenticate its API interactions with the
+Fastly Object Storage system.
```
gaylin: Is "normal" the correct word to consider here or would "standard" work better? If the intent is to highlight that there can be abnormal things, then it's fine. I suspect, however, we're aiming for the sense of "Hey, nothing complicated. Just do this like normal.", in which case the following would be more correct:

```suggestion
With the credentials obtained, a standard Terraform `apply` step can be
used to create the remaining infrastructure; in this case, a bucket in
Fastly Object Storage named `my-test-bucket-123`. Terraform will pass
the Fastly Object Storage credentials to the AWS Terraform provider so
that it can use them to authenticate its API interactions with the
Fastly Object Storage system.
```

gaylin (changes requested):
@kpfleming , I've put in a few suggestions, some of which aren't blockers, and tried to explain the logic for everything I've suggested. Happy to review this again at any time.

templates/guides/fastly_object_storage.md
kpfleming: Yes, that is the case, this file must be named `main.tf` and that's standard practice for creating a Terraform configuration (it's the only required file in any configuration).

kpfleming:
Thanks for the review! All of the suggestions made sense to me and have been applied.

gaylin (approved):
LGTM! :shipit:

PR #1033, opened 2025-07-23
[CDTOOL-1108] - Support for Virtual Patches in Terraform

rcaril:
All Submissions:

### New Feature Submissions:

image

`make test` output:
image

`make testacc TESTARGS='-run=TestAccFastlyNGWAFVirtualPatch*'` output
image


fastly/resource_fastly_ngwaf_virtualpatches.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
+	gofastly "github.com/fastly/go-fastly/v11/fastly"
+	ws "github.com/fastly/go-fastly/v11/fastly/ngwaf/v1/workspaces/virtualpatches"
+)
+
+func resourceFastlyNGWAFVirtualPatches() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceFastlyNGWAFVirtualPatchUpdate,
```
anthony-gomez-fastly: I'm not super terraform fluent, but is this how we're supposed to handle objects that can't be created via the API?

rcaril: This is what @philippschulte suggested I do when I chatted with him earlier in the week. He said the Create / Delete functions still need to exist as this is what TF expects to utilize with the built in functions.

fastly/resource_fastly_ngwaf_virtualpatch_test.go
```diff
@@ -0,0 +1,162 @@
...
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "description", "Test NGWAF Workspace"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "mode", "block"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "ip_anonymization", "hashed"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "client_ip_headers.0", "X-Forwarded-For"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "client_ip_headers.1", "X-Real-IP"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "default_blocking_response_code", "429"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "attack_signal_thresholds.0.one_minute", "100"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "attack_signal_thresholds.0.ten_minutes", "500"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "attack_signal_thresholds.0.one_hour", "1000"),
+					resource.TestCheckResourceAttr("fastly_ngwaf_workspace.example", "attack_signal_thresholds.0.immediate", "true"),
```
kpfleming: None of these need to be checked here; checking that behavior is the responsibility of the `workspace` resource, not this one.

fastly/resource_fastly_ngwaf_virtualpatch_test.go
```diff
@@ -0,0 +1,162 @@
...
+    ip_anonymization            = "hashed"
+    client_ip_headers           = ["True-Client-IP"]
+    default_blocking_response_code = 406
+
+    attack_signal_thresholds {
+      one_minute  = 200
+      ten_minutes = 1000
+      one_hour    = 2000
+      immediate   = false
+    }
```
kpfleming: This 'workspace' configuration should only include required attributes and any attributes which may be required for testing the Virtual Patch resource. The rest can be left at their default values.

fastly/resource_fastly_ngwaf_virtualpatch_test.go
```diff
@@ -0,0 +1,162 @@
...
```
kpfleming: This should be identical to the `workspace` resource when it was created, there is no need to modify the workspace in this test.

fastly/resource_fastly_ngwaf_virtualpatches.go
kpfleming: I would expect there to be a `Create` function which returns an error of some type since the resource cannot be created by HCL. I don't know if there's an example of that already in our provider, though.

fastly/resource_fastly_ngwaf_virtualpatches.go
```diff
@@ -0,0 +1,157 @@
...
+		ReadContext:   resourceFastlyNGWAFVirtualPatchRead,
+		UpdateContext: resourceFastlyNGWAFVirtualPatchUpdate,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceFastlyNGWAFVirtualPatchImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"action": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "User-configured action of the virtual patch. Accepted values are [`block`, `log`]",
```
kpfleming: All of the attribute descriptions should match the ones at https://branch-apidocs-console-1533-ngwaf-functional-parity.developer.fastly.com/documentation/reference/api/ngwaf/virtual-patches/ (also in the data source).

fastly/resource_fastly_ngwaf_virtualpatches.go
```diff
@@ -0,0 +1,157 @@
...
+	conn := meta.(*APIClient).conn
+
+	i := ws.GetInput{
+		WorkspaceID:    gofastly.ToPointer(d.Get("workspace_id").(string)),
+		VirtualPatchID: gofastly.ToPointer(d.Get("virtual_patch_id").(string)),
+	}
+
+	log.Printf("[DEBUG] REFRESH: NGWAF virtual patch input: %#v", i)
+
+	virtualpatch, err := ws.Get(gofastly.NewContextForResourceID(ctx, d.Id()), conn, &i)
```
kpfleming: The ID passed to `NewContextForResourceID` must be the workspace ID, not the ID of the Virtual Patch resource.

kpfleming (changes requested):
Please also add a documentation template for the data source; you can use the one for redactions as an example.

rcaril:
Custom Create error handling: 
image

fastly/resource_fastly_ngwaf_virtualpatches.go
```diff
@@ -36,24 +36,30 @@ func resourceFastlyNGWAFVirtualPatches() *schema.Resource {
...
-				Description:      "User-submitted identifier (ID) of the workspace",
+				Description:      "The ID of the workspace.",
 				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
 			},
 		},
 	}
 }
 
+func resourceFastlyNGWAFVirtualPatchCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	return diag.Errorf("New Virtual patches cannot be created. Virtual patches must already exist in your Fastly NGWAF workspace. Use this resource to configure existing virtual patches only. Ensure the virtual_patch_id '%s' exists in workspace '%s' before applying.",
```
kpfleming:
```suggestion
	return diag.Errorf("Virtual Patches cannot be created. Virtual Patches must already exist in your Fastly Next-Gen WAF workspace. Use this resource to configure existing Virtual Patches only. Ensure the virtual_patch_id '%s' exists in workspace '%s' before applying.",
```

kpfleming: How do customers get Virtual Patches 'into' their workspaces?

fastly/resource_fastly_ngwaf_virtualpatches.go
```diff
@@ -64,7 +70,7 @@ func resourceFastlyNGWAFVirtualPatchRead(ctx context.Context, d *schema.Resource
 
 	log.Printf("[DEBUG] REFRESH: NGWAF virtual patch input: %#v", i)
 
-	virtualpatch, err := ws.Get(gofastly.NewContextForResourceID(ctx, d.Id()), conn, &i)
+	virtualpatch, err := ws.Get(gofastly.NewContextForResourceID(ctx, d.Get("workspace_id").(string)), conn, &i)
```
kpfleming, line 73: Please make the same change for the `Update` and `Delete` functions.

kpfleming (changes requested):
We still need `templates/data-sources/ngwaf_virtual_patches.md`, and some minor other changes, but otherwise this is in great shape.

fastly/resource_fastly_ngwaf_virtualpatches.go
rcaril: VPs are pre-existing, they exist in every workspace no matter what. The customer only has the option to turn one off and change the mode.

fastly/resource_fastly_ngwaf_virtualpatches.go
rcaril: ooops

fastly/resource_fastly_ngwaf_virtualpatches.go
kpfleming: OK, we might need to improve the language here then, as it implies that the customer needs to do something to get the VP to exist in the workspace. I would just remove the second sentence and leave the others, and remove 'only'.

kpfleming (approved):
Ship it!

PR #1053, opened 2025-08-01
Add default values to the attack signal threshold fields to allow for an empty block

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

Screenshot 2025-08-01 at 9 32 13 AM
Screenshot 2025-08-01 at 9 33 05 AM


fastly/resource_fastly_ngwaf_workspace.go
```diff
@@ -31,29 +31,33 @@ func resourceFastlyNGWAFWorkspace() *schema.Resource {
...
 							Description:      "The ten-minute interval threshold. Minimum 1 and maximum 10,000",
 							Optional:         true,
+							Default:          60,
 							ValidateDiagFunc: validation.ToDiagFunc(validation.IntBetween(1, 10000)),
 						},
 					},
 				},
 				MaxItems:    1,
-				Description: "Attack threshold parameters for system site alerts. Each threshold value is the number of attack signals per IP address that must be detected during the interval before the related IP address is flagged",
+				Description: "Attack threshold parameters for system site alerts. Each threshold value is the number of attack signals per IP address that must be detected during the interval before the related IP address is flagged. If no values are set then the defaults will be applied",
```
philippschulte: We have a lot of default values in the provider already: if I remember correctly each default value is stated in each fields description. Please check how we document default values inside the provider.

philippschulte (dismissed):
Please double-check how default values are documented throughout the provider. As far as I’m aware, we typically include each default value directly in the field’s description.

Also, could you add a dedicated test for this in the workspace resource? In the basic configuration, the initial test case should define an empty attack_signal_thresholds block (as we’ve done with other child resources) and then assert the default values. A second test case should explicitly define attack_signal_thresholds and verify that the updated values differ from the defaults. Thanks!

fastly/resource_fastly_ngwaf_workspace.go
anthony-gomez-fastly: I added default values to the comment on each field

anthony-gomez-fastly: in `cfa5e32201d75decf5b0551471719fab391b0322`

anthony-gomez-fastly:
per @kpfleming 
`I would not try to do this; I think you'll end up causing hard-to-find bugs in the future. Because we don't have nested attribute support today, the attack_signal_thresholds block is a TypeList, and that means Terraform won't create it from nothing (even if the attributes inside it have defaults) - it can't know how many of those you want to create. If the provider requires users to include an empty block in the HCL, that's the best we can do for now, and that should cause the defaults to be applied to all of the attributes in that block.`

kpfleming:
> Please double-check how default values are documented throughout the provider. As far as I’m aware, we typically include each default value directly in the field’s description.

The documentation guide covers this: https://github.com/fastly/terraform-provider-fastly/blob/main/DOCUMENTATION.md#schema-attributes

philippschulte (approved):
This looks fine! I would recommend to remove `testAccNGWAFWorkspaceConfigDefaults()` and just remove the initialization block for `account_signal_thresholds` from `testAccNGWAFWorkspaceConfig()` and test right there that the defaults have been applied. Then in the update test case you check that the defaults have been overwritten.

PR #1057, opened 2025-08-04
fix(product_enablement/ngwaf): Allow traffic_ramp to be set to zero

kpfleming:
Customers occasionally want to set traffic_ramp to zero but leave NGWAF enabled so that they can inspect the VCL snippets used for sending traffic to NGWAF.

 All Submissions:


anthony-gomez-fastly (approved):
LGTM

PR #1066, opened 2025-08-05
[CDTOOL-343] Add Defaults to Logging Formats in Terraform

rcaril:
All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

This PR will correct the behavior where not specifying a logging format will present a diff in the next `terraform apply` as the state file will not match the defaults set by the API.  By settings defaults in TF, this undesired behavior will be rectified. 

Source of truth for these endpoints can be found here:
https://github.com/fastly/Tango/blob/main/app/enums/logging.js 

** This solves issue: https://github.com/fastly/terraform-provider-fastly/issues/798 **


rcaril:
Sample test 1:
image

Sample test 2:
image

Sample test 3:
image

anthony-gomez-fastly (approved):

rcaril:
Working on improving interface main.tf to accept new values for the BQ test.

rcaril:
Requesting re-review of changes to main.tf interface test.

anthony-gomez-fastly (approved):

philippschulte (changes requested):
Please update all test suites to do the following:
- update flatten() to use the new default format string
- update the acceptance test suite to check for the following:
    1. only apply the required fields (remove the format string) and check that the default values have been applied
    2. the update request should update the format string with a custom string and check that the default value has been overwritten
    3. check that the resource can be imported as expected
    
Thanks!

rcaril:
@philippschulte  - I made some changes just for the BQ test, can you take a look and ensure these changes make sense before I replicate this to the rest?

fastly/block_fastly_service_logging_bigquery_test.go
```diff
@@ -134,9 +140,51 @@ func TestBigqueryloggingEnvDefaultFuncAttributes(t *testing.T) {
 	if result2 != secretKey {
 		t.Fatalf("Error matching:\nexpected: %#v\ngot: %#v", secretKey, result2)
 	}
+
+	result3, err3 := loggingResourceSchema["format"].DefaultFunc()
+	if err3 != nil {
```
philippschulte: Please get rid of this error handling with 1,2,3! Please declare an error var right before it is used and then utilize it but replacing `:=` with just `=".

fastly/block_fastly_service_logging_bigquery_test.go
```diff
@@ -15,6 +15,12 @@ import (
 	gofastly "github.com/fastly/go-fastly/v11/fastly"
 )
 
+// Specifies the default logging format for BigQuery from constants.go.
+var expectedFormat = LoggingBigQueryDefaultFormat
```
philippschulte: You already have a variable so why creating another one? Please just use the constant. You also don't need any comments for those vars since they are self explanatory and only local.

fastly/block_fastly_service_logging_bigquery_test.go
```diff
@@ -134,9 +140,51 @@ func TestBigqueryloggingEnvDefaultFuncAttributes(t *testing.T) {
 	if result2 != secretKey {
 		t.Fatalf("Error matching:\nexpected: %#v\ngot: %#v", secretKey, result2)
 	}
+
+	result3, err3 := loggingResourceSchema["format"].DefaultFunc()
+	if err3 != nil {
+		t.Fatalf("Unexpected err %#v when calling format DefaultFunc", result3)
```
philippschulte: I think you need to pass `err` in this call.

fastly/block_fastly_service_logging_bigquery_test.go
```diff
@@ -163,11 +211,12 @@ func testAccCheckFastlyServiceVCLAttributesBQ(service *gofastly.ServiceDetail, n
 			return fmt.Errorf("bigQuery logging endpoint user/email mismatch, expected: %s, got: %#v", email, gofastly.ToValue(bqList[0].User))
 		}
 
+		// Compute does not support the format attribute, so we won't include that here.
```
philippschulte: I didn't know that! So for compute service you can only use the default format? Please make sure that this applies to all logging providers in compute. It would also be great if you find that fact in the docs so that we can reference it here via a comment. Thanks

fastly/block_fastly_service_logging_bigquery_test.go
```diff
@@ -163,11 +211,12 @@ func testAccCheckFastlyServiceVCLAttributesBQ(service *gofastly.ServiceDetail, n
...
 		return nil
 	}
 }
 
-func testAccServiceVCLConfigBigQuery(name, gcsName, secretKey, email string) string {
+func testAccServiceVCLConfigBigQuery(name, gcsName, secretKey, email, expectedFormat string) string {
```
philippschulte: This is not doing anything without another format specifier in that format string and also adding it at the of the `fmt.Sprintf()` call:

```
func foo(bar string) {
    fmt.Printf("%s", "empty")
}
```

No use of `bar` is the same as `expectedFormat` in `testAccServiceVCLConfigBigQuery()`.

Just remove the `expectedFormat` param from the `testAccServiceVCLConfigBigQuery() with no `default` as you did. What you need on top of that is another function:

```
func testAccServiceVCLConfigBigQueryUpdate(name, gcsName, secretKey, email, format string) string {
	backendName := fmt.Sprintf("%s.aws.amazon.com", acctest.RandString(3))
	domainName := fmt.Sprintf("fastly-test.tf-%s.com", acctest.RandString(10))
	return fmt.Sprintf(`
resource "fastly_service_vcl" "foo" {
  name = "%s"

  domain {
    name    = "%s"
    comment = "tf-testing-domain"
	}

  backend {
    address = "%s"
    name    = "tf -test backend"
  }

  logging_bigquery {
    name       = "%s"
    secret_key = trimspace(%q)
    email      = "%s"
    project_id = "example-gcp-project"
    dataset    = "example_bq_dataset"
    table      = "example_bq_table"
    format = "%s"
    format     = "%%h %%l %%u %%t %%r %%>s"
    processing_region = "us"
  }

  force_destroy = true
}`, name, domainName, backendName, gcsName, secretKey, email, format)
}
```

and then use that new config here:

```
			Config: testAccServiceVCLConfigBigQueryUpdate(name, bqName, secretKey, emailUpdate, formatUpdate),
					testAccCheckServiceExists("fastly_service_vcl.foo", &service),
					testAccCheckFastlyServiceVCLAttributesBQ(&service, name, bqName, emailUpdate, formatUpdate),
				),
			},
```

fastly/block_fastly_service_logging_bigquery_test.go
```diff
@@ -191,7 +240,6 @@ resource "fastly_service_vcl" "foo" {
     project_id = "example-gcp-project"
     dataset    = "example_bq_dataset"
     table      = "example_bq_table"
-    format     = "%%h %%l %%u %%t %%r %%>s"
```
philippschulte, line 194: That is what we want :+1:

philippschulte (dismissed):
You're getting there! Let me know if you would like to pair or if you have any questions. Thanks for doing this!

fastly/block_fastly_service_logging_bigquery_test.go
rcaril: This is old code, the 1/2 syntax was already there. I replaced result3 with the following code:
```
	formatSchema := loggingResourceSchema["format"]
	if formatSchema == nil {
		t.Fatalf("Expected format field to exist in schema")
	}
	if formatSchema.Default != expectedFormat {
		t.Fatalf("Error matching format default:\nexpected: %#v\ngot: %#v", expectedFormat, formatSchema.Default)
	}```

File: fastly/block_fastly_service_logging_bigquery_test.go (hunk 3 above)
Comment (rcaril) [reply fastly/block_fastly_service_logging_bigquery_test.go]: This was replaced as per the above note.

File: fastly/block_fastly_service_logging_bigquery_test.go (hunk 1 above)
Comment (philippschulte) [reply fastly/block_fastly_service_logging_bigquery_test.go]: I know that but instead of continuing with this bad pattern it would be great if you could refactor it.

Comment (rcaril) [reply fastly/block_fastly_service_logging_bigquery_test.go]: Sure, no problem!

File: fastly/block_fastly_service_logging_bigquery_test.go (hunk 4 above)
Comment (rcaril) [reply fastly/block_fastly_service_logging_bigquery_test.go]: Yes, we don't document this clearly it seems. After some testing, I see that we also set the same defaults for Compute logging, but the `format` field is actually not used for Compute logging itself to my understanding.

File: fastly/block_fastly_service_logging_bigquery_test.go (hunk 5 above)
Comment (rcaril) [reply fastly/block_fastly_service_logging_bigquery_test.go]: I believe I did this already in [0837fc5](https://github.com/fastly/terraform-provider-fastly/pull/1066/commits/0837fc5646222c2d852ad34807a86e6c7d318291)

**Review by anthony-gomez-fastly — APPROVED** [review]

**Comment by rcaril** [comment]
I still need to rewrite the Sumologic and Syslog tests, so ignore those for now. The rest should be ready to review.

**Comment by rcaril** [comment]
Test examples:

```
--- PASS: TestAccFastlyServiceVCL_splunk_basic_compute (12.74s)
--- PASS: TestAccFastlyServiceVCL_splunk_basic (26.43s)
PASS
ok   
```
```
--- PASS: TestAccFastlyServiceVCL_logging_syslog_basic_compute (11.77s)
--- PASS: TestAccFastlyServiceVCL_logging_syslog_basic (24.15s)
PASS
ok      github.com/fastly/terraform-provider-fastly/fastly      24.159s
```

```
--- PASS: TestAccFastlyServiceVCL_logging_sumologic_basic_compute (12.84s)
--- PASS: TestAccFastlyServiceVCL_logging_sumologic_basic (25.66s)
PASS
ok  
```

```
--- PASS: TestAccFastlyServiceVCL_logging_sftp_basic_compute (12.83s)
--- PASS: TestAccFastlyServiceVCL_logging_sftp_basic (25.34s)
PASS
```

File: fastly/constants.go [hunk 7]
```diff
@@ -11,3 +11,682 @@ const TimestampFormatDescription = "The `strftime` specified timestamp formattin
...
+    "response_state": "%{json.escape(fastly_info.state)}V",
+    "response_status": %{resp.status}V,
+    "response_reason": %{if(resp.response, "%22"+json.escape(resp.response)+"%22", "null")}V,
+    "response_body_size": %{resp.body_bytes_written}V,
+    "fastly_server": "%{json.escape(server.identity)}V",
+    "fastly_is_edge": %{if(fastly.ff.visits_this_service == 0, "true", "false")}V
+  }`
+
+// LoggingLogentriesDefaultFormat - Default format for Logentries logging
+// ** This seems to be depricated **
```
kpfleming:
```suggestion
// ** This seems to be deprecated **
```

fastly/constants.go
```diff
@@ -11,3 +11,682 @@ const TimestampFormatDescription = "The `strftime` specified timestamp formattin
 
 // SnippetTypeDescription describes the VCL snippet location.
 const SnippetTypeDescription = "The location in generated VCL where the snippet should be placed (can be one of `init`, `recv`, `hash`, `hit`, `miss`, `pass`, `fetch`, `error`, `deliver`, `log` or `none`)"
+
+// LoggingFormatUpdate is the generic logging format used for tests.
+const LoggingFormatUpdate = "%h %l %u %t \"%r\" %>s %b"
+
+// LoggingBigQueryDefaultFormat - Default format for BigQuery logging.
```
kpfleming: The standard for `godoc` documentation is full sentences, so all of these should replace '- Default' with 'is the default'.

kpfleming (changes requested):
Excellent work! A couple of small changes requested, but non-functional.

fastly/constants.go
```diff
@@ -15,7 +15,7 @@ const SnippetTypeDescription = "The location in generated VCL where the snippet
 // LoggingFormatUpdate is the generic logging format used for tests.
 const LoggingFormatUpdate = "%h %l %u %t \"%r\" %>s %b"
 
-// LoggingBigQueryDefaultFormat - Default format for BigQuery logging.
+// LoggingBigQueryDefaultFormat - is the default format for BigQuery logging.
```
kpfleming: Please change all of them, not just this one :-)

rcaril: Oops!

kpfleming: Also no hyphen, just a regular sentence.

rcaril: Corrected in [bb9c12d](https://github.com/fastly/terraform-provider-fastly/pull/1066/commits/bb9c12d2b882ba214cd283af9ade040837812283)

rcaril: [572c950](https://github.com/fastly/terraform-provider-fastly/pull/1066/commits/572c9509b77dc52fdca81dbf69273aa8642faf27) removed the extra character here.

kpfleming (approved):

rcaril:
Test for last two defaults:
```
--- PASS: TestAccFastlyServiceVCL_logging_digitalocean_basic_compute (13.86s)
--- PASS: TestAccFastlyServiceVCL_logging_digitalocean_basic (25.35s)
PASS
```

```
--- PASS: TestAccFastlyServiceVCL_logging_elasticsearch_basic_compute (12.90s)
--- PASS: TestAccFastlyServiceVCL_logging_elasticsearch_basic (23.48s)
PASS
```

Issue #67, opened 2018-02-08
xff request setting default should be leave


jrynyt:
If you add a `request_setting` but do not specify an `xff` parameter, it defaults to `append`, which actually inserts vcl to change X-Forwarded-For even if your request setting has nothing to do with that. Or if you have multiple request settings, one of which specifies `xff`, a subsequent request setting that doesn't include `xff` will override it.

The default should be `leave`, which does nothing. The fastly API reference does not specify a default: https://docs.fastly.com/api/config#request_settings

sjparkinson:
This is somewhat related to https://github.com/fastly/terraform-provider-fastly/issues/489, as re-creating the Force TLS setting currently needs a block like:

```hcl
request_setting {
  name          = "Generated by force TLS and enable HSTS"
  force_ssl     = true
  xff           = ""
}
```

rcaril:
This has been corrected in https://github.com/fastly/terraform-provider-fastly/pull/1078.

PR #1078, opened 2025-08-19
fix(service/requestsetting): remove incorrect default value for xff attribute 

rcaril:
All Submissions:

### New Feature Submissions:

### Changes to Core Features:

When configuring the `request_setting` block for a `fastly_service_vcl` resource, the XFF field was being set to a default value in which was not intended, not matches the behavior of the API. This PR updates the TF behavior to match that of the API. 

Testing:
```
--- PASS: TestAccFastlyServiceVCLRequestSetting_basic (38.60s)
PASS
ok      github.com/fastly/terraform-provider-fastly/fastly      38.615s
```

Resolves issue https://github.com/fastly/terraform-provider-fastly/issues/67

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - feat(compute acls): add support for compute ACLs ([#1031](https://github.com/fastly/terraform-provider-fastly/pull/1031))
 - refactor(resource_fastly_domain_v1) to use domainmanagement imports and types ([#1074](https://github.com/fastly/terraform-provider-fastly/pull/1074))
 - feat(ngwaf/rules): add support for deception and allow_interactive actions ([#1077](https://github.com/fastly/terraform-provider-fastly/pull/1077))
+- bug(ngwaf/rules): removed a incorrect default value for xff attribute in a request_setting block for a fastly_service_vcl resource  ([#1078](https://github.com/fastly/terraform-provider-fastly/pull/1078))
 
 ### BUG FIXES:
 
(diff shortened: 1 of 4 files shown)


rcaril:
Confirmed this is working as expected locally.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - feat(compute acls): add support for compute ACLs ([#1031](https://github.com/fastly/terraform-provider-fastly/pull/1031))
 - refactor(resource_fastly_domain_v1) to use domainmanagement imports and types ([#1074](https://github.com/fastly/terraform-provider-fastly/pull/1074))
 - feat(ngwaf/rules): add support for deception and allow_interactive actions ([#1077](https://github.com/fastly/terraform-provider-fastly/pull/1077))
+- bug(ngwaf/rules): removed a incorrect default value for xff attribute in a request_setting block for a fastly_service_vcl resource  ([#1078](https://github.com/fastly/terraform-provider-fastly/pull/1078))
```
kpfleming: This is not correct; the tag should be `fix` and not `bug`, and the fix was made in `service_vcl` and not `ngwaf/rules`. The description of the fix does not need to include the resource which was modified because it will be specified in the tag/component prefix.

kpfleming (changes requested):

CHANGELOG.md
rcaril: Copied and pasted too quickly, updating now.

kpfleming (approved):

PR #1084, opened 2025-08-29
Fix the documented ngwaf_workspace example to make a a valid declaration

loxal:
This PR fixes a doc issue and I've validated the fix with v8.0.0.

All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/examples/resources/ngwaf_workspace_basic_usage.tf
+++ b/examples/resources/ngwaf_workspace_basic_usage.tf
@@ -1,12 +1,12 @@
 resource "fastly_ngwaf_workspace" "demo" {
-  name = "demofastly"
+  name                         = "Demo"
+  description                  = "Testing"
+  mode                         = "block"
 
-  description = "A reference setup"
-  mode        = "block"
   attack_signal_thresholds {
-    immediate = true
+    one_minute  = 100
+    ten_minutes = 500
+    one_hour    = 1000
+    immediate   = true
   }
-  default_blocking_response_code = 406
-  ip_anonymization               = null
-  client_ip_headers              = []
 }
--- a/examples/resources/ngwaf_workspace_basic_usage.tf
+++ b/examples/resources/ngwaf_workspace_basic_usage.tf
@@ -1,5 +1,12 @@
 resource "fastly_ngwaf_workspace" "demo" {
   name = "demofastly"
 
-  force_destroy = true
+  description = "A reference setup"
+  mode        = "block"
+  attack_signal_thresholds {
+    immediate = true
+  }
+  default_blocking_response_code = 406
+  ip_anonymization               = null
+  client_ip_headers              = []
 }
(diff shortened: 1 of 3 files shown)


examples/resources/ngwaf_workspace_basic_usage.tf
```diff
@@ -1,5 +1,12 @@
 resource "fastly_ngwaf_workspace" "demo" {
   name = "demofastly"
 
-  force_destroy = true
-}
\ No newline at end of file
+  description = "A reference setup"
+  mode        = "block"
+  attack_signal_thresholds {
```
rcaril, line 6: The defaults we are requiring in TF are the following:

```
resource "fastly_ngwaf_workspace" "example" {
  name                         = "Sample"
  description                  = "Testing"
  mode                         = "block"

  attack_signal_thresholds {
    one_minute  = 100
    ten_minutes = 500
    one_hour    = 1000
    immediate   = true
  }
}
```

This basic example does indeed need an update, but the  default_blocking_response_code,  ip_anonymization  and client_ip_headers  fields are not absolutely necessary.

loxal: True those fields are not required but other example contain also non-mandatory fields. I am fine in adjusting the PR if everything else is ok. The idea behind my example was to provide the user with additional relevant fields. The 
```
    one_minute  = 100
    ten_minutes = 500
    one_hour    = 1000
```
...fields are not required as well. I am fine in using your example as long as we fix invalid docs.

rcaril: Ya, I think the rest is fine. You can also run `make generate-docs` after you update the example to update the doc files automatically. 

Looping in @anthony-gomez-fastly as well to eyeball this PR.

examples/resources/ngwaf_workspace_basic_usage.tf
```diff
@@ -1,5 +1,12 @@
 resource "fastly_ngwaf_workspace" "demo" {
-  name = "demofastly"
+  name                         = "Demo"
+  description                  = "Testing"
+  mode                         = "block"
 
-  force_destroy = true
```
anthony-gomez-fastly, line 4: was this purposely removed?

rcaril: Terraform actually throws an error when this exists, I just tried leaving it in my test 🤔

loxal: Yes:

```
│ Error: Unsupported argument
│ 
│   on lontu.tf line 83, in resource "fastly_ngwaf_workspace" "lontu":
│   83:   force_destroy = true
│ 
│ An argument named "force_destroy" is not expected here.
```

rcaril (approved):
Looks good after the suggested changes from my side.

anthony-gomez-fastly (approved):

PR #1095, opened 2025-09-09
[CDTOOL-1160] - Support for New Relic OTLP on Compute Services

rcaril:
### Change summary

This PR introduces support for New Relic OTLP logging on Compute Services. 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### Other Consideration: 

Due to API inconstancy between VCL and Compute services, we needed to override some attribute defaults in Terraform to prevent drift. 

Tests:
```
?       github.com/fastly/terraform-provider-fastly     [no test files]
=== RUN   TestAccFastlyServiceVCL_logging_newrelicotlp_basic_compute
=== PAUSE TestAccFastlyServiceVCL_logging_newrelicotlp_basic_compute
=== CONT  TestAccFastlyServiceVCL_logging_newrelicotlp_basic_compute
--- PASS: TestAccFastlyServiceVCL_logging_newrelicotlp_basic_compute (13.21s)
PASS
ok      github.com/fastly/terraform-provider-fastly/fastly      13.227s
```


rcaril:
Looks like the `TestAccFastlyServiceVCL_logging_newrelicotlp_basic` test was already failing in the `main` branch. I'll try to correct this in my PR.

philippschulte:
@rcaril I am working to fix the test!

rcaril:
> @rcaril I am working to fix the test!

I believe it's been corrected in [e882fae](https://github.com/fastly/terraform-provider-fastly/pull/1095/commits/e882faefdd04223ccefa863991273023e3e87696) @philippschulte

fastly/block_fastly_service_logging_newrelicotlp.go
```diff
@@ -255,3 +275,8 @@ func (h *NewRelicOTLPServiceAttributeHandler) buildCreate(newrelicotlpMap any, s
 
 	return opts
 }
+
+func (h *NewRelicOTLPServiceAttributeHandler) pruneVCLLoggingAttributesForOTLP(data map[string]any) {
```
philippschulte: This is only called once and its logic is trivial and unlikely to grow. I don't insist but I wouldn't use a function and instead replace the call with:

```
for _, element := range dll {
	if h.GetServiceMetadata().serviceType == ServiceTypeCompute {
		delete(element, "format")
		delete(element, "response_condition")
	} else {
		h.pruneVCLLoggingAttributes(element)
	}
}
```

That is just my personal reference but I keep the decision up to you! If you keep the function I would recommend to rename it to something like: `pruneComputeLoggingAttributes` to better reflect why these fields are being pruned.

philippschulte (approved):
LGTM. I would recommend to just inline the two `delete()` calls and remove the function entirely but I keep this up to you. Thanks

fastly/block_fastly_service_logging_newrelicotlp.go
rcaril: This is good feedback. I did request some API changes from the logging team for Compute logging specifically, so it is possible we may need to prune more fields in the future. I think leaving this as a modular code block makes sense with that in mind, also to provide a good template for future use cases.

rcaril: Updated the name to be more clear in [b6d8a2b](https://github.com/fastly/terraform-provider-fastly/pull/1095/commits/b6d8a2ba30a36308910ae757f29d589b2905b6c1).

PR #1096, opened 2025-09-11
CDTOOL-1115 fix example terraform blocks to be closer to real code

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

### DIFF
--- a/docs/resources/service_vcl.md
+++ b/docs/resources/service_vcl.md
@@ -99,12 +99,6 @@ resource "fastly_service_vcl" "demo" {
 
 resource "aws_s3_bucket" "website" {
   bucket = "demo.notexample.com"
-  acl    = "public-read"
-
-  website {
-    index_document = "index.html"
-    error_document = "error.html"
-  }
 }
 ```
 
--- a/docs/resources/service_vcl.md
+++ b/docs/resources/service_vcl.md
@@ -97,8 +97,20 @@ resource "fastly_service_vcl" "demo" {
   force_destroy = true
 }
 
-resource "aws_s3_bucket" "website" {
-  bucket = "demo.notexample.com"
+resource "aws_s3_bucket" "website_bucket" {
+  bucket = "your-unique-website-bucket-name" # Replace with a globally unique name
+}
+
+resource "aws_s3_bucket_website_configuration" "website_config" {
+  bucket = aws_s3_bucket.website_bucket.id
+
+  index_document {
+    suffix = "index.html"
+  }
+
+  error_document {
+    key = "error.html"
+  }
 }
 ```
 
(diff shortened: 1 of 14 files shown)


docs/resources/service_vcl.md
```diff
@@ -99,12 +99,6 @@ resource "fastly_service_vcl" "demo" {
 
 resource "aws_s3_bucket" "website" {
   bucket = "demo.notexample.com"
-  acl    = "public-read"
-
-  website {
-    index_document = "index.html"
-    error_document = "error.html"
-  }
```
philippschulte, line 107: Why is this getting removed?

examples/resources/alert_basic_usage.tf
```diff
@@ -1,22 +1,28 @@
 resource "fastly_service_vcl" "example" {
-  name = "my_vcl_service"
-  # ...
+  name = "demofastly"
+  domain {
+    name = "demo.notexample.com"
```
philippschulte: Some `domain` blocks include a comment, others don’t. I know it is optional but maybe we should be consistent across all examples but I keep this up to you.

philippschulte (dismissed):
I don't see any issues but I don't understand why we remove fields from the `aws` resource blocks. I don't even see the reference of the `aws` resource blocks in the `fastly` resource blocks! Maybe we should remove those blocks completely?

docs/resources/service_vcl.md
anthony-gomez-fastly: Screenshot 2025-09-11 at 12 48 23 PM
these warnings

examples/resources/alert_basic_usage.tf
anthony-gomez-fastly: good call, i will add them

anthony-gomez-fastly: fixed in `4b258ce7`

docs/resources/service_vcl.md
anthony-gomez-fastly: this is using aws's terraform resource, which has changes since these examples were last updated

philippschulte (approved):
Please consider to update the `aws` resources  for static website hosting to the latest syntax. Here is an example:

```
resource "aws_s3_bucket" "website_bucket" {
  bucket = "your-unique-website-bucket-name" # Replace with a globally unique name
  acl    = "public-read" # Allows public read access for static website hosting
}
```

```
resource "aws_s3_bucket_website_configuration" "website_config" {
  bucket = aws_s3_bucket.website_bucket.id

  index_document {
    suffix = "index.html"
  }

  error_document {
    key = "error.html"
  }
}
```

However, I leave this up to you! Thanks for your work!

PR #1097, opened 2025-09-11
Add Period attribute to logging https endpoint.

bowrocker:
### Change summary

 

 All Submissions:

### Changes to Core Features:

### User Impact

* [ ] What is the user impact of this change?  This change maintains the status quo in the case where customers make no changes to their HTTPS endpoint configuration and only leverages the new functionality if the customer explicitly opts into it. The risk of impact is low because customers using the HTTPS logging endpoint will be unaffected by this change by default.


philippschulte:
This PR depends on https://github.com/fastly/go-fastly/pull/749. The `period` field is not yet documented in our public docs.

bowrocker:
@philippschulte the documentation PR has been merged: https://github.com/fastly/api-documentation/pull/1192

Please let me know what else you may need, we'd like to get this merged as soon as possible. Thanks!

kpfleming:
As I noted on the PR in the fastly-cli repository, a proper review of this PR can't be done until there has been a go-fastly release which supports the new attribute and the tests in this PR can be confirmed to pass.

kpfleming:
This PR can now be rebased on `main` and will have go-fastly 12.0.0, which means the code should compile and the tests should be runnable.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### BREAKING:
 
 ### ENHANCEMENTS:
+- feat(logging): add support for Period HTTPS logging endpoint ([#](https://github.com/fastly/terraform-provider-fastly/pull/))
```
rcaril: We'll want to ensure that we specify logging endpoint here ('https') as well as the PR number. 

```suggestion
- feat(logging/https): add support for Period HTTPS logging endpoint ([#1097](https://github.com/fastly/terraform-provider-fastly/pull/1097))
```

docs/resources/service_compute.md
```diff
@@ -513,6 +513,7 @@ Optional:
 - `json_format` (String) Formats log entries as JSON. Can be either disabled (`0`), array of json (`1`), or newline delimited json (`2`)
 - `message_type` (String) How the message should be formatted. Can be either `classic`, `loggly`, `logplex` or `blank`. Default is `classic`
 - `method` (String) HTTP method used for request. Can be either `POST` or `PUT`. Default `POST`
+- `period` (Number)  How frequently the logs should be transferred, in seconds. Default `3600`
```
rcaril: The default here seems to be incorrect. 

Please use the `make validate-docs` command to generate docs, you can find instructions on how to use this in the `DOCUMENTATION.md` [file](https://github.com/fastly/terraform-provider-fastly/blob/main/DOCUMENTATION.md)!

docs/resources/service_vcl.md
```diff
@@ -814,6 +814,7 @@ Optional:
 - `json_format` (String) Formats log entries as JSON. Can be either disabled (`0`), array of json (`1`), or newline delimited json (`2`)
 - `message_type` (String) How the message should be formatted. Can be either `classic`, `loggly`, `logplex` or `blank`. Default is `classic`
 - `method` (String) HTTP method used for request. Can be either `POST` or `PUT`. Default `POST`
+- `period` (Number) How frequently the logs should be transferred in seconds. Default `3600`
```
rcaril: The default here is also incorrect, running the `make validate-docs` command will correct this as mentioned above.

fastly/block_fastly_service_logging_https.go
```diff
@@ -92,6 +92,12 @@ func (h *HTTPSLoggingServiceAttributeHandler) GetSchema() *schema.Schema {
 			Required:    true,
 			Description: "The unique name of the HTTPS logging endpoint. It is important to note that changing this attribute will delete and recreate the resource",
 		},
+		"period": {
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Default:     5,
```
rcaril: Since the value of `0` is equivalent to `5`, according to our docs, we should use validation to ensure that this value is greater than 0.

fastly/block_fastly_service_logging_https_test.go
```diff
@@ -72,6 +74,7 @@ func TestAccFastlyServiceVCL_httpslogging_basic(t *testing.T) {
 		MessageType:       gofastly.ToPointer("blank"),
 		Method:            gofastly.ToPointer("POST"),
 		Name:              gofastly.ToPointer("httpslogger2"),
+		Period:            gofastly.ToPointer(0),
```
rcaril: We should try to use different values for testing, specifically values that do not signify something on the Fastly side. An arbitrary value, such as `3`, would be better.

rcaril (changes requested):
A few changes requested here. 

We'll also need to ensure we handle default values for imports. You can look at the flatten pattern implemented [here](https://github.com/fastly/terraform-provider-fastly/pull/1103/files) if you like for reference. It would be great if you can test this locally as well. 

In addition to this, please run the updated tests locally and provide the output in a comment. You can read more about how to use the `make testacc` feature in our [README.md](https://github.com/fastly/terraform-provider-fastly/blob/main/README.md#testing) file.

fastly/block_fastly_service_logging_https.go
```diff
@@ -92,6 +92,12 @@ func (h *HTTPSLoggingServiceAttributeHandler) GetSchema() *schema.Schema {
...
+			Description: "How frequently the logs should be transferred, in seconds (Default 5)",
```
rcaril: Please ensure that we use the same `Description` as provided in our public API docs.

fastly/block_fastly_service_logging_https.go
bowrocker: @rcaril A `0` period is accepted by the API and is equivalent to the default:

https://github.com/fastly/Northstar/blob/master/lib/models/logging_https.rb#L63

So I think this is ok?

docs/resources/service_compute.md
bowrocker: Default fixed in 2d670c6ade89f928355c4dd9514ef7f864c00fae.

I did run `make validate-docs` but I am getting a lot of errors in unrelated resources, would you point me to how to run only for my change?

fastly/block_fastly_service_logging_https_test.go
bowrocker: Fixed in 2d670c6ade89f928355c4dd9514ef7f864c00fae. Thanks!

docs/resources/service_compute.md
kpfleming: You only need to run `make generate-docs`, not `make validate-docs`.

rcaril: Ah good catch - typo on my side when I was copying from our MD file.

fastly/block_fastly_service_logging_https.go
rcaril: When we talked about this internally, the understanding was that if the value of `0` had the same result as the value of `5`, why would a user use `0`? If this understanding is true, it may be more clear direction to restrict the options available.

fastly/block_fastly_service_logging_https_test.go
rcaril: Thanks! We are still using the value of `0` for several tests, it may be best here to use some other inputs. 

We should also ensure we test the default behavior in one scenario by omitting the period field in one of the Terraform blocks to ensure that the resulting value is being set to `5`. This is also a good test to ensure we can track if the API default ever changes. You can take a look at the `block_fastly_service_logging_s3_test.go` file to see how we are doing this there.

rcaril (changes requested):
Updated conversations above with a few items. Please also output the results of a successful test.

rcaril:
Also a reminder on this note as well :) 

> We'll also need to ensure we handle default values for imports. You can look at the flatten pattern implemented [here](https://github.com/fastly/terraform-provider-fastly/pull/1103/files) if you like for reference. It would be great if you can test this locally as well.

rcaril:
Updates in [fadd166](https://github.com/fastly/terraform-provider-fastly/pull/1097/commits/fadd166e75fd8279bbdead2ebad9661d04ff48e7):

- Added a prune function for VCL attributes in the Compute test to ensure that `period` wasn't being removed
- Removed the prior iteration of state file setting for `period` that I had suggested 
- Updated values of `period` in tests to ensure that we weren't using the default values in tests

rcaril:
Tests:

```
-- PASS: TestAccFastlyServiceVCL_httpslogging_basic_compute (15.13s)
--- PASS: TestAccFastlyServiceVCL_httpslogging_basic (42.92s)
PASS
ok      github.com/fastly/terraform-provider-fastly/fastly      42.934s
```

rcaril:
Local testing works as expected as well.

rcaril (approved):

kpfleming (approved):

PR #1117, opened 2025-10-06
[Breaking] [CDTOOL-1179] - Unify 'gzip_level' and 'compression_codec' logging attributes

rcaril:
### Change summary

This PR combined the  `gzip_level` and `compression_codec` logging attributes into a single attribute `compression`. 

This stemmed from a conversation that was initially raised internally [here](https://fastly.slack.com/archives/C01E7FV8P5H/p1758555293782509), which went over the poor Terraform experience of trying to figure out the right combinations of 'gzip_level' and 'compression_codec'. To remedy this, we are replacing these confusing combinations with a given list of possible options that are are mapping back to the API accordingly via one new attribute. 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

The `gzip_level` and `compression_codec` attributes have been removed from the following logging endpoint schemas:

- `block_fastly_service_logging_blobstorag`
- `block_fastly_service_logging_cloudfiles`
- `block_fastly_service_logging_digitalocean`
- `block_fastly_service_logging_ftp`
- `block_fastly_service_logging_gcs`
- `block_fastly_service_logging_https`
- `block_fastly_service_logging_openstack`
- `block_fastly_service_logging_s3`
- `block_fastly_service_logging_sftp`

These attributes have been replaced with `compression` which has the following options available:
- "zstd"
- "snappy"
- "gzip-0"
- "gzip-1"
- "gzip-2"
- "gzip-3"
- "gzip-4"
- "gzip-5"
- "gzip-6"
- "gzip-7"
- "gzip-8"
- "gzip-9"
- "none"

More reading on logging compression can be found here: 
https://www.fastly.com/documentation/guides/integrations/streaming-logs/changing-log-compression-options/


rcaril:
## Tests:
```
--- PASS: TestAccFastlyServiceVCL_blobstoragelogging_basic_compute (15.91s)
--- PASS: TestAccFastlyServiceVCL_blobstoragelogging_basic (45.98s)
PASS
ok 
```

```
--- PASS: TestAccFastlyServiceVCL_logging_cloudfiles_basic_compute (15.27s)
--- PASS: TestAccFastlyServiceVCL_logging_cloudfiles_basic (43.25s)
PASS
ok  
```

```
--- PASS: TestAccFastlyServiceVCL_logging_digitalocean_basic_compute (13.81s)
--- PASS: TestAccFastlyServiceVCL_logging_digitalocean_basic (43.61s)
PASS
ok 
```

```
--- PASS: TestAccFastlyServiceVCL_logging_ftp_basic_compute (14.64s)
--- PASS: TestAccFastlyServiceVCL_logging_ftp_basic (26.95s)
PASS
ok
```

```
--- PASS: TestAccFastlyServiceVCL_gcslogging_basic_compute (14.76s)
--- PASS: TestAccFastlyServiceVCL_gcslogging_basic (47.07s)
PASS
ok  
```

```
--- PASS: TestAccFastlyServiceVCL_httpslogging_basic_compute (15.50s)
--- PASS: TestAccFastlyServiceVCL_httpslogging_basic (42.38s)
PASS
ok  
```

```
--- PASS: TestAccFastlyServiceVCL_logging_openstack_basic_compute (14.32s)
--- PASS: TestAccFastlyServiceVCL_logging_openstack_basic (43.29s)
PASS
ok 
```

```
--- PASS: TestAccFastlyServiceVCL_s3logging_basic_compute (13.97s)
--- PASS: TestAccFastlyServiceVCL_s3logging_basic (45.25s)
PASS
ok 
```

```
--- PASS: TestAccFastlyServiceVCL_logging_sftp_basic_compute (15.16s)
--- PASS: TestAccFastlyServiceVCL_logging_sftp_basic (43.89s)
PASS
ok 
```

mdenomy:
I was wondering if there is a deprecation pathway we can explore, going directly to a breaking change seems abrupt and could cause customer problems. I think this is a good abstraction, but I wonder if the new attribute were optional and the others were left as is we could give customers a gradual off-ramp and give them time to change over.

I know you (and me) all have discussed it, but would be curious about the breaking/deprecated aspect of the change

rcaril:
> I was wondering if there is a deprecation pathway we can explore, going directly to a breaking change seems abrupt and could cause customer problems. I think this is a good abstraction, but I wonder if the new attribute were optional and the others were left as is we could give customers a gradual off-ramp and give them time to change over.
> 
> I know you (and me) all have discussed it, but would be curious about the breaking/deprecated aspect of the change

You make a good point, perhaps the quality of life update will outweigh the breaking change impact? 

@kpfleming - what are your thoughts?

fastly/logging_compression.go
```diff
@@ -0,0 +1,134 @@
...
+	CompressionGzip5  = "gzip-5"
+	CompressionGzip6  = "gzip-6"
+	CompressionGzip7  = "gzip-7"
+	CompressionGzip8  = "gzip-8"
+	CompressionGzip9  = "gzip-9"
+	CompressionNone   = "none"
+)
+
+// CompressionDescription describes the compression field for logging endpoints.
+const CompressionDescription = "Compression format for log data. Valid values: zstd, snappy, gzip-0 through gzip-9, or none"
```
kpfleming:
```suggestion
const CompressionDescription = "Compression format for log data. Valid values: zstd, snappy, gzip-0 through gzip-9, and none"
```
Is there any functional difference between specifying `compression = none` and not specifying `compression` at all?

fastly/logging_compression.go
```diff
@@ -0,0 +1,134 @@
...
+// Mapping:
+// - "zstd"   -> compression_codec: "zstd", gzip_level: nil
+// - "snappy" -> compression_codec: "snappy", gzip_level: nil
+// - "none"   -> compression_codec: "" (empty string), gzip_level: 0
+// - "gzip-0" -> compression_codec: "gzip", gzip_level: 0
+// - "gzip-1" through "gzip-9" -> (nil), gzip_level: 1-9
+//
+// Note: When transitioning from zstd/snappy to none or gzip levels, we must
+// explicitly send compression_codec as empty string ("") to clear the previous value.
+func CompressionToAPIFields(compression string) (compressionCodec *string, gzipLevel *int) {
```
kpfleming, line 60: I'm tempted to say that this function (and the one below it) should have unit tests, but since they are just table-based mapping functions the unit tests would look almost the same. @anthony-gomez-fastly any thoughts here?

fastly/logging_compression.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+		} else if *compressionCodec == "" && gzipLevel != nil && *gzipLevel == 0 {
+			// When compression_codec is explicitly "" (empty string) and gzip_level is 0,
+			// this means the user set compression = "none"
+			return CompressionNone
+		}
+	}
+
+	// If compressionCodec is nil and gzipLevel is 0 or nil, compression was not set by the user
+	// Return empty string so it won't be added to state
```
kpfleming: The comment makes sense but describes the opposite of what the code does!

fastly/block_fastly_service_logging_sftp.go
```diff
@@ -246,11 +236,10 @@ func (h *SFTPServiceAttributeHandler) Update(ctx context.Context, d *schema.Reso
...
 		opts.FormatVersion = gofastly.ToPointer(v.(int))
 	}
-	if v, ok := modified["compression_codec"]; ok {
-		opts.CompressionCodec = gofastly.ToPointer(v.(string))
-	}
-	if v, ok := modified["gzip_level"]; ok {
-		opts.GzipLevel = gofastly.ToPointer(v.(int))
+	if v, ok := modified["compression"]; ok {
+		compressionCodec, gzipLevel := CompressionToAPIFields(v.(string))
+		opts.CompressionCodec = compressionCodec
```
kpfleming: Nit: These can be collapsed to a single line.

fastly/block_fastly_service_logging_sftp.go
```diff
@@ -301,33 +290,22 @@ func (h *SFTPServiceAttributeHandler) Delete(ctx context.Context, d *schema.Reso
 	return nil
 }
 
+// pruneVCLLoggingAttributes removes VCL-only attributes from Compute service data.
+// For SFTP logging, period is not VCL-only, so we preserve it.
```
kpfleming, line 292: This appears to have been copied from the HTTPS endpoint, which is the only one that supports `period`. There shouldn't be any references to `period` in any of the other logging endpoints.

fastly/block_fastly_service_logging_sftp.go
```diff
@@ -381,9 +356,14 @@ func flattenSFTP(remoteState []*gofastly.SFTP, localState []any) []map[string]an
 		if resource.Placement != nil {
 			data["placement"] = *resource.Placement
 		}
-		if resource.CompressionCodec != nil {
-			data["compression_codec"] = *resource.CompressionCodec
+
+		// compression represents the combined value of the compression_codec and gzip_level
+		// attributes that we will need to parse to the API accordingly
+		compression := APIFieldsToCompression(resource.CompressionCodec, resource.GzipLevel)
+		if compression != "" {
```
kpfleming: Nit: These can be combined into a single line.

kpfleming (changes requested):
Some minor changes requested and one clarification question.

kpfleming:
> I was wondering if there is a deprecation pathway we can explore, going directly to a breaking change seems abrupt and could cause customer problems. I think this is a good abstraction, but I wonder if the new attribute were optional and the others were left as is we could give customers a gradual off-ramp and give them time to change over.

It appears that the Terraform provider framework does support deprecations, so it would be possible to label the existing fields as `deprecated`, which would generate warning messages to users but not be a breaking change.

It is also possible to mark schema attributes as "conflicts with" (or something similar) to ensure that a user cannot use both the old and new attributes at the same time. I think it would be worth the time to experiment with that to see if we can provide a non-breaking upgrade path.

fastly/logging_compression.go
rcaril: I'll double check, I know I made some changes to the logic halfway though!

fastly/block_fastly_service_logging_sftp.go
rcaril: I believe several other logging endpoints have the `period` attribute: 

https://github.com/fastly/terraform-provider-fastly/blob/22c6da075e4f214330546c30750f6dedbbf7c33a/fastly/block_fastly_service_logging_digitalocean.go#L87
https://github.com/fastly/terraform-provider-fastly/blob/22c6da075e4f214330546c30750f6dedbbf7c33a/fastly/block_fastly_service_logging_ftp.go#L81
https://github.com/fastly/terraform-provider-fastly/blob/22c6da075e4f214330546c30750f6dedbbf7c33a/fastly/block_fastly_service_logging_gcs.go#L81

I don't know why our prior tests weren't picking this up, but I had a drift issue whenever I attempted to run the Compute test without appending the appropriate pruning function. 

I'll investigate more as to why this is.

fastly/logging_compression.go
rcaril: Nope, setting the attribute to `none` should be the same as removing it from HCL. 

I'll update the text - I thought I had changed the `or` to `and` already 🤔

kpfleming: In that case I'm not sure I see what the value of offering `none` as an option is, although I don't have a strong opinion about it. If the default on the Fastly side is no compression, and it's not specified in the HCL, then user should understand that they won't get compression.

rcaril: This makes sense, I think I was trying to mirror the user experience elsewhere - giving the option of setting compression to the actual of option of  `none`. But to your point, normally this can be done by removing the attribute itself as is standard in Terraform. 

I concur, I also am indifferent on if we should provide the `none` option.

fastly/logging_compression.go
rcaril: I did add some additional tests for all endpoints that support this combination of attributes to change a given compression type to `none` to ensure the API handles this operation correctly. I am happy to add another operation of switching from compression X to Y.

fastly/block_fastly_service_logging_sftp.go
rcaril: Addressed in https://github.com/fastly/terraform-provider-fastly/pull/1117/commits/d4070b1d9ce7941b4ca6e463e9ff3ad1d3c4bb3b 

```
		if compression := APIFieldsToCompression(resource.CompressionCodec, resource.GzipLevel); compression != "" {
			data["compression"] = compression
		}
```

fastly/block_fastly_service_logging_sftp.go
rcaril: Addressed in https://github.com/fastly/terraform-provider-fastly/commit/d4070b1d9ce7941b4ca6e463e9ff3ad1d3c4bb3b

```
	if v, ok := modified["compression"]; ok {
		opts.CompressionCodec, opts.GzipLevel = CompressionToAPIFields(v.(string))
	}
```

fastly/logging_compression.go
rcaril: Strangely enough the comments wound up being accurate with the `else` statement in mind, they were just phrased in a very unclear / illogical way. I've updated the comments to be much more clear in https://github.com/fastly/terraform-provider-fastly/pull/1117/commits/d4070b1d9ce7941b4ca6e463e9ff3ad1d3c4bb3b. 

```
// compressionCodec is nil at this point, so we have to  handle cases where gzipLevel is set.
	if gzipLevel != nil && *gzipLevel >= 1 && *gzipLevel <= 9 {
		// Return the specific gzip level.
		return fmt.Sprintf("gzip-%d", *gzipLevel)
	}
	// Otherwise return an empty string to indicate that
	// compression wasn't set by the end user.
	return ""
}
```

rcaril:
Looking into the origins of this request, all other logging endpoints (other then HTTPS) that used compression have the same behavior of needing to exchange the values of both existing compression attributes in order to change the compression type. (Using `-1` as a gzip_level option was never intended) 

While this most likely is a uncommon workflow, a user would still have to either comment out or remove one of the attributes in this scenario. What is being suggested here is removing both values with a clear selection of options via a single attribuite. I would argue this is the same effort, if not less, then having to manipulate the two existing attributes. 


I'll start the effort of re-adding the old `gzip_level` & `compression_codec` code back and mark these items are deprecated so it's clear folks should use the new attribute. (assuming there aren't any conflicts with maintaining both)

rcaril:
After another report of this exact behavior, we realized that there was an issue in which we were sending the default value of `gzip_level` to the API during Update operations which would only happen when someone upgraded to >= v8.1.0, which is where compression for HTTPS was included in this provider. As such, we have decided to just fix the upgrade bug and implement this new compression attribute in the future instead. 

A in depth explanation can be found internally [here](https://fastly.slack.com/archives/C01E7FV8P5H/p1760037303399599).

PR #1118, opened 2025-10-09
[CDTOOL-1185] - Correct HTTPS 'gzip_level' Default Handling

rcaril:
### Change summary

This PR addresses a bug where users upgrading from the Terraform provider prior to `v8.1.0` to >= `8.1.0` will face an issue where HTTPS logging endpoints will throw a `gziip_level` error when no compression attributes are associated. 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

Full description can be found internally [here](https://fastly.slack.com/archives/C01E7FV8P5H/p1760037303399599). 

Essentially we are adding the same exception to the Update operation that the Read operation has, to not push the `gzip_level` value of `-1` (which is a flag) to the API. This will prevent users that are upgrading from another provider without a default `gzip_level` to have the appropriate flag be appended to their state file without sending an invalid payload.

### Additional Notes:
While this fix primarily was for the HTTPS logging endpoint (as this was the only endpoint that recently had support for compression added), the same fix was applied to the 8 other logging endpoints that use compression in case any users were on very old versions of the provider.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### ENHANCEMENTS:
 
 ### BUG FIXES:
+- fix(logging/https): corrected a bug where users that had a HTTPS logging block would recieve 'gzip_level' API errors after upgrading to the to v8.1.0 provider or later ([#1118](https://github.com/fastly/terraform-provider-fastly/pull/1118))
 
 ### DEPENDENCIES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### ENHANCEMENTS:
 
 ### BUG FIXES:
-- fix(logging/https): corrected a bug where users that had a HTTPS logging block would recieve 'gzip_level' API errors after upgrading to the to v8.1.0 provider or later ([#1118](https://github.com/fastly/terraform-provider-fastly/pull/1118))
+- fix(logging/https): corrected a bug where users that had a HTTPS logging block would encounter 'gzip_level' API errors after upgrading to the v8.1.0 provider or later ([#1118](https://github.com/fastly/terraform-provider-fastly/pull/1118))
 
 ### DEPENDENCIES:
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -25,7 +25,7 @@ terraform {
   required_providers {
     fastly = {
       source  = "fastly/fastly"
-      version = ">= 8.2.0"
+      version = ">= 8.3.0"
     }
   }
 }
--- a/fastly/block_fastly_service_logging_s3.go
+++ b/fastly/block_fastly_service_logging_s3.go
@@ -301,7 +301,11 @@ func (h *S3LoggingServiceAttributeHandler) Update(ctx context.Context, d *schema
 		opts.CompressionCodec = gofastly.ToPointer(v.(string))
 	}
 	if v, ok := modified["gzip_level"]; ok {
-		opts.GzipLevel = gofastly.ToPointer(v.(int))
+		// This condition prevents users on old provider versions from having
+		// compatability issues with the default 'gzip_level' value of `-1` when upgrading to more recent versions.
+		if gl := v.(int); gl != -1 {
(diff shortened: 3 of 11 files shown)


rcaril:
Local testing confirms that when upgrading from a prior provider with no `gzip_level` defined in state that we no longer get a `  Detail: gzip level must be within 0-9` error.

rcaril:
HTTPS Test:
```
=== CONT  TestAccFastlyServiceVCL_httpslogging_basic_compute
--- PASS: TestAccFastlyServiceVCL_httpslogging_basic_compute (19.43s)
--- PASS: TestAccFastlyServiceVCL_httpslogging_basic (47.27s)
PASS
ok      github.com/fastly/terraform-provider-fastly/fastly      47.282s
```

docs/index.md
```diff
@@ -25,7 +25,7 @@ terraform {
   required_providers {
     fastly = {
       source  = "fastly/fastly"
-      version = ">= 8.2.0"
+      version = ">= 8.3.0"
```
anthony-gomez-fastly, line 28: must have missed this in the last release, ty!

anthony-gomez-fastly (approved):
LGTM!

docs/index.md
rcaril: The linter actions test actually caught that a generate-docs needed to be ran haha

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### ENHANCEMENTS:
 
 ### BUG FIXES:
+- fix(logging/https): corrected a bug where users that had a HTTPS logging block would recieve 'gzip_level' API errors after upgrading to the to v8.1.0 provider or later ([#1118](https://github.com/fastly/terraform-provider-fastly/pull/1118))
```
kpfleming:
```suggestion
- fix(logging/https): corrected a bug where users that had a HTTPS logging block would encounter 'gzip_level' API errors after upgrading to the v8.1.0 provider or later ([#1118](https://github.com/fastly/terraform-provider-fastly/pull/1118))
```

fastly/block_fastly_service_logging_s3.go
```diff
@@ -301,7 +301,11 @@ func (h *S3LoggingServiceAttributeHandler) Update(ctx context.Context, d *schema
 		opts.CompressionCodec = gofastly.ToPointer(v.(string))
 	}
 	if v, ok := modified["gzip_level"]; ok {
-		opts.GzipLevel = gofastly.ToPointer(v.(int))
+		// This condition prevents users on old provider versions from having
+		// compatibility issues with the default 'gzip_level' value of `-1` when upgrading to more recent versions.
+		if gl := v.(int); gl != -1 {
+			opts.GzipLevel = gofastly.ToPointer(gl)
+		}
```
kpfleming, line 308: Note that this leaves a latent bug in place, I believe. If a customer has one of these logging endpoints with `gzip_level` configured but then wants to disable compression, they can remove `gzip_level` from their HCL but the change will not be sent to the API (because the attribute will now have the default value of -1 and this code won't send any update to the API). The workaround is to leave `gzip_level = 0` in the HCL instead of removing the attribute.

This problem has existed for as long as the provider has supported compression in these logging endpoints, so I don't think we need to do anything to address it; it's also quite unlikely that a customer would actually want to disable compression on an existing endpoint in my opinion.

kpfleming (approved):
Just some minor changes in the changelog entry, otherwise this looks good.

fastly/block_fastly_service_logging_s3.go
rcaril: @kpfleming - fair enough. To be super clear, you are suggesting we only update the HTTPS endpoint, correct?

kpfleming: No, I'm suggesting that we can't fix this in code at all, it's just the way this attribute works across all of the endpoints. If we get a customer complaint that they tried to disable compression by removing `gzip_level` from their HCL and the `apply` operation failed, we'll just acknowledge that it is a quirk of the provider and they'll have to use `gzip_level = 0` instead to disable compression.

PR #1143, opened 2025-11-10
test(block_fastly_service_logging_logentries_test): fix tests for logentries to account for API behavior

anthony-gomez-fastly:
### Change summary

Adjust tests to account for API changes. Seems like the default we were using no longer validates through the API so replaced it with a hardcoded string. The objects also changed so that deep equality no longer was usable so swapped to comparing the fields individually.

 All Submissions:

### New Feature Submissions:

Screenshot 2025-11-10 at 4 21 38 PM
Screenshot 2025-11-10 at 4 21 56 PM


fastly/block_fastly_service_logging_logentries_test.go
```diff
@@ -38,7 +38,7 @@ func TestAccFastlyServiceVCL_logentries_basic(t *testing.T) {
 		Port:              gofastly.ToPointer(10000),
 		UseTLS:            gofastly.ToPointer(false),
 		Token:             gofastly.ToPointer("newtoken"),
-		Format:            gofastly.ToPointer(LoggingFormatUpdate),
+		Format:            gofastly.ToPointer(`%h %u %t %r %>s`),
```
anthony-gomez-fastly, line 41: part of this is that it appears like we we don't allow non vcl services to set the format, so it always returns the default, is this correct behavior? @kpfleming @philippschulte

anthony-gomez-fastly: behavior defined [here](https://github.com/fastly/terraform-provider-fastly/blob/main/fastly/service_attribute_definition.go#L79)

kpfleming: That is correct, the format can only be set for Delivery (VCL) services because they are the only ones where the output to the log is created by the platform. For Compute (WASM) services the customer's code is completely responsible for constructing the log output.

kpfleming (approved):

PR #1178, opened 2026-01-16
[CDTOOL-1244] Add import support for `fastly_domain_service_link`

philippschulte:
### Change summary

This pull request adds proper import support for the `fastly_domain_service_link` resource and significantly improves its acceptance test coverage.

## Key changes

- **Import support**
  - Implement a custom importer for `fastly_domain_service_link` that derives both `domain_id` and `service_id` from the Fastly Domain Management API.
  - This allows importing a domain service link using only the domain ID, matching Fastly’s API semantics.

- **Domain read improvement**
  - Remove unnecessary service-scoped context from `fastly_domain` read operations, since reads are non-mutating and do not require request serialization.

- **Acceptance test refactor**
  - Replace hard-coded domain and service IDs with dynamically created services and domains.
  - Add full lifecycle coverage for `fastly_domain_service_link`: create → update (service relink) → import.
  - Update destroy verification to correctly assert domain unlinking behavior.

- **Plan drift prevention**
  - Use `lifecycle.ignore_changes = [service_id]` on `fastly_domain` in tests to prevent Terraform from attempting to reconcile `service_id` changes made indirectly by `fastly_domain_service_link`.

## Testing

Updated acceptance test covers:
- Initial domain-to-service linking
- Updating the link to a different service
- Importing the resource into state
- Proper cleanup and unlinking on destroy

```
=== RUN   TestAccFastlyDomainServiceLink_Basic
=== PAUSE TestAccFastlyDomainServiceLink_Basic
=== CONT  TestAccFastlyDomainServiceLink_Basic
--- PASS: TestAccFastlyDomainServiceLink_Basic (16.36s)
```

 All Submissions:

### New Feature Submissions:


fastly/resource_fastly_domain_service_link_test.go
```diff
@@ -5,79 +5,114 @@ import (
...
 
-const (
-	domainID   = "VrU9QvSHFBqvO3WJZoro0w"
-	serviceID  = "4Y3PRaauBiVcQXPOJUURj6"
-	serviceID2 = "iWBciPXoEl8PfUOW2hvIm4"
-)
-
-// TestAccFastlyDomainServiceLink_Basic tests resource creation from scratch (without import).
-// This ensures the Create → Read flow works correctly, as import can mask certain behaviors where
-// setting d.Id() before Read is called [CDTOOL-1198].
```
philippschulte, line 23: @rcaril in Terraform acceptance testing:

- Each test step is isolated
- Import behavior doesn’t affect Create or Update logic
- CreateContext sets the ID and calls ReadContext right after
- ImportState sets the ID and calls ReadContext separately

So the import cannot mask bugs in the Create → Read flow because it doesn’t run until after those steps finish.

philippschulte:
@rcaril I might need to update the docs to document the new import command for the domain management resource. I'll check this later today.

fastly/resource_fastly_domain.go
```diff
@@ -82,7 +82,7 @@ func resourceFastlyDomainRead(ctx context.Context, d *schema.ResourceData, meta
 		DomainID: gofastly.ToPointer(d.Id()),
 	}
 
-	data, err := domains.Get(gofastly.NewContextForResourceID(ctx, d.Get("service_id").(string)), conn, input)
+	data, err := domains.Get(ctx, conn, input) // ← no need for context with service_id
```
rcaril: Thanks for correcting this `service_id` related potential panic. Can we make the same changes for the Create and Delete operations?

rcaril (changes requested):
Looks good, one minor update requested.

fastly/resource_fastly_domain_service_link_test.go
rcaril: After doing some additional research, I concur - the test here would not test this given scenario nor would it be possible based on the interactions in `resource_fastly_domain_service_link.go`.

fastly/resource_fastly_domain.go
philippschulte: @rcaril as discussed in our 1:1 last week:

The `gofastly.NewContextForResourceID(...)` mechanism is designed to **serialize concurrent mutating requests** (e.g., Create, Update, Delete) targeting the same resource, such as a Service ID or Workspace. This helps prevent race conditions when modifying the same resource from different goroutines or clients.

However, in this specific case, the `domains.Get(...)` call is a **read-only (non-mutating)** operation. According to [go-fastly#715](https://github.com/fastly/go-fastly/pull/715), the resource-level locking mechanism only applies to **mutating operations**. Therefore, passing a resource-specific context is unnecessary here and can safely be omitted.

rcaril: Indeed, I thought I had followed up on this and replied.

rcaril (approved):
LGTM - thanks!

PR #1187, opened 2026-02-09
object_storage_access_key: properly set access_key_id

jedisct1:
### Change summary

`access_key_id` was being set using the secret key instead of the key identifier.

 All Submissions:

* [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/fastly/terraform-provider-fastly/pulls) for the same update/change?

* [ ] Have you written new tests for your core changes, as applicable?

### User Impact

The key identifier is now properly  returned.


rcaril (changes requested):
This looks good, but can we update the tests if the values of `access_key_id` and `secret_key` are different to ensure that we have coverage to catch this in the future?

rcaril (approved):
Thanks for adding the test!

anthony-gomez-fastly (approved):
LGTM, ty

PR #1189, opened 2026-02-11
[CDTOOL-1270] Add support for nested `multival` conditions in `group operators`

rcaril:
### Change summary

This PR adds supported for nested `multival` conditions inside of `group_operator` blocks. 

 All Submissions:

### New Feature Submissions:

```
=== RUN   TestFlattenNGWAFRuleResponse
--- PASS: TestFlattenNGWAFRuleResponse (0.00s)
PASS
---
=== RUN   TestAccFastlyNGWAFWorkspaceRule_emptyGroupCondition
=== PAUSE TestAccFastlyNGWAFWorkspaceRule_emptyGroupCondition
=== CONT  TestAccFastlyNGWAFWorkspaceRule_emptyGroupCondition
--- PASS: TestAccFastlyNGWAFWorkspaceRule_emptyGroupCondition (0.13s)
PASS
```

### Changes to Core Features:

### User Impact

N/A - more support is added.


docs/data-sources/domains_v1.md
```diff
@@ -0,0 +1,34 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "fastly_domains_v1 Data Source - terraform-provider-fastly"
+subcategory: ""
+description: |-
+  
+---
+
+# fastly_domains_v1 (Data Source)
```
rcaril, line 9: It looks like the generate-docs command appended some domain docs from the prior PR. I think it's fine to leave these updates here?

go.mod
```diff
@@ -6,7 +6,7 @@ toolchain go1.25.7
 
 require (
 	github.com/deckarep/golang-set/v2 v2.8.0
-	github.com/fastly/go-fastly/v12 v12.1.2
+	github.com/fastly/go-fastly/v12 v12.1.3-0.20260211144541-3a400219a0df
```
rcaril: We can re-update this after the go-fastly release.

fastly/ngwaf_rule_expand.go
```diff
@@ -206,65 +206,155 @@ func expandNGWAFRuleGroupCreateConditions(raw []any) []*rules.CreateGroupConditi
...
-			for i, v := range conds {
-				c[i] = v.(*rules.CreateCondition)
+	var groupConditions []*rules.CreateGroupCondition
+	for _, item := range raw {
+		m, ok := item.(map[string]any)
+		if !ok {
+			continue
+		}
+
+		groupOp := m["group_operator"].(string)
```
philippschulte: Those lookups can panic! Please use “comma ok” assertions and skip instead of crashing:

```
groupOp, ok := m["group_operator"].(string)
if !ok || groupOp == "" {
    continue
}
```

Please apply this idiom to all other map lookups.

fastly/ngwaf_rule_schema.go
```diff
@@ -95,14 +95,13 @@ func resourceFastlyNGWAFRuleBase() *schema.Resource {
...
-				Description: "List of grouped conditions with nested logic. Each group must define a `group_operator` and at least one condition.",
+				Description: "List of grouped conditions with nested logic. Each group must define a `group_operator` and at least one condition or multival_condition.",
 				Elem: &schema.Resource{
 					Description: "Group of conditions using logical operators.",
 					Schema: map[string]*schema.Schema{
 						"condition": {
 							Type:        schema.TypeList,
-							Required:    true,
-							MinItems:    1,
+							Optional:    true,
```
philippschulte, line 112: Now the schema changed to:

- `condition` is **optional**
- `multival_condition` is **optional**

But there’s no provider-side check that **at least one of them is set**.

That means users can write:

```
group_condition {
  group_operator = "all"
}
```

…and your expand will build a `CreateGroupCondition{Conditions:nil, MultivalConditions:nil}`. That flows into `go-fastly’`s marshaling logic and likely ends up as a `"conditions": [] group` — which the API will reject because to my knowledge an ngwaf rule needs at least one condition.

I think you need to implement a `CustomizeDiff` because right now it silently accepts empty groups.

philippschulte (changes requested):

go.mod
philippschulte: Please handle dependencies separately. Let's first merge dependabot's future go-fastly upgrade pull request first. Then you can update this pull request and resolve the merge conflicts. Thanks.

rcaril: Yup, this was just temporary to run the PR against the the Github actions.

fastly/ngwaf_rule_schema.go
rcaril: Indeed, I realized this but I figured we would rely on the API for validation like we do with some other features. I am okay with implementing a custom check - I just wasn't sure the direction we wanted to go here.

fastly/ngwaf_rule_expand.go
rcaril: Ah, I missed this - good catch, thanks!

rcaril: Fixed in [8baacbc](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/8baacbc0da58deafd4ace64a855a5743bb125f59).

fastly/ngwaf_rule_schema.go
rcaril: Added in [bfd51d6](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/bfd51d6fc86f386af443a40f789be497691425a6) as well as the test for this, `TestAccFastlyNGWAFWorkspaceRule_emptyGroupCondition` in `fastly/resource_fastly_ngwaf_workspace_rule_test.go`.

philippschulte: The added validation in [bfd51d6](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/bfd51d6fc86f386af443a40f789be497691425a6) checks group blocks, but not overall rule validity.

[bfd51d6](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/bfd51d6fc86f386af443a40f789be497691425a6) validates:

> each group_condition must contain at least one of condition or multival_condition

That’s correct — but the API also requires the rule to have at least one condition somewhere.

Right now, a config could potentially be:

- no condition
- no multival_condition
- no group_condition

…and still pass schema validation (and only fail at API time). The provider should validate that too.

A rule must define at least one of `condition`, `multival_condition`, or `group_condition` (non-empty).

This is separate from “group_condition must not be empty”.

Here is what I would do:

1. Remove `CustomizeDiff` from `resourceFastlyNGWAFRuleBase()` -> Keep base schema purely schema.
2. Keep `validateGroupConditionNotEmpty` only in the concrete resources:
    - workspace rule resource (already added)
    - account rule resource (if it exists and supports the same schema)
3. Add another validation `validateRuleHasAtLeastOneCondition` into the same `customdiff.All(...)` chain:
    - checks at least one of top-level blocks is non-empty:
        - `condition`
        - `multival_condition`
        - `group_condition`
4. Keep the acceptance test for empty group condition, and consider adding another test for “rule has zero conditions anywhere” if that case is currently possible in schema.

Possible implementation (centralize the diff validators and explicitly compose them in both resources):

Create a helper returning the common validators:
```
func ngwafRuleCommonCustomizeDiff() []schema.CustomizeDiffFunc {
  return []schema.CustomizeDiffFunc{
    validateRuleHasAtLeastOneCondition,
    validateGroupConditionNotEmpty,
  }
}
```

Then wire it like:

Workspace rule:
```
r.CustomizeDiff = customdiff.All(append(
  ngwafRuleCommonCustomizeDiff(),
  []schema.CustomizeDiffFunc{
    // templated_signal description validation
    func(...) error { ... },
    forceNewOnTemplatedSignalChange(r.Schema),
  }...,
)...)
```

Account rule:
```
r.CustomizeDiff = customdiff.All(ngwafRuleCommonCustomizeDiff()...)
```

This avoids relying on base inheritance for diff logic (because workspace overrides anyway). The PR’s validation is currently fragile by design unless it’s also wired into the workspace resource.

rcaril: I made some updates in [5818efa](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/5818efa5f4e92bc162a677a10442d5f3fb8e5f7a)[5818efa](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/5818efa5f4e92bc162a677a10442d5f3fb8e5f7a) - let's chat more around the location of the `CustomizeDiff`.

philippschulte: Thanks for the update — I walked through [5818efa](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/5818efa5f4e92bc162a677a10442d5f3fb8e5f7a)[5818efa](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/5818efa5f4e92bc162a677a10442d5f3fb8e5f7a) and there’s one important functional issue around the validation wiring (**very easy fix**):

In `resourceFastlyNGWAFRuleBase()` you added:

- `validateRuleHasConditions`
- `validateGroupConditionNotEmpty`

…but in `resourceFastlyNGWAFWorkspaceRule()` you assign `r.CustomizeDiff = customdiff.All(...)`, which replaces whatever the base resource set.

That means `validateRuleHasConditions` is not executed for workspace rules, so the new acceptance test `TestAccFastlyNGWAFWorkspaceRule_noConditions` will not reliably produce the expected error message (`rule must define at least one...`) at plan time.

Right now the workspace resource only includes `validateGroupConditionNotEmpty` (plus templated-signal validations), so “no conditions at all” can slip through validation and fail later (or not match the expected regex).

So you have two options to fix this:

1. Include `validateRuleHasConditions` in the workspace rule `CustomizeDiff` list (**not a good solution at all because you have to keep the child schema in sync with the base**):

```go
r.CustomizeDiff = customdiff.All(
    validateRuleHasConditions,
    validateGroupConditionNotEmpty,
    // templated_signal description check...
    // forceNewOnTemplatedSignalChange...
)
```

2. Preserve the base `CustomizeDiff` instead of overwriting it (**more robust / preferred solution**):

```go
r.CustomizeDiff = customdiff.All(
    r.CustomizeDiff, // keep base validations
    // workspace-specific validations
    func(...) error { ... }, // templated_signal check
    forceNewOnTemplatedSignalChange(r.Schema),
)
```
So you just have to remove `validateGroupConditionNotEmpty` from the workspace CustomizeDiff resource. This prevents future regressions where base validations silently stop running on workspace/account resources.

rcaril: Great catch! I've corrected this in [934f12c](https://github.com/fastly/terraform-provider-fastly/pull/1189/commits/934f12c382579cb6b26f0e182e12199b1308ade1) and preserved `CustomizeDiff`.

rcaril:
@philippschulte  - the code is refactored now. Please let me know if you have any additional thoughts - thanks!

philippschulte (approved):
Thank you!

PR #1194, opened 2026-02-23
Add a guide for adding a versionless domain to a service using a wildcard tls subscription

anthony-gomez-fastly:
### Change summary

Adds a guide with functional HCL for a Delivery service using a Wildcard TLS subscription and a versionless domain

 All Submissions:

### New Feature Submissions:

~~* [ ] Does your submission pass tests?~~
~~* [ ] Post the output of your test runs~~

### DIFF
--- a/docs/guides/versionless_domains.md
+++ b/docs/guides/versionless_domains.md
@@ -29,17 +29,13 @@ data "fastly_tls_subscription" "subscription" {
   certificate_authority = "certainly"
 }
 
-resource "fastly_domain" "domain_demo" {
-  fqdn = "example.fastlyversionlessdomain.com"
-}
-
-resource "fastly_service_vcl" "vcl_demo" {
-  name          = "demofastly"
+resource "fastly_service_vcl" "vcl_example" {
+  name          = "versionlessdomainexample"
   force_destroy = true
 }
 
-resource "fastly_domain_service_link" "link" {
-  domain_id  = fastly_domain.domain_demo.id
-  service_id = fastly_service_vcl.vcl_demo.id
+resource "fastly_domain" "domain_example" {
+  fqdn       = "example.fastlyversionlessdomain.com"
+  service_id = fastly_service_vcl.vcl_example.id
 }
 ```
--- a/docs/guides/versionless_domains.md
+++ b/docs/guides/versionless_domains.md
@@ -1,15 +1,11 @@
 ---
-page_title: versionless_domains
+page_title: versionless_domain_with_wildcard_tls_subscription
 subcategory: "Guides"
 ---
 
-## Versionless Domains
+## Creating a Delivery Service using an existing Wildcard TLS Subscription
 
-There are several different ways your Fastly services can use versionless domains
-
-## Delivery Service
-
-If you have an existing wildcard TLS subscription managed outside of Terraform, you can create a new subdomain in Terraform and then link it to a Delivery service
+If you have an existing wildcard TLS subscription managed outside of Terraform, you can create a new Delivery service in Terraform and then link it to a domain
 
 ```
 terraform {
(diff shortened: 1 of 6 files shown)


docs/guides/versionless_domains.md
```diff
@@ -0,0 +1,45 @@
...
+}
+
+resource "fastly_service_vcl" "vcl_demo" {
+  name          = "demofastly"
+  force_destroy = true
+}
+
+resource "fastly_domain_service_link" "link" {
+  domain_id  = fastly_domain.domain_demo.id
+  service_id = fastly_service_vcl.vcl_demo.id
```
kpfleming: You should be able to move `service_id` into the `fastly_domain` resource, there's no need to use the split resources for this configuration.

docs/guides/versionless_domains.md
```diff
@@ -0,0 +1,45 @@
+---
+page_title: versionless_domains
```
kpfleming: This will need to be a more expressive title which covers this specific scenario, so that the other scenarios can have their own pages - I'd like to avoid having all of the versionless domain examples in a single page.

kpfleming (dismissed):

docs/guides/versionless_domains.md
anthony-gomez-fastly: I was thinking that we'd group all the guides, but if we want to have individual ones then agree

kpfleming: It will be easier to link to individual scenarios if they are separate pages - that was my thinking anyway.

anthony-gomez-fastly: fixed in `7b21538b`

docs/guides/versionless_domains.md
anthony-gomez-fastly: fixed in `35879bc5`

docs/index.md
```diff
@@ -25,7 +25,7 @@ terraform {
   required_providers {
     fastly = {
       source  = "fastly/fastly"
-      version = ">= 8.7.0"
+      version = ">= 8.6.0"
```
kpfleming: oops?

templates/guides/versionless_domains.md
```diff
@@ -0,0 +1,37 @@
+---
+page_title: versionless_domain_with_wildcard_tls_subscription
+subcategory: "Guides"
+---
+
+## Creating a Delivery Service using an existing Wildcard TLS Subscription
+
+If you have an existing wildcard TLS subscription managed outside of Terraform, you can create a new Delivery service in Terraform and then link it to a domain
```
kpfleming:
```suggestion
If you have an existing wildcard TLS subscription managed outside of Terraform, you can create a new Delivery service in Terraform and then link it to a subdomain of the wildcard domain.
```

templates/guides/versionless_domains.md
```diff
@@ -0,0 +1,37 @@
+---
+page_title: versionless_domain_with_wildcard_tls_subscription
```
kpfleming: See https://github.com/fastly/terraform-provider-fastly/blob/55d82994836f822ef997bc2fa1cdb6a138dd264b/templates/guides/fastly_object_storage.md?plain=1#L2, these can actually be normal sentences and they show up nicely in the documentation site.
```suggestion
page_title: Delivery service with a versionless subdomain of a wildcard TLS subscription
```

kpfleming (dismissed):
A couple more tweaks and this is ready to go!

docs/index.md
anthony-gomez-fastly: idk why my branch is like this....

anthony-gomez-fastly: running the generate command makes this change, interesting

anthony-gomez-fastly: ah, so i think that there was an issue with the last release, because the last tag isn't 8.7
Screenshot 2026-02-23 at 12 16 11 PM

kpfleming (approved):

PR #1196, opened 2026-02-23
Add guide for migrating delivery service classic domain to a versionless domain

anthony-gomez-fastly:
### Change summary

Add docs for migrating classic domains attached to a service to versionless domains
 
All Submissions:

### DIFF
--- a/docs/guides/versionless_domain_migration.md
+++ b/docs/guides/versionless_domain_migration.md
@@ -18,7 +18,7 @@ resource "fastly_service_vcl" "vcl_example" {
 }
 ```
 
-Once you [use the control panel to migrate this domain to a versionless domain](https://www.fastly.com/documentation/guides/getting-started/domains/working-with-domains/migrating-classic-domains/), you will need to update your HCL to look something like this.
+Once you [use the control panel to migrate this domain to a versionless domain](https://www.fastly.com/documentation/guides/getting-started/domains/working-with-domains/migrating-classic-domains/), you will need to update your HCL to match the pattern below.
 You can find the documentation on versionless domain patterns here: https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/domain. Do not apply your changes before running the import step below, as that will result in an error.
 
 ```
@@ -32,7 +32,7 @@ resource "fastly_domain" "domain_example" {
 }
 ```
 
-Before making other changes you will need to import domain using terraform. The Domain ID can using the Fastly CLI by running `fastly domain list --fqdn=foo.example.com` and using the Domain ID fromt the record.
+Before making other changes you will need to import domain using terraform. The Domain ID can using the Fastly CLI by running `fastly domain list --fqdn=foo.example.com` and using the Domain ID from the record.
 
 ```
 terraform import fastly_domain.domain_example YOUR_DOMAIN_ID
--- a/docs/guides/versionless_domain_migration.md
+++ b/docs/guides/versionless_domain_migration.md
@@ -32,7 +32,7 @@ resource "fastly_domain" "domain_example" {
 }
 ```
 
-Before making other changes you will need to import domain using terraform. The Domain ID can using the Fastly CLI by running `fastly domain list --fqdn=foo.example.com` and using the Domain ID from the record.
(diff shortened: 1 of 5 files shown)


templates/guides/versionless_domain_migration.md
```diff
@@ -0,0 +1,57 @@
...
+  name = "versionlessdomainexample"
+  domain {
+    name    = "demo.example.com"
+    comment = "demo"
+  }
+  force_destroy = true
+}
+```
+
+Once you [use the control panel to migrate this domain to a versionless domain](https://www.fastly.com/documentation/guides/getting-started/domains/working-with-domains/migrating-classic-domains/), you will need to update your HCL to look something like this.
```
rcaril: Does it make sense to link to the versionless domain Terraform provider page here as well for a reference?

rcaril (approved):
Looks good to me, but we may want to add a link to the existing docs for reference on how to build using the new versionless domain resource.

templates/guides/versionless_domain_migration.md
anthony-gomez-fastly: sure! i'll add that link

anthony-gomez-fastly: added in `ff17189f`

docs/guides/versionless_domain_migration.md
```diff
@@ -0,0 +1,66 @@
...
+  domain {
+    name    = "demo.example.com"
+    comment = "demo"
+  }
+  force_destroy = true
+}
+```
+
+Once you [use the control panel to migrate this domain to a versionless domain](https://www.fastly.com/documentation/guides/getting-started/domains/working-with-domains/migrating-classic-domains/), you will need to update your HCL to look something like this.
+[Here are the versionless domain docs.](https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/domain)
```
rcaril: I think this phrasing is fine, but I would also considering phrasing this as "You can find the documentation on versionless domain patterns here:" (or something similar to this)

docs/guides/versionless_domain_migration.md
```diff
@@ -0,0 +1,66 @@
...
+  force_destroy = true
+}
+
+resource "fastly_domain" "domain_example" {
+  fqdn       = "demo.example.com"
+  service_id = fastly_service_vcl.vcl_example.id
+}
+```
+
+Before making other changes you will need to import domain using terraform.
```
kpfleming: This will need to show the user how they can obtain the domain ID.

rcaril: Does it make sense to move this instruction above the HCL example?

anthony-gomez-fastly: I was thinking that since you need to set up the resource first, it makes sense to modify the HCL, then give the import command

anthony-gomez-fastly: Is there a nicer way of doing that than just grabbing it from the URL?

rcaril: You would have to use a Terraform expression to get it from the `fastly_domain` data source.

rcaril: Fair enough!

anthony-gomez-fastly: i'll change this to add a step afterwards to actually add the service id, makes the flow a little smoother

anthony-gomez-fastly: added a line to use the CLI

docs/guides/versionless_domain_migration.md
```diff
@@ -0,0 +1,69 @@
...
+  name = "versionlessdomainexample"
+  force_destroy = true
+}
+
+resource "fastly_domain" "domain_example" {
+  fqdn       = "demo.example.com"
+}
+```
+
+Before making other changes you will need to import domain using terraform. The Domain ID can using the Fastly CLI by running `fastly domain list --fqdn=foo.example.com` and using the Domain ID from the record.
```
rcaril: If we can make the `Fastly CLI` text a link to our CLI doc, I think that would be a perfect touch here. (https://www.fastly.com/documentation/reference/tools/cli/)

rcaril (approved):
Solid updates, one more minor item I think would help for those that may not use the CLI normally.

docs/guides/versionless_domain_migration.md
anthony-gomez-fastly: done in `ff1bb948`

PR #1221, opened 2026-04-02
Breaking[CDTOOL-1309]  Add support for ContentGuard

rcaril:
### Change summary

This PR adds support for ContentGuard to the Bot Management Product Enablement block. 

 All Submissions:

### New Feature Submissions:

```
=== RUN   TestAccFastlyServiceProductEnablement_vcl_botManagementUpdate
=== PAUSE TestAccFastlyServiceProductEnablement_vcl_botManagementUpdate
=== CONT  TestAccFastlyServiceProductEnablement_vcl_botManagementUpdate
--- PASS: TestAccFastlyServiceProductEnablement_vcl_botManagementUpdate (41.55s)
PASS
```

### Changes to Core Features:

### User Impact

Users will now need to supply the `contentguard` and `enabled` parameters for the `product_enablement.bot_management` block.

### DIFF
--- a/fastly/block_fastly_service_product_enablement.go
+++ b/fastly/block_fastly_service_product_enablement.go
@@ -68,9 +68,29 @@ func (h *ProductEnablementServiceAttributeHandler) GetSchema() *schema.Schema {
 	// These products are supported only on Delivery (VCL) services.
 	if h.GetServiceMetadata().serviceType == ServiceTypeVCL {
 		blockAttributes["bot_management"] = &schema.Schema{
-			Type:        schema.TypeBool,
+			Type:        schema.TypeList,
 			Optional:    true,
 			Description: "Enable Bot Management support",
+			MaxItems:    1,
+			MinItems:    1,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"contentguard": {
+						Type:        schema.TypeString,
+						Required:    true,
+						Description: "ContentGuard status. Can be either `off`, or `on`.",
+						ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(
+							[]string{"off", "on"},
+							false,
+						)),
+					},
+					"enabled": {
+						Type:        schema.TypeBool,
+						Required:    true,
+						Description: "Enable Bot Management support",
+					},
+				},
+			},
 		}
 		blockAttributes["brotli_compression"] = &schema.Schema{
 			Type:        schema.TypeBool,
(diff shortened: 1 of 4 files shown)


rcaril:
Confirmed with local TF testing that there are no issues.

fastly/block_fastly_service_product_enablement.go
```diff
@@ -192,11 +212,25 @@ func (h *ProductEnablementServiceAttributeHandler) Create(ctx context.Context, d
...
+		bp := resource["bot_management"].([]any)
+		if len(bp) != 0 {
+			if bp[0].(map[string]any)["enabled"].(bool) {
+				log.Println("[DEBUG] bot_management set")
+				_, err := botmanagement.Enable(gofastly.NewContextForResourceID(ctx, d.Id()), conn, serviceID)
+				if err != nil {
+					return fmt.Errorf("failed to enable bot_management: %w", err)
+				}
+
+				contentguard := bp[0].(map[string]any)["contentguard"].(string)
```
anthony-gomez-fastly, line 224: are these resources always paired this way? as in if `enabled` is set `contentguard` will also be set?

anthony-gomez-fastly: looks like the resource block is designed that way at least. not sure if line 217 is needed if so?

rcaril: `contentguard` is now a required field - where as `enabled` was always required prior for these types of Product Enablement blocks. This is a similar pattern that we have implemented for the `ngwaf` block here:
https://github.com/fastly/terraform-provider-fastly/blob/ef2783457e15f73e07ba1c7b61095578014a5829/fastly/block_fastly_service_product_enablement.go#L265

anthony-gomez-fastly: makes sense!

anthony-gomez-fastly (approved):
LGTM

fastly/block_fastly_service_product_enablement.go
```diff
@@ -338,7 +372,21 @@ func (h *ProductEnablementServiceAttributeHandler) Read(ctx context.Context, d *
...
+
+				bp := []map[string]any{}
+				bp = append(bp, map[string]any{
+					"enabled":      true,
+					"contentguard": *c.Configuration.ContentGuard,
+				})
+
+				result["bot_management"] = bp
+			} else if len(localState) > 0 {
+				bp := localState[0].(map[string]any)["bot_management"].([]any)
```
philippschulte: This assumes `localState[0]["bot_management"]` always exists and is `[]any`. If the block is omitted from config, that lookup can be `nil`, and the type assertion can panic. Please guard the map lookup before asserting.

The current acceptance tests do not cover this case because they always include a bot_management block, even in the “basic” test where it is disabled.

fastly/block_fastly_service_product_enablement.go
```diff
@@ -68,9 +68,29 @@ func (h *ProductEnablementServiceAttributeHandler) GetSchema() *schema.Schema {
...
+							false,
+						)),
+					},
+					"enabled": {
+						Type:        schema.TypeBool,
+						Required:    true,
+						Description: "Enable Bot Management support",
+					},
+				},
+			},
```
philippschulte, line 93: Before this PR, `bot_management` was a boolean, so moving from `true` to `false` clearly disabled the product. After this schema change, `Update()` only handles the case where the block exists and `len(bp) != 0`. If a user removes the block entirely after previously enabling Bot Management, this code path appears to do nothing, which could leave the product enabled remotely even though it was removed from config.

I think we need to explicitly handle the “block removed” case and disable the product when appropriate (enabled block → block removed entirely). As written, this looks like it would cause drift on the next refresh/plan.

philippschulte (changes requested):
Requesting changes here. The overall direction makes sense, but I think this PR introduces a regression in how `bot_management` is disabled/removed after changing it from `bool` to nested block.

I think the acceptance coverage for Bot Management still needs a few more transition cases, especially because this PR changes the shape of the field rather than just adding a new attribute.

The current test covers:

- block present with `contentguard = "off"`
- block present with `contentguard = "on"`

But it does not cover:

- enabled block -> block removed entirely
- `enabled = true` -> `enabled = false`
- config with no `bot_management block` at all -----> `Read()`

I understand other nested products don’t have this coverage either, but we should add those too. For this PR in particular, these cases seem important for validating the new behavior and would likely catch the edge cases above.

fastly/block_fastly_service_product_enablement.go
rcaril: Fixed in [76bc5d9](https://github.com/fastly/terraform-provider-fastly/pull/1221/commits/76bc5d9956ed373f2eb05c00d4ad81048c56f7a4). We removed these fallbacks as we should only read products that are enabled remotely. For the purpose of entitlement to products, we shouldn't be storing a product to state if a user doesn't have access anyway.

fastly/block_fastly_service_product_enablement.go
rcaril: I tested locally, and removing the block does call the DELETE operation - so this is working as expected. However, adding a test for the block removal isn't possible at this time - the full reasoning is documented [here](https://github.com/fastly/terraform-provider-fastly/blob/ef2783457e15f73e07ba1c7b61095578014a5829/fastly/block_fastly_service_product_enablement.go#L671). (A rewrite of this resource should be done in the new provider instead.)

philippschulte (approved):
Thanks!

PR #1233, opened 2026-04-16
[CDTOOL-1318] Add support for backend connection reuse limits

rcaril:
### Change summary

This PR adds support for the `MaxLifetime` and `MaxUse` backend attributes for both VCL and Compute services. 

 All Submissions:

### New Feature Submissions:

```
=== RUN   TestResourceFastlyFlattenBackend
--- PASS: TestResourceFastlyFlattenBackend (0.00s)
=== RUN   TestResourceFastlyFlattenBackendCompute
--- PASS: TestResourceFastlyFlattenBackendCompute (0.00s)
PASS
```

### Changes to Core Features:


rcaril:
Confirmed that these updates work locally as well.

philippschulte (changes requested):
The schema and flattening look generally right, but I think there is a blocking issue in create/update handling.

The Fastly API docs state that `0` is a valid value for both fields and means unlimited. However, the provider currently only sends these fields when the value is `> 0`. That means Terraform cannot clear/reset an existing non-zero value back to `0`; the field will be omitted from the update request and the remote value will remain unchanged, likely causing a persistent diff.

It would also be good to add tests for resetting `max_lifetime` and `max_use` from non-zero values back to `0`, since the current tests only cover flattening non-zero API values.

Also please consider adding validation to reject negative values, since negative connection reuse limits do not appear valid and are currently silently skipped by the > 0 checks.

rcaril:
> The schema and flattening look generally right, but I think there is a blocking issue in create/update handling.
> 
> The Fastly API docs state that `0` is a valid value for both fields and means unlimited. However, the provider currently only sends these fields when the value is `> 0`. That means Terraform cannot clear/reset an existing non-zero value back to `0`; the field will be omitted from the update request and the remote value will remain unchanged, likely causing a persistent diff.
> 
> It would also be good to add tests for resetting `max_lifetime` and `max_use` from non-zero values back to `0`, since the current tests only cover flattening non-zero API values.
> 
> Also please consider adding validation to reject negative values, since negative connection reuse limits do not appear valid and are currently silently skipped by the > 0 checks.

[844bbdc](https://github.com/fastly/terraform-provider-fastly/pull/1233/commits/844bbdcc5361083a29c54e8f5b41e00d840e2e67) has the following updates: 

- Added a `validation.IntAtLeast(0)` validator
- Added a test to set `max_lifetime` and `max_use` from non-zero values back to 0
- Removed the `Computed` parameter and replaced this with a default of `0` instead
- Removed the condition wrapping these attributes that only allowed a value greater of 0 (not needed anymore as we have validation built into the schema) 

I believe these updates cover the gaps you pointed out.

ulyssa (approved):

philippschulte (changes requested):
Thanks for the update. This resolves my previous concern.

However your latest changes introduce another concern: the schema now uses `Optional + Default: 0` instead of `Optional + Computed`, which means omitted config is treated as unlimited. To me this is a breaking change!

If users do not add these attributes to config, Terraform may now plan to reset them to `0`.

Could we keep these as `Optional + Computed` with `IntAtLeast(0)` validation, and only send the values when explicitly configured or modified? The update path already handles the important reset-to-zero case correctly by sending modified values even when the value is `0`.

rcaril:
> Thanks for the update. This resolves my previous concern.
> 
> However your latest changes introduce another concern: the schema now uses `Optional + Default: 0` instead of `Optional + Computed`, which means omitted config is treated as unlimited. To me this is a breaking change!
> 
> If users do not add these attributes to config, Terraform may now plan to reset them to `0`.
> 
> Could we keep these as `Optional + Computed` with `IntAtLeast(0)` validation, and only send the values when explicitly configured or modified? The update path already handles the important reset-to-zero case correctly by sending modified values even when the value is `0`.

I believe I wound up reverting computed due to the explanation of how the defaults worked [here](https://fastly.slack.com/archives/C09DJJJN37T/p1776351449563709). However, upon further review - you are right. We need to maintain these as Computed as the API doesn't actually set these values to `0` when unused. 

Updated in [497015e](https://github.com/fastly/terraform-provider-fastly/pull/1233/commits/497015e3e2ce6975b9e1004a3bf88fddb6266d27).

philippschulte (changes requested):
The update path looks correct, and the schema direction is correct with `Optional + Computed`.

However, the create path still needs to avoid sending `max_lifetime` / `max_use` when those attributes are omitted from config.

I tested this locally against the PR branch with the minimal backend acceptance test and go-fastly debugging enabled.

The test confirms that when `max_lifetime` and `max_use` are omitted from config, the create request still sends them as explicit zero values:

```text
max_lifetime=0&max_use=0
```

The API then persists and returns them as explicit `0` values:

```json
"max_lifetime": 0,
"max_use": 0
```

So the concrete issue here is in the create path: newly created backends that omit these fields no longer preserve the API’s unset/null state, and are instead created with explicit `0` values.

This is still a behavior change for newly created backends, and it may also affect any existing config that causes a backend to be recreated.

Since these fields are modeled as `Optional + Computed` and the API can represent them as unset/null when unused, create should preserve that behavior unless the user explicitly configures the attributes.

rcaril:
After further review the API behavior as well as chatting with internal folks on this - I have reverted the code back to using `0` as a default in [9d47dc3](https://github.com/fastly/terraform-provider-fastly/pull/1233/commits/9d47dc3162ec7be5728e95e799fbd4895bed5bda). 

The important clarification here is that when the values of `max_lifetime` and `max_use` are unset, they are considered unlimited - which is the same behavior as setting the value of these to `0`. You can follow the internal discussion on this [here](https://fastly.slack.com/archives/C09DJJJN37T/p1778268449335749?thread_ts=1778257012.279849&cid=C09DJJJN37T). 

From local testing I can see that when a new backend is created, the state is set to `0` for both  `max_lifetime` and `max_use`, which aligns with other backend _int_ attributes, such as `keepalive_time`.  This also applies for any update operations, as we do a Destroy + Recreate for `Backend` blocks. 

With all of this being said, setting these back to `null` is a moot point since the API behavior is identical as when the value are set to `0` - which further supports why the default value for `max_lifetime` and `max_use` should be left as `0`.

philippschulte (approved):
I pulled the latest PR branch and tested the upgrade path locally.

Test performed:

1. Applied a minimal backend config with the current `main` provider, omitting `max_lifetime` and `max_use`.
2. Confirmed the existing state did not contain either field.
3. Switched to the latest PR branch, rebuilt the provider, and ran `terraform plan` against the same config and state.

The normal plan was clean:

```text
No changes. Your infrastructure matches the configuration.
```

I also ran a refresh-only plan. Terraform reported a backend block state change, likely because the new defaulted fields are now being normalized into state, but the normal plan does not require an update or recreate.

Given the team has confirmed that unset and 0 are functionally equivalent, and the normal upgrade plan is clean, I’m okay with the `Default: 0` approach.

PR #1253, opened 2026-04-22
Check if the pr relates to go.mod and if it doesn't, add the skip changelog label and ignore

anthony-gomez-fastly:
### Change summary
We don't need to add changelog entries for tools and GH actions dependabot updates, this PR adds steps that short circuit the changelog

 All Submissions:

### New Feature Submissions:


.github/workflows/dependabot_changelog_update.yml
```diff
@@ -13,33 +14,75 @@ jobs:
     if: github.actor == 'dependabot[bot]'
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout code (for path check)
```
rcaril: This looks good - I am wondering if checking for the `tools` label instead makes this any simpler? No action needed, was just thinking out loud here.

rcaril (approved):

.github/workflows/dependabot_changelog_update.yml
anthony-gomez-fastly: great idea, the other workflow wasn't using the label so i added that as well

.github/workflows/dependabot_changelog_update.yml
```diff
@@ -13,33 +14,72 @@ jobs:
...
+            }
+
+            console.log('Proceeding with changelog update');
+            core.setOutput('continue', 'true');
+      - name: Add Skip-Changelog label
+        if: steps.check-paths.outputs.continue == 'false'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.addLabels({
```
rcaril: I think we need an `await` here to ensure the label is set? 


```suggestion
           await github.rest.issues.addLabels({
```

rcaril (approved):

PR #1265, opened 2026-05-13
Bootstrap dual-model framework rewrite docs and examples

philippschulte:
### Change summary

This PR bootstraps the `dual-model-framework-rewrite` branch with the initial documentation and examples for the Terraform Plugin Framework rewrite.

It adds:

- the initial README for the rewrite branch
- the dual-model provider design document
- Terraform query documentation
- initial compatibility and explicit workflow examples

### Rewrite branch workflow

`dual-model-framework-rewrite` is intended to be the clean development line for the next major version of the Fastly Terraform provider.

Going forward, rewrite work should happen through feature PRs targeting `dual-model-framework-rewrite`, not `main`.

Example workflow:

1. branch from `dual-model-framework-rewrite`
2. implement a focused piece of the rewrite
3. open a PR back into `dual-model-framework-rewrite`
4. repeat until the rewrite is complete and stable

The current `main` branch remains the existing provider line while the rewrite is in progress. Once the rewrite is complete, we can decide how to promote `dual-model-framework-rewrite` as the new main development line while keeping the current provider line available for maintenance as needed.

### Reviewer notes

The most important files to review are:

1. `docs/dual-model-provider-design.md`
2. `docs/terraform-query.md`

The examples are included to show how the two resource families are expected to be used, but the main design decisions are captured in the two documents above.


philippschulte:
One thing I would like us to decide as part of this initial docs/examples PR is resource naming.

One of the reasons I wanted this first PR to contain only the README, design docs, and examples is so we can settle naming before we start adding a lot of implementation.

My current proposal keeps `fastly_service_vcl` as the compatibility resource because that is the closest continuation of the current provider’s nested-resource model, and puts the suffix on the new first-class resource family with `*_explicit`.

That said, I’m happy to update the naming based on team/PM feedback.

The main options I see are:

### Option 1: keep current-provider-style naming for compatibility

```text
fastly_service_vcl
fastly_service_compute

fastly_service_vcl_explicit
fastly_service_compute_explicit
```

This keeps the compatibility resource closest to the current provider naming.

### Option 2: make lifecycle model explicit on both families

```text
fastly_service_vcl_auto
fastly_service_compute_auto

fastly_service_vcl
fastly_service_compute
```

This gives the “clean” names to the explicit resource family, but makes the compatibility lifecycle behavior more visible.

### Option 3: use product-oriented service names

```text
fastly_service_cdn
fastly_service_compute
```

Potentially with explicit variants such as:

```text
fastly_service_cdn_explicit
fastly_service_compute_explicit
```

This aligns better with current Fastly product language: CDN and Compute, rather than API implementation terms like VCL/WASM.

If the team prefers `fastly_service_cdn` / `fastly_service_compute`, I’m happy to update the docs and examples before we start implementing the resources.

jaskiratr:
`fastly_service_cdn` / `fastly_service_compute`. Fastly docs use CDN and Compute as the correct terms. Let's align with that.

**Naming feedback on the `_explicit` suffix:**
I don't think `_explicit` communicates anything meaningful to developers. They can't tell what "explicit" refers to without reading docs.

Can we use `_versioned` instead? `fastly_service_domain_versioned` signals that the resource requires a version pin.

The compatibility family names (`fastly_service_cdn`, `fastly_service_compute`) stay as-is since they don't need a suffix.

-- 
Separately, when we deprecate `fastly_service_vcl` in favor of `fastly_service_cdn`, `_vcl` should continue to work until the next major version release (or something like 6 months). The provider will surface deprecation warnings to give users time to migrate.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+| ----------------------------------- | ---------------- | ----------------- | ----------------- | ------------------- |
+| Compatibility: `fastly_service_vcl` | Provider-managed | Provider-managed  | Automatic         | Not supported       |
+| Explicit: `*_explicit`              | Caller-managed   | Caller-managed    | Manual / explicit | Supported only here |
+
+The provider should not support intermediate variants:
+
+- compatibility + manual activation
+- explicit resources + automatic activation
+
+There is intentionally no `activate` field on `fastly_service_vcl`. The
```
rcaril: Is this a limitation of the the design changes? This seems like another high friction point.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+activate selected version automatically
+```
+
+### Draft selection order for compatibility updates
+
+When `fastly_service_vcl` needs a working version for an update, it should use
+this selection order:
+
+1. If the service has an active version, clone the active version.
+2. Otherwise, if the service has exactly one version and that version is
```
rcaril: I think it would make more sense to perform a clone step here. We don't want to overwrite an unknown state of an existed drafted version. We lose a versioned history by using an existing draft as is. This also goes against our general best practices for making changes services.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+All explicit versioned resources for service S write directly to vN
+  |
+  v
+User or pipeline activates or stages vN explicitly
+```
+
+### Notes
+
+- In explicit mode, the provider does not choose the target version.
+- The caller is responsible for ensuring the chosen version is writable.
```
rcaril: > - In explicit mode, the provider does not choose the target version.
> - The caller is responsible for ensuring the chosen version is writable.

For customers that aren't using full CI/CD pipelines, this is creating a lot of extra work for a given user. If we are set on using this pattern, we should provide a specific data source that provides both the active service version as well as latest version so non pipeline users can still get the information they need to perform a `terraform apply` without having to use several tools or CLI commands.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+
+## Activation
+
+### Automatic activation
+
+Automatic activation is supported only by the compatibility resource family.
+
+For the current VCL implementation, that means:
+
+- `fastly_service_vcl`
```
rcaril: I assume we'll also include Compute here?

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+
+### Manual activation
+
+Manual or explicit activation is supported only by the explicit resource family.
+
+It is better suited to:
+
+- controlled rollout order
+- pre-production validation
+- service chaining
```
rcaril: How is service chaining impacted with manual activation? Are we saying that the changes to a given service impact another? If so, one would argue that these changes would be made in on a non-production service first - so this may be a moot point.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+
+Automatic activation must not be combined with staging.
+
+### Manual activation
+
+Manual or explicit activation is supported only by the explicit resource family.
+
+It is better suited to:
+
+- controlled rollout order
```
rcaril: I would imagine customers have isolated Terraform configurations for this reason.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+
+```bash
+terraform import fastly_service_vcl.example <service_id>
+```
+
+During read/import, `fastly_service_vcl` selects the version to read from using
+this order:
+
+1. If the service has an active version, read from the active version.
+2. If the service has no active version, read from the latest service version.
```
rcaril, line 621: Is there a specific reason why we aren't allowing a user to specify a version here? If a user has a number of draft versions from prior iterations, it's very possible that they do not want to use the latest draft.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+
+A Fastly service is expected to have at least one version because service
+creation creates version `1`.
+
+Generated explicit resources include the version number that was read.
+
+If that version is active or locked, the generated configuration is still useful
```
rcaril, line 665: `terraform query` alone doesn't generate a configuration. We may want to be clear that a file can be generated with an additional parameter and that resulting file that will still need cleanup.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,507 @@
...
+  |
+  | active version exists?
+  |---- yes ---> read domains/backends from active version
+  |
+  |---- no ----> read domains/backends from latest version
+  v
+generate *_explicit resources with version pinned to the version read
+  |
+  v
+no clone, no activation, no staging, no mutation
```
rcaril: This may be a given considering the scope of what this feature does.

docs/dual-model-provider-design.md
philippschulte: Good question. I would not describe this as a hard API limitation, but as an intentional design boundary for the rewrite.

The dual-model provider is meant to support two clear workflows:

1. **Compatibility model**: nested resources with provider-managed clone, validate, and automatic activation.
2. **Explicit model**: first-class resources where clone, activation, staging, and rollout timing are controlled explicitly by the caller or workflow automation.

The goal is to avoid the ambiguous middle states that made the hybrid approach difficult to reason about:

- compatibility resources with provider-managed cloning but manual activation
- explicit first-class resources with provider-managed automatic activation

If users want current-provider-style convenience, they use the compatibility resource family. If they want control over activation, staging, or rollout order, they use the explicit resource family.

One additional reason I support removing the old `activate` field is that it is provider-only metadata. It is not part of the Fastly API and cannot be discovered during import. That means generated configuration would require manual cleanup or adjustment after import. Removing it makes the resource behavior clearer: `fastly_service_vcl` always owns the automatic lifecycle, and users who do not want that behavior should use the explicit resources instead.

So yes, this is a tradeoff, but it is intentional. The design tries to reduce friction by making each resource family internally consistent rather than supporting partially automatic lifecycle combinations.

kpfleming: Yes, this is intentional. If you don't want the cloned version to be automatically activated, don't use the 'automatic' workflow.

docs/dual-model-provider-design.md
philippschulte: I agree that we should not reuse an arbitrary existing draft of unknown origin. But I think point 2 is intentionally narrower than that.

The rule only applies when all of the following are true:

- the service has no active version
- the service has exactly one version
- that one version is unlocked

That is the brand-new / never-activated service case. Fastly creates version `1` when the service is created, and if that version has not been activated yet, it is still the only editable service state. In that case, using version `1` directly seems correct and matches the current-provider workflow I tested locally with `activate = false`: create the service, write config to draft version `1`, then continue applying changes to version `1` until it is activated.

I agree with your concern for services that have multiple drafts or unknown draft history. That is why point 3 says that if there is no active version and we are not in the single-unlocked-version case, we clone the latest existing version instead of mutating an arbitrary draft directly.

So I think the current selection order is the right distinction:

1. active version exists → clone active
2. no active version + exactly one unlocked version → use that version directly
3. otherwise → clone latest existing version

I’m happy to clarify the wording in the doc to make it explicit that point 2 is meant only for the brand-new / never-activated single-version case, not as general draft reuse.

docs/dual-model-provider-design.md
philippschulte: Yes, agreed that explicit-mode users should have provider support for discovering active/latest versions and should not need to stitch everything together manually.

A couple of clarifications:

First, the explicit workflow already includes `data.fastly_service_version`, which is intended to expose service version information such as the active/latest version. The Actions-based examples use that data source so users can stay inside Terraform for clone/activation workflows instead of relying only on external CLI commands.

Second, “the caller is responsible for ensuring the chosen version is writable” does not mean the provider should blindly attempt the write. The version-agnostic PoC included a mutability check for explicit resources, and I expect the rewrite to keep that behavior. Before writing to a pinned version, the provider can inspect the target version and return a clear diagnostic if it is active or locked.

So the model is:

- the caller chooses the target version
- the provider can expose active/latest version information through `data.fastly_service_version`
- the provider validates that the chosen target version is writable before modifying it
- clone/activate/stage remain explicit lifecycle operations, done through Terraform Actions, the Fastly CLI, or workflow automation

The key distinction is that explicit resources do not silently choose, clone, or activate versions during normal resource CRUD. But that does not mean the provider gives users no help; it should provide discovery data sources, explicit lifecycle actions, and clear validation errors.

docs/dual-model-provider-design.md
philippschulte: Yes, that needs to be added. Thanks!

docs/dual-model-provider-design.md
rcaril: ~I don't agree here. For an update specific workflow, we should not assume that a given service with only 1 draft version has no existing changes on that given version.~

~For newer Fastly customers, it's very feasible that changes have been staged on a single draft version as they begin to on-board. Another key point is that there is no way to roll back nor history of specific changes made other then high level audit logs if there in fact were some real changes on the first draft.~

~To your point about the create step, I agree - for a newly created service we should in fact be on version `1`. However, I don't know that we can safely assume that this initial version will always be unmodified in the scope of an update operation.~

docs/dual-model-provider-design.md
rcaril: Perfect, thanks for clarifying!

docs/dual-model-provider-design.md
philippschulte: Isolated Terraform configurations can help, but they do not fully solve this problem.

The point here is about separating “write the configuration changes to a draft” from “promote those changes to production.” Even if customers split services across Terraform configurations, automatic activation still couples the apply step with production promotion for each changed service.

With the explicit resource family, users can apply configuration changes to pinned draft versions first, review/validate them, and then activate services in the exact order required by their rollout process. That can be useful whether the services live in one Terraform configuration or several isolated ones.

docs/dual-model-provider-design.md
philippschulte: The concern is not that a Terraform change to one service directly mutates another service.

The concern is that traffic can flow through multiple Fastly services, and coordinated changes across chained services may need to be promoted in a specific production order. For example, if both the first service in the chain and a downstream service need related changes, automatic activation gives limited control over which production service is promoted first.

I agree that these changes should be validated in a non-production environment first, but staging/non-production validation does not replace production rollout control. The explicit resource family keeps the write step and activation step separate, so users or pipelines can decide the final activation order service by service.

docs/dual-model-provider-design.md
philippschulte: I do not think we should add user-selected import versions to `fastly_service_vcl` right now.

A few reasons:

1. This is not supported by the current `fastly_service_vcl` resource either. Import is service-based, not `service_id + version` based, and I am not aware of customer feedback asking for arbitrary draft-version import for the compatibility resource.

2. `fastly_service_vcl` is the compatibility / automatic lifecycle resource. Its behavior should stay simple: read the active production version when one exists, otherwise fall back to the latest version for services that have not been activated yet.

3. Allowing arbitrary version selection during import would introduce another lifecycle knob into the compatibility resource family. That starts to blur the line between compatibility mode and explicit mode.

4. If a user needs to work with a specific draft version, that is better handled by the explicit resource family, where `version` is part of the resource model and the user intentionally controls the target version.

So my recommendation is to keep compatibility import behavior deterministic:

- active version if one exists
- latest version if no active version exists

That keeps import aligned with the current provider behavior and with the compatibility model’s goal: represent the service’s current effective configuration, not expose arbitrary draft selection.

I’m open to documenting this more clearly, but I would not add `version` selection to `fastly_service_vcl` import unless we have a concrete customer need for it.

docs/dual-model-provider-design.md
philippschulte: We do have a dedicated `docs/terraform-query.md` that explains the actual command usage, including `terraform query -generate-config-out=...`, so I do not want to duplicate the full workflow here.

But I agree this section should be precise. `terraform query` by itself performs discovery; generated configuration requires running it with `-generate-config-out`.

I can update this wording to something like:

> Because these resources are first-class, `terraform query` is the right mechanism for discovery. When used with `-generate-config-out`, it can generate configuration and import blocks for the explicit resource family.

The most important behavior for this design doc is that the generated explicit resources include the version number that was read. If that version is active or locked, users need to clone or select a writable version and update the version pin before making changes with the explicit resources.

docs/dual-model-provider-design.md
philippschulte: Agreed that this is mostly implied by `terraform query` being a discovery operation.

I included the note because this design doc is specifically comparing query behavior against the compatibility resource family, where normal CRUD can clone and activate versions. So I think it is still useful to make the distinction explicit.

I’ll simplify the wording in the diagram from:

`no clone, no activation, no staging, no mutation`

to:

`read-only discovery; no version lifecycle operations`

rcaril: Sounds good!

docs/dual-model-provider-design.md
rcaril: I think the suggested update is perfect! 

> The most important behavior for this design doc is that the generated explicit resources include the version number that was read. If that version is active or locked, users need to clone or select a writable version and update the version pin before making changes with the explicit resources.

Agreed, I just wanted to make sure we were a bit more clear on the created config.

docs/dual-model-provider-design.md
anthony-gomez-fastly: I think there's a lot of potential for harm in not cloning, yes people aren't using best practices by continually modifying a single unactivated version but that's not a reason to blow away their stuff. it also seems kind of arbitrary to build in a case for it, it seems a bit more streamlined to just have steps 1 and 3?

rcaril: Upon further thought on this - the point I made is irrelevant as Terraform will still perform a Read and provide a diff prior to pushing changes. Resolving this as such.

philippschulte: I checked the current provider code again.

The current provider uses version `1` directly only during the initial Terraform create path, because `d.IsNewResource()` is true and Fastly has just created an editable version `1`.

For an existing or imported service, even if the service has no active version and only an unlocked version `1`, the provider sets `cloned_version` to the latest version during read/import, but then clones that version on the next versioned update when `stage = false`, which is the default.

So I think we should adjust the design doc to separate bootstrap from update behavior:

- bootstrap: new `fastly_service_vcl` uses Fastly-created version `1` directly
- update: active version exists → clone active
- update: no active version exists → clone latest existing version

That keeps the efficient initial create behavior while avoiding mutation of an existing/imported draft directly.

philippschulte: Updated section in c6c07d58ab5fe864e26a16efe9d24cb89c529453.

docs/dual-model-provider-design.md
philippschulte: Updated section in c8334301baab24e72ae65b699a1f6d843db6067b.

docs/dual-model-provider-design.md
philippschulte: Updated in 9578422a169baaacb03222a97917f6957147cfc3.

docs/dual-model-provider-design.md
philippschulte: Added in f9559f7d9feaa13294c5d89101c2dad85591f7a4.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,524 @@
...
+during normal CRUD. Users who do not want that behavior should use the explicit
+resource family instead.
+
+---
+
+## Compatibility resource behavior: `fastly_service_vcl`
+
+### Core rule
+
+`fastly_service_vcl` is the compatibility surface.
```
anthony-gomez-fastly: I think we decided on making this `_cdn`?

philippschulte:
> `fastly_service_cdn` / `fastly_service_compute`. Fastly docs use CDN and Compute as the correct terms. Let's align with that.

> **Naming feedback on the `_explicit` suffix:**
> I don't think `_explicit` communicates anything meaningful to developers. They can't tell what "explicit" refers to without reading docs.

> Can we use `_versioned` instead? `fastly_service_domain_versioned` signals that the resource requires a version pin.

> The compatibility family names (`fastly_service_cdn`, `fastly_service_compute`) stay as-is since they don't need a suffix.

> Separately, when we deprecate `fastly_service_vcl` in favor of `fastly_service_cdn`, `_vcl` should continue to work until the next major version release (or something like 6 months). The provider will surface deprecation warnings to give users time to migrate.

I agree with using `fastly_service_cdn` / `fastly_service_compute` for the compatibility family. That aligns better with current Fastly product terminology than `vcl` / `wasm`.

I’m less sure about `_versioned` for the explicit family. My concern is that both resource families manage versioned Fastly configuration; the difference is not whether the resources are versioned, but whether the user explicitly pins and controls the service version.

Because of that, I think `_pinned` may communicate the workflow more clearly than `_versioned`:

```text
fastly_service_domain_pinned
fastly_service_backend_pinned
```

That said, I agree that `_explicit` may not be obvious enough to users without reading the docs. I’d suggest we decide between `_pinned` and `_versioned`, with my preference being `_pinned` because it better describes the required version-pin behavior.

I also agree that `fastly_service_vcl` should continue to work as a deprecated alias for a transition period, with warnings pointing users to `fastly_service_cdn`.

docs/dual-model-provider-design.md
philippschulte: I haven't updated the docs and examples yet because I am waiting for Kevin to weigh in on https://github.com/fastly/terraform-provider-fastly/pull/1265#issuecomment-4488654211 and to make a final call on naming.

jaskiratr:
I prefer `_pinned` term.

`_versioned` could confuse people into thinking the auto-versioned resources are somehow _unversioned_

With `_pinned` one might ask "pinned to what?" But that's a one-time learning curve, and the required version attribute in the schema will answer that immediately.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,692 @@
...
+
+### `fastly_service_vcl` compatibility alias
+
+`fastly_service_cdn` is the canonical compatibility resource name for CDN
+services.
+
+`fastly_service_vcl` should continue to work as a deprecated alias for a
+transition period. The provider should surface deprecation warnings that point
+users from `fastly_service_vcl` to `fastly_service_cdn` so existing users have
+time to migrate without an immediate configuration break.
```
kpfleming: Unless I'm mistaken, there are going to be immediate configuration breaks (non-backwards-compatible HCL syntax). Off the top of my head, they include:

1. Removing the nested `product_enablement` block and using first-class unversioned resources for each product (which actually aligns with the product API).
2. Using 'nested attributes' for credentials inside `backend` and `logging` blocks so that plan diffs won't be redacted due to sensitive attributes.
3. Possibly moving away from `TypeSet` for `backend`, `domain`, `logging`, etc. and using a better schema entity type if one is available.

In my mind these improvements will happen in spite of the decision to retain the 'automatic activation' workflow, because they are independent of that workflow and will improve the provider in ways that have been long-standing issues for customers.

If that's true, then I don't see any need to retain the `fastly_service_vcl` resource name in the new provider; that simple name change is very easy for customers to adopt.

docs/dual-model-provider-design.md
```diff
@@ -0,0 +1,692 @@
...
+
+The provider should not support intermediate variants:
+
+- compatibility + manual activation
+- explicit resources + automatic activation
+
+There is intentionally no `activate` field on compatibility service resources.
+The compatibility resource family always performs automatic clone and activation
+during normal CRUD. Users who do not want that behavior should use the explicit
+resource family instead.
```
kpfleming: The same is true for 'stage', and should probably be called out here.

kpfleming (changes requested):

kpfleming:
Apologies for taking so long to review this. First, I'll admit that I only reviewed the Markdown files, I did not spend any time reviewing the HCL or shell-script files.

Overall I think this is in excellent shape, I've suggested a couple of small changes.

On the resource-naming front:

After reading the comments above, I'm not sold on the idea of using an `_explicit` suffix for the explicit-workflow resources. There are a few reasons:

1. This requires adding a suffix to a couple of dozen resources, instead of only adding a suffix to *two* compatibility-mode resources.
2. We want to encourage users to use the explicit-workflow resources, but adding a suffix makes them seem 'less desirable'.
3. Resources for versionless items won't have the suffix because it would be illogical, but this could lead users to believing that those resources are 'compatibility mode' resources because they don't have the suffix.

My personal preference is:

* Use `fastly_service_cdn` and `fastly_service_compute` for the default, explicit-workflow, service resources. Do not add suffixes to the first-class versioned resources which are linked to these service resources.
* Use `fastly_service_cdn_auto` and `fastly_service_compute_auto` (or some other ordering of those words) for the compatibility mode, automatic-worfklow, service resources.

jaskiratr:
Those are good points.

A couple of things that come to my mind:
1. Today `fastly_service_compute` is the automatic lifecycle resource. Under @kpfleming's  proposal, `fastly_service_compute` will become the explicit resource. We can provide thorough migration guides to avoid any sort of confusion.
2. `fastly_service_vcl` alias will still be deprecated during the transition. If you want to retain automatic workflow you'll need to switch to `fastly_service_cdn_auto`, otherwise if you're pinning it to a version you'll need to provide the service version. Either way, you'll need to face the schema changes anyway.

Overall, I like the idea of using `_auto` for automatically versioned resources. 👍🏽

jaskiratr (approved):

philippschulte:
I agree with Kevin’s naming recommendation. The design and examples have been updated so the explicit/default workflow gets the clean resource names, and the automatic compatibility workflow uses the `_auto` suffix.

| Workflow | Resource naming |
| --- | --- |
| Explicit/default CDN service | `fastly_service_cdn` |
| Explicit/default Compute service | `fastly_service_compute` |
| Explicit/default versioned resources | `fastly_service_domain`, `fastly_service_backend`, etc. |
| Automatic compatibility CDN service | `fastly_service_cdn_auto` |
| Automatic compatibility Compute service | `fastly_service_compute_auto` |

This removes the `_explicit` suffix from the preferred/default workflow and makes the automatic lifecycle behavior explicit in the resource name instead.

docs/dual-model-provider-design.md
philippschulte: This has been address in ad2a006095bc93d54f3e7e51bc5529beaaa33f7f.

docs/dual-model-provider-design.md
philippschulte: Address in ad2a006095bc93d54f3e7e51bc5529beaaa33f7f.

kpfleming (approved):

anthony-gomez-fastly (approved):

PR #1275, opened 2026-05-26
Add dual-model provider foundation

philippschulte:
### Change summary

This PR adds the initial dual-model provider implementation foundation.

It includes:

- provider entrypoint and registration
- initial Terraform Plugin Framework provider implementation
- `go.mod` and `go.sum`
- `.gitignore`
- `Makefile`
- local development override script support
- versioned domain and backend resources
- CDN and Compute service resources for the explicit/default workflow
- CDN and Compute `_auto` service resources for the automatic compatibility workflow

### Reviewer notes

This PR intentionally does not include:

- detailed provider/resource documentation for the versioned domain and backend resources
- acceptance tests or unit tests for the provider foundation

Both will be addressed in follow-up PRs after this foundation PR has been reviewed, approved, and merged.


internal/provider/action_service_compute_package_upload.go
```diff
@@ -0,0 +1,109 @@
...
+	Filename  types.String `tfsdk:"filename"`
+}
+
+func (a *serviceComputePackageUploadAction) Metadata(_ context.Context, req action.MetadataRequest, resp *action.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_service_compute_package_upload"
+}
+
+func (a *serviceComputePackageUploadAction) Schema(_ context.Context, _ action.SchemaRequest, resp *action.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Uploads or replaces a Compute package on a specific Fastly service version. Intended for explicit lifecycle workflows.",
```
anthony-gomez-fastly: I would say `Not for use in workflows with auto versioned resources` instead, but not that's a personal preference

internal/provider/action_service_compute_package_upload.go
```diff
@@ -0,0 +1,109 @@
...
+			fmt.Sprintf("Expected *providerData, got: %T", req.ProviderData),
+		)
+		return
+	}
+
+	a.providerData = providerData
+}
+
+func (a *serviceComputePackageUploadAction) Invoke(ctx context.Context, req action.InvokeRequest, resp *action.InvokeResponse) {
+	var cfg serviceComputePackageUploadModel
```
anthony-gomez-fastly: should we run `validateComputePackageInput` here so that we don't make extra API calls if the request is not configured correctly?

internal/datasources/serviceversion/data_source.go
```diff
@@ -0,0 +1,325 @@
...
+		return
+	}
+
+	serviceID := state.ServiceID.ValueString()
+
+	tflog.Debug(ctx, "Reading Fastly service versions", map[string]any{
+		"service_id": serviceID,
+	})
+
+	versions, err := d.client.ListVersions(ctx, &fastly.ListVersionsInput{
```
anthony-gomez-fastly: i think we can use [get service details](https://www.fastly.com/documentation/reference/api/services/service/#get-service-detail) and get rid of the active version setting logic

internal/provider/resource_service_backend.go
```diff
@@ -0,0 +1,301 @@
...
+	"github.com/fastly/go-fastly/v15/fastly"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/identityschema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+var _ resource.Resource = &serviceBackendResource{}
```
anthony-gomez-fastly: what do these do?

anthony-gomez-fastly: oh interesting, it's a compile check

.gitignore
```diff
@@ -0,0 +1,14 @@
+# Provider binary
+bin/
+terraform-provider-fastly
+
+# Terraform
```
anthony-gomez-fastly, line 5: can we add a `main.tf` or `example.tf` file for local debugging?

internal/provider/resource_service_backend.go
philippschulte: These are compile-time interface assertions.

They tell the Go compiler to verify that `*serviceBackendResource` implements the expected Terraform Plugin Framework interfaces:

```go
var _ resource.Resource = &serviceBackendResource{}
var _ resource.ResourceWithImportState = &serviceBackendResource{}
var _ resource.ResourceWithIdentity = &serviceBackendResource{}
```

So if someone later removes or incorrectly changes a required method, the provider fails to compile immediately.

In this case they verify that the backend resource:

- is a valid Terraform resource
- supports Terraform import
- supports identity-based import/query behavior

They do not create runtime behavior; they are just compile-time safety checks. This is a standard Go idiom used in Terraform provider development to compile-time check interface implementations.

.gitignore
philippschulte: 9dce8f29d0f0d69531a217bfa3fe9938cb275db6 adds `example.tf`.

internal/provider/action_service_compute_package_upload.go
philippschulte: Good suggestion.  da94e392cee444146fa896a1e30789c2643ee98f updates the wording to make the boundary clearer.

internal/provider/action_service_compute_package_upload.go
philippschulte: Yes, good call. Updated in df8ffe335766558a23f19d6e1f019673f678fbf3.

internal/datasources/serviceversion/data_source.go
philippschulte: I considered that, but I think `ListVersions` is the better fit for this data source.

`data.fastly_service_version` is intentionally scoped to version metadata: latest version, active version, staging version, locked versions, and the per-version metadata list. `ListVersions` gives us that directly without fetching the full service details payload.

`GetServiceDetails` is a much broader endpoint and can include a large service configuration payload for a selected version. For complex services that can be significantly heavier than needed here. Since this data source does not need domains, backends, conditions, dictionaries, snippets, etc., I would rather keep it on the narrower versions endpoint.

e1aaac2591c97fb03349d397131f690f03bb0b87 adds a short comment explaining why this uses `ListVersions` instead of `GetServiceDetails`.

anthony-gomez-fastly (approved):
LGTM! thank you!

internal/actions/computepackageupload/action.go
```diff
@@ -0,0 +1,114 @@
...
+				Required:    true,
+				Description: "Fastly Compute service ID.",
+			},
+			"version": schema.Int64Attribute{
+				Required:    true,
+				Description: "Writable Fastly service version to modify.",
+			},
+			"content": schema.StringAttribute{
+				Optional:    true,
+				Description: "The contents of the Compute deployment package as a base64-encoded string. Conflicts with `filename`.",
```
kpfleming, line 55: This can be expressed in the schema using https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.19.0/resourcevalidator#Conflicting, so that Terraform will enforce it.

internal/provider/action_service_version_activate.go
```diff
@@ -0,0 +1,114 @@
...
+		Attributes: map[string]schema.Attribute{
+			"service_id": schema.StringAttribute{
+				Required:    true,
+				Description: "Fastly service ID.",
+			},
+			"version": schema.Int64Attribute{
+				Required:    true,
+				Description: "Service version number to activate.",
+			},
+			"staging": schema.BoolAttribute{
```
kpfleming: Since Fastly uses the term 'activation' to mean *production* in most cases, I suspect that users will find it confusing to have an action named 'activate' that is also used for staging. I would prefer (in a future PR) to have this functionality split into separate actions named 'activate' and 'stage', even if they end up sharing most of their code.

internal/provider/action_service_version_clone.go
```diff
@@ -0,0 +1,103 @@
...
+	providerData, ok := req.ProviderData.(*providerData)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected ProviderData type",
+			fmt.Sprintf("Expected *providerData, got: %T", req.ProviderData),
+		)
+		return
+	}
+
+	a.client = providerData.client
```
kpfleming: This sequence is identical in most source files in this PR, so could be refactored into a helper function.

internal/provider/shared_service.go
```diff
@@ -0,0 +1,221 @@
...
+
+// ensureServiceTypeSupported checks that a service ID belongs to one of the
+// supported Fastly service types for a resource. This is primarily used by
+// first-class resources. Some resources are valid for both CDN and
+// Compute services, while future resources may only support one service type.
+//
+// This check runs during CRUD when service_id is known. It cannot reliably catch
+// all invalid combinations during `terraform validate`, because service_id may
+// be computed or come from a different state/workspace.
+func ensureServiceTypeSupported(ctx context.Context, client *fastly.Client, serviceID string, resourceName string, supportedTypes ...string) error {
```
kpfleming: This is going to be very expensive/slow to run, as it issues an API call for every resource attached to every service. I have an idea in mind for a way that this could be done more efficiently, which we can discuss during a regular meeting.

internal/provider/version_mutability.go
```diff
@@ -0,0 +1,193 @@
...
+	if mutability.Active {
+		diags.AddError(
+			"Fastly service version is not mutable",
+			fmt.Sprintf(
+				"Service %q version %d is currently active and cannot be modified. Clone the active version first and update your Terraform input to pin the new draft version before applying changes.",
+				serviceID,
+				version,
+			),
+		)
+	}
```
kpfleming: It is not possible for a service-version to be active and not locked - there is no need to check for 'active', checking for 'locked' is sufficient.

kpfleming (commented):
I've left a few comments about changes which could be made, but I don't think it is time to make those changes yet - I'd like to have a more extensive discussion in our planning meetings.

philippschulte:
I pushed an update in 46523444a48f1eac663518b650e915a26312351e:

This commit does two things:

1. Reorganizes the provider foundation into focused packages instead of keeping everything under `internal/provider`.

   The provider package is now mostly responsible for provider registration/configuration, while resources, actions, data sources, service/version helpers, validation, client helpers, and Compute package helpers live in separate packages. This should make the foundation easier to extend as we add more resources.

2. Addresses Kevin’s review feedback from the first pass:

   - adds schema-level conflict validation for Compute package `content` vs `filename`
   - splits staging out of the activate action into a dedicated stage action
   - refactors repeated provider data extraction into a shared helper
   - simplifies version mutability checks to rely on `locked`, since active versions are locked

This should give us a cleaner foundation before we add more resources, documentation, and tests.

internal/provider/version_mutability.go
philippschulte: Addressed in 46523444a48f1eac663518b650e915a26312351e.

internal/provider/shared_service.go
philippschulte: We have decided to improve this in a separate pull request after this one has been merged.

internal/provider/action_service_version_clone.go
philippschulte: Addressed in 46523444a48f1eac663518b650e915a26312351e.

internal/provider/action_service_version_activate.go
philippschulte: Addressed in 46523444a48f1eac663518b650e915a26312351e.

internal/actions/computepackageupload/action.go
philippschulte: Addressed in 46523444a48f1eac663518b650e915a26312351e.

go.mod
```diff
@@ -0,0 +1,43 @@
+module terraform-provider-fastly-dual-model-poc
```
kpfleming:
```suggestion
module github.com/fastly/terraform-provider-fastly
```

Can we change this now, or will it cause problems since the name matches the existing provider's module declaration?

kpfleming (approved):
This looks great, we can get started with this baseline in place.

go.mod
philippschulte: Good catch! Since this code now lives in `fastly/terraform-provider-fastly`, the module path should use the canonical repository path. 83a2cab8df0887b8b443d1c3b6da5dd719fa7de1 updates `go.mod` and the internal imports from the PoC module name to `github.com/fastly/terraform-provider-fastly`.

internal/client/not_found.go
```diff
@@ -0,0 +1,8 @@
+package client
+
+import fastly "github.com/fastly/go-fastly/v15/fastly"
+
+func IsNotFound(err error) bool {
```
anthony-gomez-fastly: Not sure this needs a whole new file?

internal/computepackage/package.go
```diff
@@ -0,0 +1,174 @@
...
+func NestedBlockSchema() schema.ListNestedBlock {
+	return schema.ListNestedBlock{
+		Description: "Compute package attached to this service version. At most one package block is supported.",
+		NestedObject: schema.NestedBlockObject{
+			Attributes: CommonAttributes(),
+		},
+	}
+}
+
+func Equal(a, b []Model) bool {
```
anthony-gomez-fastly, line 63: Maybe name this something like `ModelSliceEqual`? from the signature `computepackage.Equal` i'd expect it to compare Compute Packages

anthony-gomez-fastly: or that is what it's doing? so instead of naming it Model, name it ComputePackage maybe?

philippschulte: I checked this against the existing first-class resource helper packages, and `Equal` is the current convention for comparing slices of package-local models, for example `domain.Equal` and `backend.Equal`.

I agree `computepackage.Equal` is a little easier to misread because "package" can mean the Compute package artifact as well as the model. To keep naming consistent with the rest of the provider, 8c8521121d82d0e4c74173cc310e175b5eca66cb keeps `Equal` but adds a doc comment clarifying that it compares compute package model slices and that only one package block is supported.

internal/provider/shared_service.go
philippschulte: This was a big design flow and therefore I addressed it here with 5e4f26d0cb966a501e069277261efe077ee8d6f7.

internal/client/not_found.go
philippschulte: Maybe there will be more helpers like this. Therefore, I kept the function but renamed the file in c7d52c3f8288cb6ea66051eafbb8f1e2cb8b03fe.

philippschulte:
Updated this PR with a few follow-up changes:

- Added a provider-scoped cached service type checker so explicit first-class resources no longer call `GetServiceDetails` once per resource operation. This now performs one lookup per unique service ID per provider run.
- Renamed `internal/client/not_found.go` to `internal/client/errors.go` and updated `IsNotFound` to use `errors.As` so wrapped Fastly HTTP errors are handled correctly.
- Kept `computepackage.Equal` naming consistent with the existing `domain.Equal` / `backend.Equal` helper pattern, but added a doc comment clarifying that it compares compute package model slices.
- Refactored the new service version data source to avoid the expensive unpaginated `GET /versions` call. It now only exposes `latest_version` and `active_version`, using targeted lookups instead of returning all versions.

This should address the main performance and naming feedback from the latest review. A final quick review of these changes would be appreciated. Thanks!

anthony-gomez-fastly (approved):
TY!

internal/service/service_type.go
```diff
@@ -0,0 +1,104 @@
...
+	key := ServiceTypeCheckKey{
+		ServiceID: serviceID,
+	}
+
+	c.mu.Lock()
+	if cached, ok := c.cache[key]; ok {
+		c.mu.Unlock()
+		return cached, nil
+	}
+	c.mu.Unlock()
```
kpfleming: Unless I'm misunderstanding something, it seems like it would be safer to just leave the Mutex locked, and use `defer` to unlock it when the function returns. That might avoid the need to use `singleflight` at all, since the Mutex would be locked while the service type information is obtained. We use the simple Mutex-based method in go-fastly for resource locks and it works (and was reviewed by at least one Go expert-type person).

kpfleming: That might also eliminate the need to wrap the result in a `struct`, because without `singleflight` the map of entries could just be `map[string]string`.

philippschulte: Agreed, a simple mutex-protected cache is good enough here and keeps the implementation easier to maintain.

I removed `singleflight` from both cache implementations:

- `internal/service/service_type.go` updated in 99b9d7bb2f738b12f8603a0262629841e2483c51
- `internal/service/mutability.go` updated in 615241c94ad5e2c26427f0ea6f822cfa61138d43

kpfleming (approved):
Ship it!

PR #1279, opened 2026-06-02
Add initial coverage for dual-model provider foundation

anthony-gomez-fastly:
Claude was used for this PR

### Change summary

* unit tests for backends
* acceptance tests for cdn and compute services
* Makefile commands for unit and acceptance tests
* Testing.md file for testing best practices and FAQs

 All Submissions:

### New Feature Submissions:

Screenshot 2026-06-02 at 6 06 46 PM

### Changes to Core Features:


philippschulte (commented):
Thanks for putting this first round together — I think the overall direction looks good.

The split between unit tests for expand/flatten/helper behavior and acceptance tests for real Terraform resource flows feels right, and the acceptance tests are using the expected `resource.Test` / `PreCheck` / `ProtoV6ProviderFactories` / `CheckDestroy` pattern.

A few early comments before this grows further:

- The shared `BuildConfig` helper is useful, but trimming the final `}` and injecting blocks feels fragile. I’d prefer an explicit placeholder in the fixture templates, then replace that placeholder with the requested blocks:

  ```hcl
  resource "fastly_service_cdn_auto" "test" {
    name          = "SERVICE_NAME"
    force_destroy = true

    # BLOCKS
  }
  ```

- The acceptance test helper code is duplicated between CDN and Compute. Since this suite will grow, I’d centralize the common provider factory, precheck, client creation, and destroy/existence helpers. For example:

  ```go
  func CheckServiceDestroy(resourceType string) func(*terraform.State) error {
  	return func(s *terraform.State) error {
  		// shared destroy check for CDN/Compute service resources
  	}
  }

  func CheckServiceExists(resourceName string) resource.TestCheckFunc {
  	return func(s *terraform.State) error {
  		// shared service existence check
  	}
  }
  ```

   Then each resource test can stay focused on resource-specific behavior:

   ```go
   CheckDestroy: testhelpers.CheckServiceDestroy("fastly_service_cdn_auto")
   ```

   I would add a small shared acceptance-test helper package, for example:

   ```
   internal/testhelpers/acceptance.go
   ```

- The `CheckDestroy` helpers call `GetServiceDetails` without a version filter. Since `/details` can be expensive, we should either use a cheaper endpoint or pass a version filter there too:

  ```go
  version := 1
  service, err := client.GetServiceDetails(ctx, &fastly.GetServiceDetailsInput{
  	ServiceID: rs.Primary.ID,
  	Version:   &version,
  })
  ```

Overall I think this is a solid start and the direction is good. I’d centralize the shared acceptance-test plumbing before adding the rest of the coverage.

internal/resources/backend/backend_test.go
```diff
@@ -0,0 +1,813 @@
...
+			forceAll: true,
+			validate: func(t *testing.T, input *fastly.UpdateBackendInput) {
+				assert.Equal(t, "service-123", input.ServiceID)
+				assert.Equal(t, 6, input.ServiceVersion)
+				assert.Equal(t, "test-backend", input.Name)
+				assert.NotNil(t, input.Address)
+				assert.Equal(t, "api.updated.com", *input.Address)
+				assert.NotNil(t, input.Port)
+				assert.Equal(t, 443, *input.Port)
+				assert.NotNil(t, input.UseSSL)
```
rcaril, line 481: Should we consider making some of these `assert.nNil` to also test that some fields are omitted as expected?

anthony-gomez-fastly: i don't think so, i already cover most of the cases. are there any specific fields you were thinking of?

rcaril: I was thinking that since fields like `input.UseSSL`, `input.Weight` and `input.ConnectTimeout` didn't change, making some of them `assert.ninl` would demonstrate that `BuildUpdateInput` with `forceAll=false` is correctly omitting unchanged  fields from the payload rather than sending everything.

anthony-gomez-fastly: makes sense, adding

rcaril (approved):
LGTM!

philippschulte (changes requested):
Thanks for updating this and moving the PR out of draft. The structure is heading in the right direction: the shared acceptance helpers, centralized `internal/acceptance_tests` package, config builder, CI setup, and initial `_auto` tests are all good foundations.

I also think the PR title should be updated. The current title, `add unit tests for backends and acceptance tests for cdn and compute services`, does not fully describe the scope anymore since this PR also adds test infrastructure, shared helpers, fixtures, CI, testing docs, and initial foundation coverage. Something like `Add initial coverage for dual-model provider foundation` might be more accurate.

That said, I do not think this PR yet provides complete coverage for everything introduced by the foundation PR or for the lifecycle behavior described in the dual-model provider design. The current tests cover basic create/update/import-style flows, but they do not yet verify the core dual-model semantics.

The biggest missing area is the automatic compatibility behavior. The design says `_auto` resources should automatically create/select a writable service version, reconcile nested versioned resources into that version, validate the version, and activate it. On update, the provider should select exactly one working version for the service, write all nested changes into that same version, and activate it. It should clone the active version when one exists, or clone the latest version when no active version exists.

I think we need explicit acceptance coverage for those behaviors, not just assertions that `active_version` and `managed_version` are set.

Missing automatic compatibility tests:

- `fastly_service_cdn_auto` create should prove bootstrap uses version `1`, writes nested domain/backend to that version, validates it, and activates it.
- `fastly_service_compute_auto` create should prove bootstrap uses version `1`, writes nested domain/backend/package to that version, validates it, and activates it.
- `_auto` update with nested domain changes should prove a new managed version is selected and activated.
- `_auto` update with nested backend changes should prove a new managed version is selected and activated.
- `_auto` update with multiple nested changes in one apply should prove all changes land in the same managed version.
- `_auto` update should prove the provider does not clone once per changed nested block.
- `_auto` update should prove `active_version == managed_version` after apply.
- `_auto` update should prove clone-from-active behavior when an active version exists.
- `_auto` update should prove clone-from-latest behavior when no active version exists.
- `_auto` no-op apply should prove no new version is cloned when there are no relevant changes.
- `_auto` service-only updates such as name/comment should prove whether they should or should not clone according to the design/implementation.
- `_auto` import/read should prove the read version selection order: active version first, latest version if no active version exists.
- `_auto` import should verify generated/imported nested domain/backend/package state, not only that import succeeds.

Missing explicit/default family tests:

- `fastly_service_cdn` basic create/read/update/delete/import.
- `fastly_service_compute` basic create/read/update/delete/import.
- `fastly_service_domain` create/read/update/delete/import against a pinned CDN service version.
- `fastly_service_domain` create/read/update/delete/import against a pinned Compute service version.
- `fastly_service_backend` create/read/update/delete/import against a pinned CDN service version.
- `fastly_service_backend` create/read/update/delete/import against a pinned Compute service version.
- Explicit resources should prove they write directly to the configured `version`.
- Explicit resources should prove they do not clone versions.
- Explicit resources should prove they do not activate versions.
- Explicit resources should reject writes to locked/active versions with the expected diagnostic.
- Explicit resources should allow writes to a writable draft version.
- Explicit resources should continue to work when the version is manually cloned through the explicit action workflow.
- Explicit resources should cover service-kind validation, including accepted CDN/Compute combinations and rejected unsupported combinations as future service-kind-specific resources are added.

Missing action tests:

- `fastly_service_version_clone` action basic success.
- Clone action output/version behavior.
- Clone action error path for missing service or invalid version.
- `fastly_service_version_activate` action basic success.
- Activate action should update the service active version.
- `fastly_service_version_stage` action basic success where staging is valid.
- Stage action error/diagnostic behavior where staging is invalid.
- `fastly_service_compute_package_upload` action should upload a package to a pinned Compute service version.
- Package upload should be triggered when the source/hash changes.
- Package upload should reject CDN services.
- Package upload should reject locked/active versions if that is expected by the implementation.

Missing data source tests:

- Service version data source should return `latest_version`.
- Service version data source should return `active_version`.
- Service version data source should return useful diagnostics for missing/invalid service IDs.

Missing provider/framework tests:

- Provider config with token set in provider configuration.
- Provider config with token from `FASTLY_API_TOKEN`.
- Missing token diagnostic.
- Invalid token/client error behavior if applicable.

Missing query/list/import coverage:

- List/query support for explicit/default `fastly_service_domain`.
- List/query support for explicit/default `fastly_service_backend`.
- List/query support for explicit/default CDN and Compute services.
- Generated explicit/default resources should include the version that was read.
- Query/read version selection should use active version first and latest version if no active version exists.
- Automatic compatibility should use import for generated nested configuration rather than query.

Missing old-provider parity coverage:

The old provider had substantial service/domain/backend acceptance and helper coverage. I think the new provider should carry forward the same intent, adjusted for the new dual-model design:

- Domain basic create without service association, if still applicable.
- Domain association/update with a service, if still applicable.
- Domain import.
- Domain destroy verification.
- Backend flatten/expand parity for CDN and Compute.
- Backend defaults and optional fields.
- Backend SSL/TLS fields.
- Backend timeout fields.
- Backend load balancing fields.
- Backend update behavior.
- Backend import behavior.
- Service lifecycle behavior around cloning, activation, staging, force_destroy/reuse-style behavior where applicable to the new model.

I would also add a coverage matrix to the PR description so reviewers can see what is currently covered and what is still missing.

Suggested PR description matrix:

| Area | Unit | Acceptance | Status |
|---|---:|---:|---|
| `fastly_service_cdn_auto` basic CRUD/import | Partial | Partial | Needs lifecycle assertions |
| `fastly_service_compute_auto` basic CRUD/import | Partial | Partial | Needs lifecycle/package assertions |
| `_auto` clone/no-clone/version activation behavior | No | No | Missing |
| `fastly_service_cdn` explicit/default | No | No | Missing |
| `fastly_service_compute` explicit/default | No | No | Missing |
| `fastly_service_domain` explicit/default | Partial | No | Missing acceptance |
| `fastly_service_backend` explicit/default | Partial | No | Missing acceptance |
| Version actions | No | No | Missing |
| Compute package upload action | No | No | Missing |
| Service version data source | No | No | Missing |
| List/query resources | No | No | Missing |
| Provider configuration | No | No | Missing |

I’m happy with the direction, but since the goal is to cover everything introduced by the foundation PR, I think we should expand this PR to include the missing explicit/default tests, action tests, data source tests, provider/framework tests, and especially the `_auto` lifecycle behavior tests from the design document.

anthony-gomez-fastly:
Ty for the feedback @philippschulte the only thing i'd push back on is query and import functionality, which we can handle separately. I'll work on covering the areas you pointed out

philippschulte:
> Ty for the feedback @philippschulte the only thing i'd push back on is query and import functionality, which we can handle separately. I'll work on covering the areas you pointed out

Thanks, that makes sense on query coverage. I’m fine with handling Terraform query/list resource tests separately since query support is only intended for the explicit/default first-class resource family, which is part of the second release.

I do want to push back on import coverage, though. Import tests are commonly part of Terraform provider acceptance tests for managed resources, and I think this PR should include them for the foundation resources in both families.

HashiCorp documents this as acceptance-test import mode, and the usual pattern is to add an import step after create/update:

```go
{
    ResourceName:      "fastly_service_cdn_auto.test",
    ImportState:       true,
    ImportStateVerify: true,
}
```

For the dual-model foundation, I think we should cover import for at least:

- `fastly_service_cdn_auto`
- `fastly_service_compute_auto`
- `fastly_service_cdn`
- `fastly_service_compute`
- `fastly_service_domain`
- `fastly_service_backend`

This is especially important for `_auto` resources because import/read needs to reconstruct state from the remote service, including nested configuration and version-related state.

So I’m good with splitting query/list tests out, but I think import tests should remain in scope for this foundation coverage PR.

anthony-gomez-fastly:
all changes have been merged, the leftover files are from being renamed 🥳

PR #1287, opened 2026-06-08
test: add CI workflow and unit tests for Backend and domain flatten and expand functions

anthony-gomez-fastly:
This PR is a subset of the changes in [1279](https://github.com/fastly/terraform-provider-fastly/pull/1279/changes), broken up for easier digestion

### Change summary

Set up GitHub Actions workflow for automated testing and add unit tests for backend and domain resources.

  - Add .github/CODEOWNERS for developer-tools team ownership
  - Add .github/workflows/pr.yml targeting dual-model-framework-rewrite branch
  - Add backend_test.go with tests for flatten, expand, and schema functions
  - Add domain_test.go with tests for flatten, expand, and schema functions
  - Add test-unit and test-acc Makefile targets

 All Submissions:

### New Feature Submissions:


internal/resources/backend/resource.go
```diff
@@ -117,6 +117,13 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp
 
 	flatten(ctx, b, &plan)
 	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if resp.Identity != nil {
+		resp.Diagnostics.Append(resp.Identity.SetAttribute(ctx, path.Root("backend_id"), plan.ID)...)
```
anthony-gomez-fastly: wasn't sure if this was the best way to represent it, but it seemed correct?

internal/resources/backend/flatten.go
```diff
@@ -27,7 +27,7 @@ func FlattenToNestedModel(b *fastly.Backend) NestedModel {
 	m.ErrorThreshold = service.Int64PointerOrDefault(b.ErrorThreshold, 0)
 	m.FirstByteTimeout = service.Int64PointerOrDefault(b.FirstByteTimeout, 15000)
 	m.HealthCheck = service.StringPointerOrDefault(b.HealthCheck, "")
-	m.KeepaliveTime = service.Int64PointerOrNull(b.KeepAliveTime)
+	m.KeepaliveTime = service.Int64PointerOrDefault(b.KeepAliveTime, 0)
```
anthony-gomez-fastly: the api kept showing inconsistencies here, happy to change it back if incorrect

rcaril: What inconsistencies were you seeing here?  From my testing it seems like this remains null until set?

anthony-gomez-fastly: Screenshot 2026-06-08 at 11 54 45 AM

anthony-gomez-fastly: for the acceptance tests in a future PR, this is the error i get

anthony-gomez-fastly: from what i understand, we were not correctly representing in state what the default should be if null is returned

rcaril: Perhaps this is a situation where we can us the new nullable helper in go-fastly? 
https://github.com/fastly/go-fastly/blob/608c8ad39e614fc0bd626091134842d0ce58bc0c/fastly/helpers.go#L34

anthony-gomez-fastly: that's for strings and not ints

internal/resources/backend/resource.go
rcaril: When doing some research on this, I noticed a few things. 

I'm not sure that we need to have a single anchor point for the backend identification. 

Looking at Hashicorp's examples on this in the framework 6 provider exemplifies sending multiple identifiers back via struct. 
https://github.com/hashicorp/terraform-provider-corner/blob/58f1ec766bd67e45e09dd45fb4b2e6cd424c82c4/internal/framework6provider/identity_resource.go#L124

We can use this suggested pattern of sending multiple identities instead, which would look something like:
```
if resp.Identity != nil {
      resp.Diagnostics.Append(resp.Identity.SetAttribute(ctx, path.Root("service_id"), plan.Service)...)
       resp.Diagnostics.Append(resp.Identity.SetAttribute(ctx, path.Root("version"), plan.Version)...)
        resp.Diagnostics.Append(resp.Identity.SetAttribute(ctx, path.Root("name"), plan.Name)...)
   }
```

The alternative here is using this new pattern of defining a dedicated `BackendIdentityModel` struct  (like the Corner provider does) - which would reduce the repeated `Diagnostics.Append` calls via the `resp.Identity.Set(ctx, &identityModel)` pattern. 

EG:
```
type BackendIdentityModel struct {
ServiceID types.String `tfsdk:"service_id"`
      Version   types.Int64  `tfsdk:"version"`
      Name      types.String `tfsdk:"name"`
  }

  resp.Diagnostics.Append(resp.Identity.Set(ctx, &BackendIdentityModel{
      ServiceID: plan.Service,
      Version:   plan.Version,
      Name:      plan.Name,
  })...)

```

anthony-gomez-fastly: sounds good, done in `d5f187a8`

anthony-gomez-fastly: also did the same for domains

internal/resources/domain/resource.go
rcaril: Should we combine these as well to a single `Append call`?

internal/resources/backend/resource.go
rcaril: Here as well - does it make sense to also merge these?

rcaril (commented):
Looks good so far! Will provide a further update after CDTOOL-1497 is completed

internal/resources/domain/resource.go
anthony-gomez-fastly: good catch

internal/resources/backend/resource.go
anthony-gomez-fastly: yup, done in `943e3d9c`

internal/resources/domain/resource.go
anthony-gomez-fastly: done in `943e3d9c`

philippschulte (approved):
Thanks for splitting the original larger test PR into smaller focused PRs. I’m good with this PR as the first CI/unit-test subset.

A few small requests before merge:

1. Please update the PR title so it is more clearly tied to the dual-model provider foundation work and the backend/domain scope.
2. Please add a link in the PR description to the original larger test PR so reviewers understand how these PRs relate to each other.
3. Please update the PR description wording around `domain_test.go`. It currently says “tests for CRUD operations”, but these are unit tests for helper/model behavior, not Terraform provider CRUD tests.

Thanks!

internal/resources/backend/expand.go
```diff
@@ -29,18 +29,10 @@ func BuildCreateInput(serviceID string, version int, m NestedModel) *fastly.Crea
...
-	if service.Int64Value(m.MaxLifetime) > 0 {
-		input.MaxLifetime = fastly.ToPointer(int(service.Int64Value(m.MaxLifetime)))
-	}
-	if service.Int64Value(m.MaxUse) > 0 {
-		input.MaxUse = fastly.ToPointer(int(service.Int64Value(m.MaxUse)))
-	}
-	if service.StringValue(m.Comment) != "" {
-		input.Comment = fastly.ToPointer(service.StringValue(m.Comment))
-	}
+	input.KeepAliveTime = fastly.NullInt(int(service.Int64Value(m.KeepaliveTime)))
```
anthony-gomez-fastly, line 32: @rcaril used new helpers here 🥳

internal/resources/backend/flatten.go
anthony-gomez-fastly: I think this change was needed to keep the plan clean, the other similarly handled values all have the default handling

anthony-gomez-fastly: oh wait, i think i need to revert the schema change as well to test

anthony-gomez-fastly: yup, that did it

internal/resources/backend/expand.go
rcaril: Glad it worked out! 🎆

rcaril (approved):
Thanks for the additional updates - looks great!

PR #1288, opened 2026-06-09
test(internal/acceptance_tests): add acceptance tests for cdn/compute services with auto mode

anthony-gomez-fastly:
This PR is a subset of the changes in [1279](https://github.com/fastly/terraform-provider-fastly/pull/1279/changes), broken up for easier digestion

### Change summary
- Add internal/acceptance_tests directory with test framework
  - Include 11 Terraform test block templates for resource scenarios:
    * Service lifecycle tests (cdn, cdn_auto, compute_auto)
    * Backend resource tests (single, multiple)
    * Domain resource tests
    * Version management actions (activate, clone)
    * Package deployment tests
  - Add config_builder.go for dynamic test configuration generation
  - Add test_helpers.go with common test utilities

 

 All Submissions:

### New Feature Submissions:

Screenshot 2026-06-09 at 1 31 22 PM


internal/resources/backend/schema.go
```diff
@@ -106,6 +106,7 @@ func CommonAttributes() map[string]schema.Attribute {
 		"keepalive_time": schema.Int64Attribute{
 			Optional:    true,
 			Computed:    true,
+			Default:     int64default.StaticInt64(0),
```
anthony-gomez-fastly: @rcaril since this value is computed, terraform requires the value to be known so we can't have it set up without a default. that was the thing that the acceptance tests were catching and the reason for these changes

rcaril: I believe we used `Computed` here previously as a workaround. This value usually is modified by the user, replacing the initial `nil` passed by the API.  With the new `null` updates, would we be able to remove the `Computed` field?

anthony-gomez-fastly: makes sense to me @philippschulte what are your thoughts

philippschulte: In the current SDK provider, `keepalive_time` is modeled as `Optional + Computed` with no default:

```go
"keepalive_time": {
    Type:     schema.TypeInt,
    Optional: true,
    Computed: true,
}
```

The SDK create path only sends it when the configured value is greater than zero:

```go
if resource["keepalive_time"].(int) > 0 {
    opts.KeepAliveTime = gofastly.ToPointer(resource["keepalive_time"].(int))
}
```

And the SDK flatten path preserves API nil/omitted values:

```go
if resource.KeepAliveTime != nil {
    data["keepalive_time"] = *resource.KeepAliveTime
}
```

By contrast, this PR changes the framework provider to default `keepalive_time` to `0` and flatten API nil to `0`:

```diff
- m.KeepaliveTime = service.Int64PointerOrNull(b.KeepAliveTime)
+ m.KeepaliveTime = service.Int64PointerOrDefault(b.KeepAliveTime, 0)
```

That is a behavior change from the SDK provider, not just a test update.

This is different from `max_lifetime` / `max_use`. For those fields, the SDK provider already models them as `Default: 0`, validates non-negative values, omits zero on create, and sends zero on update when explicitly modified. We also previously validated that unset and `0` are functionally equivalent for those fields.

I do not see the same proof for `keepalive_time`.

Could we either keep framework `keepalive_time` aligned with the SDK provider by preserving null/unknown behavior, or add explicit compatibility tests/proof that:

1. omitted `keepalive_time` and explicit `0` are API-equivalent;
2. existing SDK state with omitted/null `keepalive_time` upgrades to the framework provider with a clean normal plan;
3. refresh does not introduce unwanted state churn;
4. update from non-zero back to `0` works if `0` is intended to be supported?

philippschulte:
One non-blocking maintainability concern on the acceptance config builder:

The current `BuildConfig` approach loads fixture files, replaces `RESOURCES`, and then applies raw string replacements for placeholders like `SERVICE_NAME`, `DOMAIN_NAME`, and `BACKEND_NAME`.

That works as a starting point, but I think it is a little fragile as this framework grows. Since the placeholders are plain words, `strings.ReplaceAll` could accidentally replace matching text in unrelated fixture content, comments, heredocs, or test data. It also makes it possible to generate invalid Terraform while Go tests still pass if they only check for expected snippets.

I would prefer either using `text/template`, or at least switching the placeholder format to something less collision-prone, for example:

```hcl
{{SERVICE_NAME}}
{{DOMAIN_NAME}}
{{BACKEND_NAME}}
{{RESOURCES}}

I think it is worth tightening before many more acceptance tests are built on top of this framework.

**Review by philippschulte — CHANGES_REQUESTED** [review]
Thanks for splitting this out from #1279. I like the direction of this PR.

I’m going to request changes on this one because I think one lifecycle expectation needs to be corrected before this PR lands.

The Compute `_auto` update test appears to change only service-level metadata, but expects a new managed service version:

```go
resource.TestCheckResourceAttr("fastly_service_compute_auto.test", "active_version", "2"),
resource.TestCheckResourceAttr("fastly_service_compute_auto.test", "managed_version", "2"),
```

I double-checked the current SDKv2 provider behavior for `fastly_service_compute`. The current provider does not appear to have an acceptance test specifically for a Compute service-name-only update. `TestAccFastlyServiceCompute_basic` covers create/import, and `TestAccFastlyServiceCompute_stage` covers staged updates with nested backend changes.

However, the current provider implementation clearly treats service `name` / `comment` changes as service metadata updates that do not require a new service version:

```go
// Update Name and/or Comment. No new version is required for this.
if d.HasChanges("name", "comment") && shouldActivate {
    _, err := conn.UpdateService(...)
    ...
}
```

Version cloning is then driven separately by changes reported from the service attribute handlers.

Because of that, I do not think the Compute `_auto` test should expect a new version when only the service name changes. This should match the CDN `_auto` behavior: service metadata changes should update the service but should not clone/activate a new managed version.

Please update the Compute `_auto` update test in one of these ways:

- If the test is intended to prove no-clone behavior for service metadata updates, keep the service-name-only update and assert that `active_version` / `managed_version` remain unchanged.
- If the test is intended to prove clone-and-activate behavior, change a versioned nested field such as backend, domain, or package configuration so the expected move to version `2` is actually caused by versioned service configuration.

After the Compute `_auto` lifecycle expectation is corrected, I think this will be a solid first acceptance-test framework PR for the dual-model foundation work.

internal/resources/backend/schema.go
anthony-gomez-fastly: i figured it out, the issue was that the state wasn't being updated in the auto versions, added those in

anthony-gomez-fastly:
@philippschulte I think the reason the test is expecting a new version is due to the nature of the package hash computation. since the hash doesn't have a data source yet the comparison in the `Equals` function of computepackage/package.go is returning false. is there a plan to implement a data source like the current provider?

philippschulte:
> @philippschulte I think the reason the test is expecting a new version is due to the nature of the package hash computation. since the hash doesn't have a data source yet the comparison in the `Equals` function of computepackage/package.go is returning false. is there a plan to implement a data source like the current provider?

@anthony-gomez-fastly please have a look at the `compute-explicit-package` example:

```
# Terraform Actions are not stateful resources, so this terraform_data resource
# provides the stateful diff that triggers package uploads during normal apply.
resource "terraform_data" "compute_package" {
  input = {
    service_id       = fastly_service_compute.app.id
    version          = var.service_version
    filename         = local.package_path
    source_code_hash = filebase64sha256(local.package_path)
  }

  lifecycle {
    action_trigger {
      action = action.fastly_service_compute_package_upload.this
      events = ["create", "update"]
    }
  }
}
```

anthony-gomez-fastly:
That makes sense for the orchestrated examples, but is my change still valid for both cases? It's mostly the auto version that's giving issues at the moment

philippschulte (approved):
Thanks, I pulled the latest `base-acceptance-tests` branch and verified the lifecycle fix locally.

The package hash explanation makes sense as the root cause, but I agree with the latest implementation direction: service metadata-only updates should not clone a new service version, and `computepackage.Equal` should not treat an API-populated package hash as a user-requested package change.

Thank you!

rcaril (approved):
The rest of the updates look good to me!

PR #1294, opened 2026-06-11
Remove the version component of the backend and domain identity

anthony-gomez-fastly:
### Change summary
This was going to cause issues when attempting to use the explicit mode. Terraform would attempt to create a new resource since the version was changed and therefore the identity of the resource in state.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-06-11 at 3 46 42 PM

### Changes to Core Features:


philippschulte (changes requested):
I agree with the direction of removing `version` from the stable Framework identity for backend/domain.

However, I think this PR should also preserve legacy `terraform import` support for these resources, because it is changing the same identity/import path.

After this change, identity import only provides `service_id + name`, but backend/domain are service-versioned API objects. The provider still needs `service_id`, `version`, and `name` to perform the initial read during import.

For `_auto` resources, importing by service ID only is correct because the service resource reconstructs nested configuration. For explicit first-class backend/domain resources, import needs the version as lookup data.

I think the correct model is:

- Stable Framework identity: `service_id + name`
- Legacy CLI import ID: `service_id/version/name`

That keeps `version` out of the stable identity so changing versions does not make Terraform treat the resource as a different logical object, while still allowing `terraform import fastly_service_backend.origin service123/3/origin` and `terraform import fastly_service_domain.www service123/3/www.example.com` to hydrate full state.

Could we add legacy composite-ID parsing and import tests in this PR before merging?

docs/import-legacy-composite-ids.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+### Domain Import
+
+```bash
+# Legacy composite ID format (supported for backwards compatibility)
+terraform import fastly_service_domain.www service123/3/www.example.com
+
+# Identity-based import (new format)
+terraform import fastly_service_domain.www '{"service_id":"service123","name":"www.example.com"}'
```
philippschulte: Where did you get this from? To my knowledge you cannot use structured identity-based import via the traditional terraform import CLI command. Did you test this locally?

philippschulte (dismissed):
Thanks for adding the legacy composite import handling and acceptance coverage. The implementation direction looks right to me: backend/domain keep stable identity as `service_id + name`, while legacy import IDs use `service_id/version/name` so the provider has enough information to perform the initial API read and populate full state.

One docs request: can we remove `docs/import-legacy-composite-ids.md` from this PR?

Since I already work on the provider documentation, I think this standalone internal docs page should not be added under `docs/`. It is not part of the normal Terraform Registry resource documentation structure, and it may confuse users because import behavior should be documented on the relevant resource pages instead.

I would rather cover this in:

- `docs/resources/service_backend.md`
- `docs/resources/service_domain.md`

Also, I would avoid documenting the JSON identity-style CLI import unless we have a tested end-to-end workflow for it. The important compatibility path for explicit backend/domain imports is:

```shell
terraform import fastly_service_backend.origin SERVICE_ID/VERSION/BACKEND_NAME
terraform import fastly_service_domain.site SERVICE_ID/VERSION/DOMAIN_NAME
```

That keeps the PR focused on the import functionality while leaving user-facing docs to my docs PR.

docs/import-legacy-composite-ids.md
anthony-gomez-fastly: it worked when we included a version, but since we don't anymore, it doesn't work

internal/resources/backend/resource.go
```diff
@@ -270,10 +278,63 @@ func (r *Resource) Delete(ctx context.Context, req resource.DeleteRequest, resp
...
+			resp.Diagnostics.Append(resp.Identity.Set(ctx, &BackendIdentityModel{
+				ServiceID: types.StringValue(serviceID),
+				Name:      types.StringValue(name),
+			})...)
+		}
+		return
+	}
+
+	// Support new identity-based import
 	if req.ID != "" {
```
philippschulte, line 332: For explicit backend/domain, a non-composite CLI import ID is not enough. These resources need:

```
service_id + version + name
```

to read from Fastly.

So if someone runs:

```
terraform import fastly_service_backend.origin service123
```

this fallback can set only `id` and leave `service_id`, `version`, and `name` unset.

Can we please replace that fallback with a clear diagnostic error:

```go
if req.ID != "" {
	resp.Diagnostics.AddError(
		"Invalid Import ID",
		"Expected import ID in format: service_id/version/name.\n"+
			"For example: service123/3/origin",
	)
	return
}
```

Same for domain.

Identity-based import should go through `req.Identity`, not `req.ID`.

internal/importutil/composite_id.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// ParseCompositeID parses a legacy composite import ID in the format "service_id/version/name"
+// and returns the individual components. This supports backwards compatibility for resources
+// that were imported with version information before the identity schema was updated.
+func ParseCompositeID(id string) (serviceID string, version int, name string, err error) {
```
philippschulte, line 12: The parser should reject versions less than `1`:

```go
if version < 1 {
	return "", 0, "", fmt.Errorf("version must be greater than 0 in import ID %q", id)
}
```

internal/importutil/composite_id_test.go
```diff
@@ -0,0 +1,109 @@
...
+			wantErr:       false,
+		},
+		{
+			name:          "name with slashes",
+			id:            "service123/5/backend/with/slashes",
+			wantServiceID: "service123",
+			wantVersion:   5,
+			wantName:      "backend/with/slashes",
+			wantErr:       false,
+		},
```
philippschulte, line 44: Please add:

```go
{
	name:    "version 0",
	id:      "service123/0/backend1",
	wantErr: true,
},
{
	name:    "negative version",
	id:      "service123/-1/backend1",
	wantErr: true,
},
```

philippschulte (changes requested):
Thanks, this is much closer.

I have two small requested changes before approving:

1. I think the remaining non-slash `req.ID` passthrough should be removed or changed to a clear diagnostic error.
2. `ParseCompositeID` should reject version `0` and negative versions.

After those two tweaks, this looks good to me.

internal/resources/backend/resource.go
anthony-gomez-fastly: fixed in `cf904ec7`

internal/importutil/composite_id.go
anthony-gomez-fastly: fixed in `cf904ec7`

internal/importutil/composite_id_test.go
anthony-gomez-fastly: added in `cf904ec7`

philippschulte (approved):
Looks good to me assuming the unit and targeted acceptance tests pass.

anthony-gomez-fastly:
Screenshot 2026-06-12 at 3 39 55 PM
Screenshot 2026-06-12 at 3 41 13 PM
Screenshot 2026-06-12 at 3 46 04 PM

PR #1298, opened 2026-06-16
Replace SetAttribute calls with Resource.Set

anthony-gomez-fastly:
### Change summary

Replace per-attribute SetAttribute pattern with single Resource.Set call in backend and domain list resources. This simplifies the code by building the complete Model struct and setting it in one operation, while reusing existing FlattenToNestedModel helpers.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-06-16 at 11 00 59 AM


internal/resources/backend/list.go
```diff
@@ -123,41 +124,44 @@ func setResourceAttrs(ctx context.Context, result *list.ListResult, b *fastly.Ba
...
-	diags.Append(result.Resource.SetAttribute(ctx, path.Root("ssl_check_cert"), model.SSLCheckCert)...)
-	diags.Append(result.Resource.SetAttribute(ctx, path.Root("ssl_ciphers"), model.SSLCiphers)...)
-	diags.Append(result.Resource.SetAttribute(ctx, path.Root("ssl_client_cert"), model.SSLClientCert)...)
-	diags.Append(result.Resource.SetAttribute(ctx, path.Root("ssl_client_key"), model.SSLClientKey)...)
-	diags.Append(result.Resource.SetAttribute(ctx, path.Root("ssl_sni_hostname"), model.SSLSNIHostname)...)
-	diags.Append(result.Resource.SetAttribute(ctx, path.Root("use_ssl"), model.UseSSL)...)
-	diags.Append(result.Resource.SetAttribute(ctx, path.Root("weight"), model.Weight)...)
+	nestedModel := FlattenToNestedModel(b)
+
+	model := Model{
```
rcaril, line 128: Good start! However, I believe we only need the identifying fields here as the rest of the backend attributes are in `flatten.go`. 

We can rewrite this block as:

```
model := Model{
      ID:      types.StringValue(id),
      Service: types.StringValue(serviceID),
      Version: types.Int64Value(int64(version)),
  }
  ApplyNestedToModel(FlattenToNestedModel(b), &model)
  diags.Append(result.Resource.Set(ctx, &model)...)
```

rcaril (dismissed):
Pretty solid! `domain/list.go` looks good.
  
 In `backend/list.go` - the struct literal still restates all 28 backend fields, just in a different form. `ApplyNestedToModel` and `FlattenToNestedModel` already exist for this purpose. I suggested a change below.

internal/resources/backend/list.go
anthony-gomez-fastly: done in `ee7f9644`

rcaril (approved):
Looks perfect now - thank you!

PR #1300, opened 2026-06-16
refactor(internal/resources): consolidate 404 checking into errors.IsNotFound

anthony-gomez-fastly:
### Change summary

Replace inline *fastly.HTTPError type assertions with wrapped-error-safe
  errors.IsNotFound helper. Move implementation from internal/client to new
  internal/errors package to avoid import cycles.

  Changes:
  - Create internal/errors/http.go with canonical IsNotFound implementation using errors.As
  - Remove internal/client/errors.go (was forwarding wrapper)
  - Delete private service.isNotFound function (used bare type assertion)
  - Update 18 call sites across resources, service, computepackage, and tests to use errors.IsNotFound

 

 All Submissions:

### New Feature Submissions:

### DIFF
--- a/internal/errors/http.go
+++ b/internal/errors/http.go
@@ -1,4 +1,4 @@
-package apierrors
+package errors
 
 import (
 	"errors"
--- a/internal/errors/http.go
+++ b/internal/errors/http.go
@@ -1,4 +1,4 @@
-package client
+package errors
 
 import (
 	"errors"
--- a/internal/acceptance_tests/test_helpers.go
+++ b/internal/acceptance_tests/test_helpers.go
@@ -8,7 +8,7 @@ import (
 	"testing"
 
 	"github.com/fastly/go-fastly/v15/fastly"
-	"github.com/fastly/terraform-provider-fastly/internal/errors"
+	"github.com/fastly/terraform-provider-fastly/internal/apierrors"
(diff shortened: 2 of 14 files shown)


internal/errors/http.go
```diff
@@ -1,4 +1,4 @@
-package client
```
rcaril, line 1: I think we want to keep this inside of `internal/client/errors.go`. 

From my research, moving to a new package `errors` may collide with the `stdlib` errors package, which will result in an alias needed whenever we import.

Was there a given motivation for moving this? If not, we should consider keeping it inside of `client`.

rcaril (dismissed):
Looks good, just one item to review. Thanks for your efforts here!

internal/errors/http.go
anthony-gomez-fastly: good point, i was thinking it was cleaner there but you're right

anthony-gomez-fastly: oh there was a dependency cycle, that was why.

anthony-gomez-fastly: I did rename it to be `apierrors` which should at least fix that collision

rcaril: As per our chat, this is a moot point as we currently don't reference `stdlib` in the current provider.

rcaril (approved):
LGTM !

PR #1302, opened 2026-06-17
Consolidate defaults, add versioned UserAgent, and replace fastly.ToPointer

anthony-gomez-fastly:
### Change summary

- feat(provider): add versioned UserAgent with mode indicator
    * Create internal/version package with Version variable set via ldflags
    * Modify Makefile to inject git version at build time
    * Set provider UserAgent in format: terraform-provider-fastly/<version> FastlyGo/...
    * Add "mode=auto" suffix to UserAgent for auto-lifecycle resources

- refactor(backend): hoist duplicated defaults to package constants
    * Add 29 constants (DefaultPort, DefaultAutoLoadbalance, DefaultBetweenBytesTimeout, etc.)
    * Update schema.go and flatten.go to reference shared constants
    * Eliminates drift between schema defaults and flatten fallback values

- refactor: replace fastly.ToPointer with new

- refactor(resources): add compile-time interface assertions
    * Add ResourceWithConfigure assertion to servicecdn and servicecompute resources

 All Submissions:

### New Feature Submissions:

Screenshot 2026-06-17 at 10 50 24 AM

### Changes to Core Features:


internal/computepackage/package.go
```diff
@@ -156,7 +156,7 @@ func Update(ctx context.Context, client *fastly.Client, serviceID string, versio
 	}
 
 	if !pkg.Filename.IsNull() && !pkg.Filename.IsUnknown() && pkg.Filename.ValueString() != "" {
-		input.PackagePath = fastly.ToPointer(pkg.Filename.ValueString())
+		input.PackagePath = new(pkg.Filename.ValueString())
```
anthony-gomez-fastly, line 159: this is built in to go 1.26 🥳

internal/client/data.go
```diff
@@ -23,6 +24,43 @@ func NewData(client *fastly.Client) *Data {
...
+func (t *userAgentTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	ua := req.Header.Get("User-Agent")
+	if ua == "" {
+		ua = fastly.UserAgent
+	}
+	req.Header.Set("User-Agent", ua+" "+t.suffix)
+	return t.base.RoundTrip(req)
+}
+
+func (d *Data) AutoClient() *fastly.Client {
```
anthony-gomez-fastly, line 68: this was suggested by Jas in our 1 on 1 so i added it in this PR. It will allow us to identify requests that use auto mode

internal/client/data.go
```diff
@@ -23,6 +24,43 @@ func NewData(client *fastly.Client) *Data {
...
+	base   http.RoundTripper
+	suffix string
+}
+
+func (t *userAgentTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	ua := req.Header.Get("User-Agent")
+	if ua == "" {
+		ua = fastly.UserAgent
+	}
+	req.Header.Set("User-Agent", ua+" "+t.suffix)
```
philippschulte: I would not recommend to modify the request directly. Please clone the request first before changing headers, for example:

```go
req = req.Clone(req.Context())
req.Header.Set("User-Agent", ua+" "+t.suffix)
```

internal/client/data.go
```diff
@@ -23,6 +24,43 @@ func NewData(client *fastly.Client) *Data {
...
+	return t.base.RoundTrip(req)
+}
+
+func (d *Data) AutoClient() *fastly.Client {
+	client := d.Client
+	if client.HTTPClient == nil {
+		client.HTTPClient = &http.Client{Transport: http.DefaultTransport}
```
philippschulte: I believe that can potentially be an issue because `AutoClient()` currently mutates the shared provider client when `HTTPClient` is `nil`.

Since `d.Client` is the provider-wide `*fastly.Client`, `client := d.Client` copies the pointer, not the underlying client value.

I would recommend to copy the Fastly client value before assigning a wrapped `HTTPClient`. Something like this:

```go
func (d *Data) AutoClient() *fastly.Client {
	base := d.Client

	baseHTTPClient := base.HTTPClient
	if baseHTTPClient == nil {
		baseHTTPClient = http.DefaultClient
	}

	baseTransport := baseHTTPClient.Transport
	if baseTransport == nil {
		baseTransport = http.DefaultTransport
	}

	wrapped := *base
	wrapped.HTTPClient = &http.Client{
		Transport: &userAgentTransport{
			base:   baseTransport,
			suffix: "mode=auto",
		},
		Timeout: baseHTTPClient.Timeout,
	}

	return &wrapped
}
```

internal/provider/provider.go
```diff
@@ -79,6 +81,9 @@ func (p *fastlyProvider) Configure(ctx context.Context, req provider.ConfigureRe
 		return
 	}
 
+	defaultUserAgent := fastly.UserAgent
+	fastly.UserAgent = fmt.Sprintf("terraform-provider-fastly/%s %s", version.Version, defaultUserAgent)
```
philippschulte: I don't know if this is concurrency-safe? I think `Configure` can run more than once in the same Terraform process.

I would recommend to avoid changing `fastly.UserAgent` globally and instead add the provider User-Agent at the HTTP transport/request layer basically the same pattern you introduced for `mode=auto`.

Then the normal client can use:

```go
prefix: fmt.Sprintf("terraform-provider-fastly/%s", version.Version)
```

and the auto client can use:

```go
prefix: fmt.Sprintf("terraform-provider-fastly/%s", version.Version),
suffix: "mode=auto",
```

With the above we would have no global mutation, no potential string stacking, and the User-Agent is still observable in API requests.

philippschulte (dismissed):
Thanks for putting this together. I checked out the branch locally and confirmed that `make build` and `go test ./internal/...` pass.

I left some comments I would like you to think about and to consider before merging.

Small scope question: It looks like that the following ticket bullet is not addressed in this PR:

* Removes ModelToNested + ApplyNestedToModel per resource.

internal/client/data.go
anthony-gomez-fastly: fixed in `89ace84f`

internal/client/data.go
anthony-gomez-fastly: fixed in `28c0e2f4`

internal/provider/provider.go
anthony-gomez-fastly: addressed in `0cd0529d`

philippschulte (approved):
Thanks for the update. The three UserAgent/client concerns I raised are addressed now, and the new per-client/per-request UserAgent approach looks much better to me.

I am still not fully convinced about the benefit of removing `ModelToNested` and `ApplyNestedToModel` in the current shape, since the replacement seems to duplicate the same field-by-field mapping directly in multiple call sites. A shared copy helper would be easier to maintain long term.

That said, I do not want to block this PR further on that point. Thanks for the improvements — this looks good to me.

anthony-gomez-fastly:
I reverted the model change, i agree that the extra code duplication wasn't worth it. ty!

PR #1312, opened 2026-06-23
Introduce generic reconciliation framework

anthony-gomez-fastly:
### Change summary

Introduce internal/reconcile package with generic Resource[T, API] to eliminate
  duplicated list→diff→create/update/delete patterns across backend and domain resources.

  Changes:
  - Add internal/reconcile package with generic Resource[T, API] reconciler
    * Run() method for reconciliation (list, diff, create/update/delete)
    * ReadForVersion() method for reading and converting API objects
    * ModelsEqual() helper for comparing slices with optional sorting
    * Comprehensive test suite covering create, update, delete, edge cases

  - Refactor backend resource to use reconciler
    * Move reconcile adapter ops into schema.go (consolidate files)
    * Reduce ReadForVersion from 17 lines to 1 line (delegates to reconciler)
    * Reduce Reconcile from 59 lines to 1 line (delegates to reconciler)
    * Reduce Equal from 13 lines to 1 line (uses reconcile.ModelsEqual)
    * Remove Normalize function (handled internally by reconciler)
    * Remove TestNormalize (behavior now tested via reconciler)
    * Replace map copy loop with maps.Copy

  - Refactor domain resource to use reconciler
    * Move reconcile adapter ops into schema.go (consolidate files)
    * Reduce ReadForVersion from 17 lines to 1 line (delegates to reconciler)
    * Reduce Reconcile from 79 lines to 1 line (delegates to reconciler)
    * Reduce Equal from 28 lines to 1 line + 16-line helper (uses reconcile.ModelsEqual)
    * Remove Normalize function (handled internally by reconciler)
    * Remove TestNormalize (behavior now tested via reconciler)
    * Replace map copy loop with maps.Copy

 

 All Submissions:

### New Feature Submissions:

Screenshot 2026-06-23 at 11 21 48 AM


internal/acceptance_tests/config_builder.go
```diff
@@ -67,9 +68,7 @@ func BuildConfig(serviceType ServiceType, replacements map[string]string, blockP
 
 	// Build template data with all replacements
 	templateData := make(map[string]string)
-	for k, v := range replacements {
-		templateData[k] = v
-	}
+	maps.Copy(templateData, replacements)
```
anthony-gomez-fastly, line 71: this was added in 16.4, decided to update since i was touching files with the same nit

internal/reconcile/reconcile.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (r *Resource[T, API]) Run(ctx context.Context, client *fastly.Client, serviceID string, version int, desired []T) error {
+	remote, err := r.Ops.List(ctx, client, serviceID, version)
+	if err != nil {
+		return err
+	}
+
+	if r.Sortable {
+		sort.Slice(desired, func(i, j int) bool {
```
philippschulte: This can mutate the caller’s slice unexpectedly. Since this reconciler is generic and may be reused in more places later, it would be safer to copy before sorting

philippschulte (dismissed):
Nice refactor.

One small request before merge: please copy the slice before sorting in `Resource.Run()`. Sorting the `desired` slice in place can mutate the caller’s backing array, which could lead to unexpected ordering changes outside the reconciler.

Once this is addressed, this is good to merge from my side.

Thanks!

internal/reconcile/reconcile.go
anthony-gomez-fastly: fixed in `0c079919`

philippschulte (approved):
Thank you!

internal/resources/domain/schema.go
```diff
@@ -59,153 +57,125 @@ func NestedBlockSchema() schema.ListNestedBlock {
...
+	name := desired.Name.ValueString()
+	input := &fastly.CreateDomainInput{
 		ServiceID:      serviceID,
 		ServiceVersion: version,
-	})
-	if err != nil {
-		return err
+		Name:           &name,
+	}
+	if !desired.Comment.IsUnknown() {
```
rcaril: Should we also add a `IsNull` guard here like we have on line `83`?

internal/reconcile/reconcile.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]T, 0, len(remote))
+	for _, item := range remote {
+		result = append(result, r.Ops.ToModel(item))
+	}
+
+	if r.Sortable {
```
rcaril, line 97: Sortable conflicts with CDTOOL-1665:

> The Sortable: true flag causes ReadForVersion to sort results before writing them to Terraform state. This is the same issue raised in CDTOOL-1665 -  `ListNestedBlock` is positional, so if a user configures `backend[0] = b`, `backend[1] = a`, the provider cannot reorder and return `[a, b] `in state without Terraform throwing "Provider produced inconsistent result after apply." The sort in Run is also a no-op - the reconcile loop is entirely map-based (remoteByName/desiredByName), so iteration order of desired never affects which items get created, updated, or deleted. Sortable belongs only in `ModelsEqual` where you control both sides for internal comparison.

We should consider holding off on this PR until we update the sorting issue first.

internal/reconcile/reconcile_test.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	return o.listResult, nil
+}
+
+func (o *testOps) GetName(api *testAPI) string {
+	return api.Name
+}
+
+func (o *testOps) Delete(ctx context.Context, client *fastly.Client, serviceID string, version int, name string) error {
+	if o.deleteError != nil {
```
rcaril, line 45: Should we add tests for the `deleteError`, `createError` and `updateError` methods? I believe we already do this elsewhere such as on line `170`:

```
t.Run("handles list error", func(t *testing.T) {
		ops := &testOps{
			listError: errors.New("list failed"),
		}
```
                                                                                                                               
 Perhaps something like:                                                                                                                           
  ```                                                                                                                          
  t.Run("returns error when delete fails", func(t *testing.T) {                                                                
        ops := &testOps{                                                                                                       
            listResult:  []*testAPI{{Name: "item1"}},                                                                          
            deleteError: errors.New("delete failed"),                                                                          
        }                                                                                                                      
        // desired is empty so delete is triggered                                                                             
        err := r.Run(ctx, nil, "service123", 1, []testModel{})                                                                 
        assert.Error(t, err)                                                                                                   
    })                                                                                                                         
  ```

rcaril (dismissed):
Great work! A few items that may need to be reviewed below.

internal/reconcile/reconcile.go
philippschulte: @rcaril, @anthony-gomez-fastly I'll address CDTOOL-1665 as soon as this pr is merged. I am just waiting for this pull request to be merged. Thanks

internal/reconcile/reconcile_test.go
anthony-gomez-fastly: addressed in `fcc189f2`

internal/resources/domain/schema.go
anthony-gomez-fastly: addressed in `a09639db`

PR #1315, opened 2026-06-24
[CDTOOL-1498] Add ACL resource with explicit and automatic resource support

anthony-gomez-fastly:
### Change summary

Add new fastly_service_acl resource with full CRUD operations and list
  resource support. Includes integration with _auto service resources
  (cdn_auto, compute_auto) via nested block pattern.

  - Add ACL resource package with schema, expand, flatten, and list
  - Implement NestedModel pattern for _auto resource integration
  - Add comprehensive unit and acceptance tests
  - Update service_cdn_auto and service_compute_auto with ACL blocks
  - Add ACL test helpers and fixtures
  - Update lifecycle test scripts with ACL examples
  - Generate documentation for resources and list-resources

 All Submissions:

### New Feature Submissions:

Screenshot 2026-06-26 at 11 00 26 AM
Screenshot 2026-06-26 at 11 02 05 AM
Screenshot 2026-06-26 at 11 03 59 AM


internal/resources/acl/resource.go
```diff
@@ -0,0 +1,392 @@
...
+type Model struct {
+	NestedModel
+	ID      types.String `tfsdk:"id"`
+	Service types.String `tfsdk:"service_id"`
+	Version types.Int64  `tfsdk:"version"`
+}
+
+type ACLIdentityModel struct {
+	ServiceID types.String `tfsdk:"service_id"`
+	Name      types.String `tfsdk:"name"`
```
philippschulte, line 44: I think the ACL resource identity needs to include `version`.

The resource schema requires `service_id`, `version`, and `name`, and the list resource is already setting `version` on the identity. But `ACLIdentityModel` / `IdentitySchema` only include `service_id` and `name`, and the identity import path only writes those two fields into state.

That seems inconsistent and likely breaks list-resource generated imports or identity-based import, because `Read` needs `version` to call `GetACL`.

Can we add `version` to the identity model/schema and make the identity import path hydrate the full resource state?

internal/resources/acl/resource.go
```diff
@@ -0,0 +1,392 @@
...
+		resp.Diagnostics.AddError("Unsupported Fastly service type", err.Error())
+		return
+	}
+
+	resp.Diagnostics.Append(r.providerData.VersionChecker.EnsureMutable(ctx, plan.Service.ValueString(), int(plan.Version.ValueInt64()))...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Check if name changed
```
philippschulte: I think `service_id` and `version` need replacement semantics, or Update needs to handle changes to them.

Right now Update only recreates the ACL if `name` changes. If `version` or `service_id` changes while the name stays the same, the code goes down the “No ACL changes detected” path and preserves `ID` / `ACLID` from state. That can leave the resource state pointing at the old service/version.

Can we either add `RequiresReplace` plan modifiers for `service_id` / `version` / maybe `name`, or add explicit handling for those identity changes?

internal/resources/servicecdnauto/resource.go
```diff
@@ -342,8 +375,10 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
 		// No version change needed - preserve version numbers and order nested state to match the plan
 		plan.ManagedVersion = state.ManagedVersion
 		plan.ActiveVersion = state.ActiveVersion
+
 		plan.Domain = domain.MatchOrder(state.Domain, plan.Domain)
 		plan.Backend = backend.MatchOrder(state.Backend, plan.Backend)
+		plan.ACL = acl.MatchOrder(state.ACL, plan.ACL)
```
philippschulte, line 381: Can we add a test for changing only `acl.force_destroy` on `cdn_auto` / `compute_auto`?

`force_destroy` is config-only and `ModelsEqual` intentionally ignores it, so changing it should probably update Terraform state without creating a new service version. But in the no-version-change path, we do `plan.ACL = acl.MatchOrder(state.ACL, plan.ACL)`, which looks like it may preserve the old state value and drop the new configured `force_destroy`.

Maybe we need a helper that matches remote/state order but preserves config-only fields from the plan.

philippschulte (dismissed):
Thanks for the work on this. The overall direction looks good to me,.

Comparing this with the existing SDK-provider ACL acceptance test, I think we are missing one important behavior: non-empty ACL handling with `force_destroy`.

The current PR tests cover creating ACLs, multiple ACLs, backend + ACL combinations, and ACL rename/update for both `cdn_auto` and `compute_auto`. They also assert the `force_destroy` state value in some cases.

What I don’t see covered is the old-provider flow where we:
1. create ACLs,
2. add ACL entries directly through the Fastly API,
3. try to rename/delete the non-empty ACL without `force_destroy` and expect it to fail,
4. then set `force_destroy = true` and verify the rename/delete succeeds.

That seems worth keeping because it validates the main behavioral purpose of `force_destroy`, not just that the field is present in state.

internal/resources/acl/resource.go
anthony-gomez-fastly: that makes sense to me! adding

internal/resources/acl/resource.go
anthony-gomez-fastly: i was unsure how to handle changes to versions because i wasn't sure if we wanted a full recreate on non name changes, ty for clarifying! i'll change this

internal/resources/servicecdnauto/resource.go
anthony-gomez-fastly: you nailed it, added tests and and a new reconcile function for auto mode in `6095bf51`

internal/resources/acl/resource.go
anthony-gomez-fastly: fixed in `469972c9`

philippschulte (commented):
I think this is almost there. The only remaining concern I have is for the standalone `fastly_service_acl` resource.

Since this PR supports the explicit/default resource family too, I think `service_id`, `version`, and probably `name` need replacement semantics, or the `Update` method needs to explicitly handle changes to all three fields.

For nested `_auto` ACL blocks this is not an issue, because the provider owns the service/version lifecycle. But for the explicit resource, these fields are user-configured and together identify the ACL location:

`service_id + version + name`

Right now it looks like `Update` only recreates the ACL when `name` changes. If `service_id` or `version` changes while `name` stays the same, I think the code can preserve the old `id` / `acl_id` from state even though the ACL now points at a different service/version.

Could we add replacement plan modifiers for `service_id`, `version`, and maybe `name`, or update the `Update` path to recreate/read the ACL when any of those identity fields change?

internal/resources/acl/resource.go
anthony-gomez-fastly: @philippschulte i made changes to add require update to ACL's, but it makes the lifecycle flow a little icky, requiring state modification and imports, what are your thoughts? in `23aeb8fa`

philippschulte (approved):
This looks great, thank you!

scripts/test-lifecycle-compute/run.sh
```diff
@@ -446,8 +446,22 @@ test_clone_from_latest_and_version_writes() {
     # Test version-locked resource writes
     # Add a new domain and backend to the latest version
     log_info "Testing version-locked resource writes on version $svc1_latest..."
+
+    # Remove version-locked resources from state before changing versions
+    # These resources are pinned to version 1 which is now locked, so we need to remove them
+    # from state, update the version in config, and then import them from the new versions
+    log_info "Removing version-locked resources from state before version change..."
```
anthony-gomez-fastly, line 453: This was what i was worried about @philippschulte

PR #1317, opened 2026-06-25
[CDTOOL-1665] Preserve _auto backend and domain list ordering

philippschulte:
### Change summary

Preserves Terraform list ordering for nested `backend` and `domain` blocks on `_auto` resources.

See https://fastly.atlassian.net/browse/CDTOOL-1665 for more info.

### Reviewer Notes

- Keeps unordered comparison/reconcile behavior through the existing generic reconciler.
- Adds `reconcile.MatchOrder()` for ordering read results against plan/state.
- Orders create/update API-read results to match the planned configuration order.
- Orders read API results to match prior state order.
- Appends API-only/new elements in stable name order.
- Handles reorder-only HCL changes without requiring a Fastly version change by reordering existing nested state to match the plan.

**Also updated `BuildConfig` to indent inserted block snippets consistently. This allows block fixture files to start at column 0 while still rendering valid nested Terraform blocks inside the generated resource config.**

### Changes to Core Features:

```
=== RUN   TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
=== PAUSE TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
=== RUN   TestAccFastlyServiceComputeAuto_preservesBackendAndDomainOrder
=== PAUSE TestAccFastlyServiceComputeAuto_preservesBackendAndDomainOrder
=== CONT  TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
=== CONT  TestAccFastlyServiceComputeAuto_preservesBackendAndDomainOrder
--- PASS: TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder (36.15s)
--- PASS: TestAccFastlyServiceComputeAuto_preservesBackendAndDomainOrder (43.77s)
```

### DIFF
--- a/internal/acceptance_tests/service_cdn_auto_test.go
+++ b/internal/acceptance_tests/service_cdn_auto_test.go
@@ -176,6 +176,7 @@ func TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder(t *testing.T) {
 	domainBName := fmt.Sprintf("b-%s.example.com", acctest.RandString(10))
 	domainAName := fmt.Sprintf("a-%s.example.com", acctest.RandString(10))
 	config := ConfigCDNAutoUnsortedBackendAndDomainBlocks(serviceName, domainBName, domainAName)
+	reversedConfig := ConfigCDNAutoReversedBackendAndDomainBlocks(serviceName, domainBName, domainAName)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:                 func() { PreCheck(t) },
@@ -194,10 +195,32 @@ func TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder(t *testing.T) {
 					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "domain.#", "2"),
 					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "domain.0.name", domainBName),
 					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "domain.1.name", domainAName),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "active_version", "1"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "managed_version", "1"),
 				),
 			},
 			{
 				Config:   config,
+				PlanOnly: true,
+			},
+			{
+				Config: reversedConfig,
+				Check: resource.ComposeTestCheckFunc(
+					CheckServiceExists("fastly_service_cdn_auto.test"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "backend.#", "2"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "backend.0.name", "a"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "backend.0.address", "a.example.com"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "backend.1.name", "b"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "backend.1.address", "b.example.com"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "domain.#", "2"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "domain.0.name", domainAName),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "domain.1.name", domainBName),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "active_version", "1"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "managed_version", "1"),
+				),
+			},
+			{
+				Config:   reversedConfig,
 				PlanOnly: true,
 			},
 		},
--- a/internal/acceptance_tests/service_cdn_auto_test.go
+++ b/internal/acceptance_tests/service_cdn_auto_test.go
@@ -170,6 +170,40 @@ func TestAccFastlyServiceCDNAuto_multipleBackends(t *testing.T) {
 	})
 }
 
+func TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder(t *testing.T) {
+	t.Parallel()
+	serviceName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	domainBName := fmt.Sprintf("b-%s.example.com", acctest.RandString(10))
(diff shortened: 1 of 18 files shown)


internal/resources/servicecdnauto/resource.go
```diff
@@ -144,7 +144,7 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp
 		resp.Diagnostics.AddError("Error reading service domains", err.Error())
 		return
 	}
-	plan.Domain = domains
+	plan.Domain = domain.MatchOrder(domains, plan.Domain)
```
anthony-gomez-fastly, line 147: do we need to remove the sortable stuff from reconcile as well or can they coexist?

philippschulte: They can coexist.

`Sortable` is still useful inside `reconcile` for unordered backend/domain reconcile behavior and equality checks. It gives us deterministic diff/create/update/delete behavior because backend/domain are logically keyed by name for reconciliation.

The CDTOOL-1665 issue is specifically about writing the sorted API read result directly back into Terraform state for `ListNestedBlock` fields. That is why this PR keeps `Sortable` in the reconciler, but adds `MatchOrder()` at the `_auto` resource boundary before assigning read results into state/plan:

So the split is:

- `Sortable`: internal reconcile/equality behavior
- `MatchOrder`: Terraform state ordering behavior

I think removing `Sortable` would make the reconciler less deterministic and would change existing unordered comparison/reconcile semantics. The important requirement is that sorted `ReadForVersion()` results are not assigned directly to Terraform state for ordered nested lists.

internal/resources/servicecdnauto/resource.go
```diff
@@ -156,7 +156,7 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp
 		resp.Diagnostics.AddError("Error reading service backends", err.Error())
 		return
 	}
-	plan.Backend = backends
+	plan.Backend = backend.MatchOrder(backends, plan.Backend)
```
rcaril, line 159: Will the existing test catch this change? I believe that  the existing `_multipleBackends` test uses alphabetically ordered names so it may not catch a regression here.

philippschulte: Good callout. The existing `_multipleBackends` test uses alphabetically ordered names, so it would not catch this specific regression.

That is why this PR adds a new regression case for `fastly_service_cdn_auto` where both nested blocks are intentionally declared in non-sorted order:

- backend: `b`, then `a`
- domain: `b.example.com`, then `a.example.com`

The test also includes a plan-only follow-up step, so it verifies both apply succeeds and the subsequent plan is clean.

internal/acceptance_tests/service_cdn_auto_test.go
```diff
@@ -170,6 +170,40 @@ func TestAccFastlyServiceCDNAuto_multipleBackends(t *testing.T) {
 	})
 }
 
+func TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder(t *testing.T) {
```
rcaril, line 173: Is is worth adding an Update test step that applies the reverse ordered config and checks that no new version is created and the new order is reflected in state?

internal/resources/backend/schema.go
rcaril, line 366: Is this still needed now?

internal/resources/domain/schema.go
rcaril, line 150: Same here - is this still used?

internal/resources/servicecdnauto/resource.go
rcaril: Ah, this was meant for `servicecomputeauto` instead - apologies. Did we need to add check for compute as well?

internal/resources/backend/schema.go
philippschulte: Yes, `Sortable: true` is still used.

It keeps backend/domain reconcile and equality behavior order-insensitive by sorting collections by name before comparing/reconciling them. For example, for reconciliation purposes `[backend "b", backend "a"]` and `[backend "a", backend "b"]` should represent the same backend set if the objects themselves are unchanged.

The CDTOOL-1665 fix does not remove that behavior. It only prevents sorted reconciler output from being written directly into Terraform `ListNestedBlock` state. `MatchOrder()` restores the plan/prior-state order at the resource boundary before assigning the read results into state.

internal/resources/domain/schema.go
philippschulte: Yes, `Sortable: true` is still needed.

internal/acceptance_tests/service_cdn_auto_test.go
philippschulte: Yes, good call! Test added in 1b851b769d0ee09bf07a52f6dfd2b6d67d1157fb.

```
==> Running _auto ordering acceptance test...
=== RUN   TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
=== PAUSE TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
=== CONT  TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
--- PASS: TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder (27.19s)
```

philippschulte: We can update the name of the test later when more block resources are being added.

rcaril: Thanks!

internal/resources/servicecdnauto/resource.go
philippschulte: Yes, test added here: cbf00f07ca32fdd1aaae0ce32c083a7cc2c7ffa1

```
=== RUN   TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
=== PAUSE TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
=== RUN   TestAccFastlyServiceComputeAuto_preservesBackendAndDomainOrder
=== PAUSE TestAccFastlyServiceComputeAuto_preservesBackendAndDomainOrder
=== CONT  TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder
=== CONT  TestAccFastlyServiceComputeAuto_preservesBackendAndDomainOrder
--- PASS: TestAccFastlyServiceCDNAuto_preservesBackendAndDomainOrder (36.15s)
--- PASS: TestAccFastlyServiceComputeAuto_preservesBackendAndDomainOrder (43.77s)
```

rcaril: Thank you!

rcaril (approved):
LGTM!

anthony-gomez-fastly (approved):
LGTM, ty!

PR #1318, opened 2026-06-26
standardize lifecycle test outputs and add more descriptive headings

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

### DIFF
--- a/scripts/test-lifecycle-compute/run.sh
+++ b/scripts/test-lifecycle-compute/run.sh
@@ -632,8 +632,21 @@ main() {
     test_clone_from_latest_and_version_writes
     test_resource_destruction
 
-    log_step "All tests passed!"
-    log_success "Compute service lifecycle tests completed successfully"
+    log_step "Test Summary - Compute Service"
+    log_success "✓ Provider build"
+    log_success "✓ Service creation (fastly_service_compute)"
+    log_success "✓ Domain attachment (fastly_service_domain)"
+    log_success "✓ Backend configuration (fastly_service_backend)"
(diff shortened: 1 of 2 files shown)


scripts/test-lifecycle-compute/run.sh
```diff
@@ -632,8 +632,21 @@ main() {
     test_clone_from_latest_and_version_writes
     test_resource_destruction
 
-    log_step "All tests passed!"
-    log_success "Compute service lifecycle tests completed successfully"
+    log_step "Test Summary - Compute Service"
+    log_success "✓ Provider build"
+    log_success "✓ Service creation (fastly_service_compute)"
+    log_success "✓ Domain attachment (fastly_service_domain)"
+    log_success "✓ Backend configuration (fastly_service_backend)"
```
rcaril, line 639: Is the intention here to have an entry for each resource? If so, is this going to be something we can easily maintain?

anthony-gomez-fastly: my idea was that this be the source of truth for lifecycle tests, so if we want to add to the test we should document that. also makes it easy to post test run outputs (which is what i was doing when i noticed this)

anthony-gomez-fastly: so yes, this output should explicitly state which things are covered in it

rcaril: Fair enough, we should ensure to add a note about this in the contributor docs when we go about creating one for this branch.

rcaril (approved):

PR #1319, opened 2026-06-29
[CDTOOL-1666] Experiment with nested backend SSL client secrets

philippschulte:
### Change summary

This PR moves the backend client TLS certificate and key out of the top-level backend schema and into a nested `ssl_client_secrets` attribute.

Before:

```hcl
backend {
  ssl_client_cert = file("${path.module}/client.crt")
  ssl_client_key  = file("${path.module}/client.key")
}
```

After:

```hcl
backend {
  ssl_client_secrets = {
    ssl_client_cert = file("${path.module}/client.crt")
    ssl_client_key  = file("${path.module}/client.key")
  }
}
```

The goal is to improve Terraform plan output when backend client TLS credentials are configured. With the Terraform Plugin Framework, grouping the sensitive certificate/key fields into their own nested attribute creates a narrower sensitive boundary. Local testing confirms that non-sensitive backend fields remain visible in the plan while only the nested `ssl_client_secrets` values are redacted.

## Why

In the current SDK-based provider, sensitive attributes inside a backend block can cause Terraform to hide more of the backend diff than desired. That made normal backend changes hard for customers to review, because unrelated public fields could be hidden along with the sensitive certificate/key values.

The old provider introduced an environment variable, `FASTLY_TF_DISPLAY_SENSITIVE_FIELDS`, to allow users to opt out of sensitivity masking for affected fields. That was useful as a workaround for the SDK/provider behavior, but it also intentionally exposes secret values in Terraform output.

In this dual-model provider, the Plugin Framework gives us a better option: keep the secrets marked sensitive, but isolate them inside `ssl_client_secrets`. This preserves secret redaction while still allowing Terraform to display the rest of the backend diff.

### Terraform config

```hcl
resource "fastly_service_cdn_auto" "test" {
  name = "tf-local-nested-backend-secrets-test"

  domain {
    name = "tf-local-nested-backend-secrets-test.example.com"
  }

  backend {
    name               = "origin"
    address            = "httpbin.org"
    port               = 443
    use_ssl            = true
    first_byte_timeout = 16000
    ssl_check_cert     = true
    ssl_sni_hostname   = "httpbin.org"
    ssl_cert_hostname = "httpbin.org"

    override_host = "httpbin.org"

    ssl_client_secrets = {
      ssl_client_cert = file("${path.module}/client.crt")
      ssl_client_key  = file("${path.module}/client.key")
    }
  }
}
```

### Plan output

```text

  # fastly_service_cdn_auto.test will be created
  + resource "fastly_service_cdn_auto" "test" {
      + active_version  = (known after apply)
      + comment         = "Managed by Terraform"
      + force_destroy   = false
      + id              = (known after apply)
      + managed_version = (known after apply)
      + name            = "tf-local-nested-backend-secrets-test"
      + reuse           = false

      + backend {
          + address               = "httpbin.org"
          + auto_loadbalance      = false
          + between_bytes_timeout = 10000
          + connect_timeout       = 1000
          + error_threshold       = 0
          + first_byte_timeout    = 16000
          + healthcheck           = ""
          + keepalive_time        = (known after apply)
          + max_conn              = 200
          + max_lifetime          = 0
          + max_tls_version       = ""
          + max_use               = 0
          + min_tls_version       = ""
          + name                  = "origin"
          + override_host         = "httpbin.org"
          + port                  = 443
          + prefer_ipv6           = false
          + request_condition     = ""
          + share_key             = ""
          + shield                = ""
          + ssl_ca_cert           = ""
          + ssl_cert_hostname     = "httpbin.org"
          + ssl_check_cert        = true
          + ssl_ciphers           = ""
          + ssl_client_secrets    = {
              + ssl_client_cert = (sensitive value)
              + ssl_client_key  = (sensitive value)
            }
          + ssl_sni_hostname      = "httpbin.org"
          + use_ssl               = true
          + weight                = 100
        }

      + domain {
          + name = "tf-local-nested-backend-secrets-test.example.com"
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

The important behavior is that public backend attributes are still visible:

```text
+ first_byte_timeout    = 16000
+ ssl_cert_hostname     = "httpbin.org"
+ ssl_check_cert        = true
+ ssl_sni_hostname      = "httpbin.org"
+ use_ssl               = true
```

while only the nested secret values are redacted:

```text
+ ssl_client_secrets    = {
    + ssl_client_cert = (sensitive value)
    + ssl_client_key  = (sensitive value)
  }
```

### Changes to Core Features:

```
PASS
ok  	github.com/fastly/terraform-provider-fastly/internal/acceptance_tests	218.456s
```


internal/resources/backend/schema.go
```diff
@@ -255,19 +303,26 @@ func CommonAttributes() map[string]schema.Attribute {
...
 			Optional:    true,
 			Computed:    true,
-			Sensitive:   true,
-			Default:     stringdefault.StaticString(DefaultSSLClientKey),
-			Description: "Client key used when connecting to the backend.",
+			Description: "Sensitive client certificate and key values used when connecting to the backend.",
+			Attributes: map[string]schema.Attribute{
+				"ssl_client_cert": schema.StringAttribute{
+					Optional:    true,
+					Computed:    true,
```
kpfleming, line 313: I realize this is not new behavior in this PR, but what is the purpose of marking these attributes as "Computed"?

philippschulte: Good question. I tested this locally by removing `Computed`, but the acceptance tests then failed with perpetual non-refresh plan diffs for backends.

These fields are effectively write-only from Terraform’s perspective: the user can configure them and we send them to the Fastly API, but we can’t rely on reading the same secret values back during refresh.

```
=== NAME  TestAccFastlyServiceCDNAuto_withBackend
    service_cdn_auto_test.go:47: Step 2/2 error: After applying this test step, the non-refresh plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
          # fastly_service_cdn_auto.test will be updated in-place
          ~ resource "fastly_service_cdn_auto" "test" {
              ~ active_version  = 2 -> (known after apply)
                id              = "AYbJz5kDIHP4Z8j9zIWelr"
              ~ managed_version = 2 -> (known after apply)
                name            = "tf-test-qe6xs8rv2u"
                # (3 unchanged attributes hidden)
        
              ~ backend {
                  + keepalive_time        = (known after apply)
                    name                  = "backend-etjh6mgpph"
                  ~ ssl_client_secrets    = {
                      + ssl_client_cert = (sensitive value)
                      + ssl_client_key  = (sensitive value)
                    } -> (known after apply)
                    # (25 unchanged attributes hidden)
                }
        
                # (1 unchanged block hidden)
            }
        
```

Without `Computed`, Terraform expects the nested `ssl_client_secrets` object to round-trip exactly, so the next plan wants to update it again.

So I think `Optional + Computed + Sensitive` is still needed here to avoid drift for secrets that are not reliably returned by the API.

kpfleming: Thanks, that makes sense. There is a `WriteOnly` concept for attributes in Terraform but I'm not sure if it would be applicable here.

kpfleming (approved):

PR #1325, opened 2026-07-01
Remove Framework identity from explicit versioned resources

anthony-gomez-fastly:
### Change summary

  Remove ResourceWithIdentity implementation from backend, domain, and ACL
  explicit resources. Framework identity requires that the identity uniquely
  identify one remote object, but service_id + name is ambiguous across
  service versions. 
  
   The explicit resource import contract is now service_id/version/name
  via the composite ID format, which identifies the exact versioned
  object. In-place version updates (for cloned versions that already contain the
  same backend/domain/ACL) are still supported.

  Changes:
  * refactor(acl): remove version field from ACLIdentityModel
  * refactor(acl): remove RequiresReplace modifier from version attribute
  * refactor(service): return locked status from EnsureMutableForDelete
  * fix(resources): gracefully skip deletion from locked versions
    - Update backend Delete to handle locked versions
    - Update domain Delete to handle locked versions
    - Update ACL Delete to handle locked versions
  * refactor(backend,domain,acl): remove ResourceWithIdentity interface
  * refactor(backend,domain,acl): remove identity model structs
  * refactor(backend,domain,acl): remove IdentitySchema() methods
  * refactor(backend,domain,acl): simplify ImportState() to composite ID only
  * refactor(backend,domain,acl): remove identity handling from CRUD methods

 All Submissions:

### New Feature Submissions:

Screenshot 2026-07-06 at 12 46 30 PM
Screenshot 2026-07-06 at 12 48 04 PM

Screenshot 2026-07-06 at 12 49 38 PM


internal/service/mutability.go
```diff
@@ -108,13 +108,11 @@ func (c *VersionChecker) EnsureMutableForDelete(
 	ctx context.Context,
 	serviceID string,
 	version int,
-) (bool, diag.Diagnostics) {
-	var diags diag.Diagnostics
-
+) (notFound bool, locked bool, diags diag.Diagnostics) {
```
anthony-gomez-fastly: This lets attempted deletes of resources continue gracefully without failing, since the API would catch it anyway

philippschulte (dismissed):
Thanks for working on this. I agree with the goal of making explicit ACL resources move across service versions more smoothly and avoiding manual state removal/import when a service version is cloned.

That said, I want to push back on removing `version` from the Framework identity for explicit versioned resources.

The Terraform Framework identity [docs](https://developer.hashicorp.com/terraform/plugin/framework/resources/identity) say that a resource identity must correspond to at most one remote object, and that given the identity during import, the provider must be able to determine whether that exact remote object exists and return state.

For explicit Fastly versioned resources, the actual remote lookup key is:

`service_id + version + name`

A versionless identity of:

`service_id + name`

is ambiguous because the same backend/domain/ACL name can exist across multiple service versions. Selecting active/latest in `Read()` when `version == 0` means the provider is choosing a version on behalf of the user, rather than importing the exact versioned object the user identified.

I also do not think Framework identity is required for these resources unless we have a clear product reason to support structured identity import. The Framework docs describe identity as an additional import mechanism. Providers can still support import through the normal import ID string. For explicit versioned resources, I think the correct import contract should remain:

`service_id/version/name`

That is explicit, unambiguous, and matches the resource schema, where `version` is required.

I checked the original foundation pull request again as well. The initial dual model provider pull request did not introduce Framework resource identity at all. Identity was added later. Also, when identity was first added for versioned backend/domain resources, it included the full version-scoped lookup key:

`service_id + version + name`

So I do not think we should treat versionless identity as part of the original explicit-resource design without revisiting the decision.

This is separate from whether `version` or `name` should be allowed to change in place.

For ACLs, the API does support renaming in place via:

`PUT /service/{service_id}/version/{version_id}/acl/{acl_name}`

where the path `acl_name` identifies the existing ACL and the request body can set the new `name`.

So I’m not arguing that `name` must require replacement for ACL. If we support ACL rename in place, the update path just needs to use the old/current name to locate the ACL and the planned name as the new value.

Similarly, changing only `version` in place can be useful for explicit resources, because a cloned Fastly version may already contain the same ACL and we want to avoid manual state removal/import.

But neither of those require making Framework identity versionless. Import identity still needs to identify one exact remote object, and for versioned ACLs that lookup key is:

`service_id + version + name`

A versionless identity of `service_id + name` is ambiguous across service versions. We can support in-place updates for mutable fields while still keeping import explicit and unambiguous as `service_id/version/name`.

My preference would be:

1. Do not implement `ResourceWithIdentity` for explicit versioned backend/domain/ACL resources unless we have a strong reason to support structured identity import.
2. Keep the normal import ID contract as `service_id/version/name`.
3. If we do keep Framework identity, then `version` should remain part of the identity, because otherwise the identity does not uniquely identify one remote object.
4. Support in-place updates separately where the Fastly API supports them, making sure the update path refreshes state from the exact `service_id + version + name` target after the change.

internal/acceptance_tests/acl_test.go
```diff
@@ -416,3 +417,89 @@ func TestAccFastlyServiceComputeAuto_withACLForceDestroy(t *testing.T) {
...
+				ResourceName: "fastly_service_acl.test",
+				ImportStateIdFunc: func(s *terraform.State) (string, error) {
+					return fmt.Sprintf("%s/%s/%s", serviceID, versionNumber, aclName), nil
+				},
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
```
philippschulte, line 601: Can we add acceptance coverage for the main behavior this PR is introducing: changing the explicit ACL `version` in place?

The new import tests are useful, but I don’t see a test that starts with `fastly_service_acl` pinned to one version, changes only the `version` attribute to a cloned/writable version, applies successfully, and verifies the ACL state points at the new version.

That test would also catch stale computed state issues such as preserving an `id` or `acl_id` from the previous version after the update.

internal/resources/acl/resource.go
```diff
@@ -191,10 +149,6 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
 	plan.ACLID = state.ACLID
```
philippschulte: I think this `Update()` path needs to refresh or recompute state when only `version` changes.

Since `version` no longer has `RequiresReplace()`, a version-only change can now enter this method. Right now `Update()` copies the old computed fields into the new plan:

`plan.ID = state.ID`
`plan.ACLID = state.ACLID`

The concern is the Terraform resource ID: `flatten()` builds it from `service_id + version + name`.

So changing from version 1 to version 4 could leave state with `version = 4` while `id` still reflects version 1.

Can we either recompute the ID from the planned `service_id/version/name`, or read the ACL from the target version and `flatten()` it into state?

internal/service/mutability.go
philippschulte: I think I understand the full-service-destroy motivation, but I don’t think this behavior is safe for normal explicit resource deletion.

If the version is locked, the backend/domain/ACL cannot be deleted from that version. Returning successfully here means Terraform removes the resource from state even though the remote object still exists.

That may be acceptable when the parent service is also being destroyed, because deleting the service will clean up all versions and child resources. But for a normal config change where the user removes only an explicit backend/domain/ACL and keeps the service, this silently orphans the remote object and creates drift.

So I don’t think locked-version deletes should generally “continue gracefully.” We should either keep returning a diagnostic for locked versions, or scope the graceful behavior only to the full-service-destroy path if we can reliably detect that case.

philippschulte (dismissed):
Thanks for updating this. I think the structured identity concern is addressed now.

I left a few remaining inline comments that I think we should address before merging.

internal/service/mutability.go
anthony-gomez-fastly: makes sense, reverted in `1aefd929`

internal/acceptance_tests/acl_test.go
anthony-gomez-fastly: done in `1aefd929`

internal/resources/acl/resource.go
anthony-gomez-fastly: done in `1aefd929`

anthony-gomez-fastly: i chose the latter, read the ACL from the target version

philippschulte (approved):
Thanks for the updates. This looks much better to me now.

I appreciate the added acceptance coverage for explicit ACL import and the in-place ACL version update path.

Thank you!

PR #1327, opened 2026-07-02
[CDTOOL-1576] Add fastly_service_acl_entries resource for managing CDN ACL entries

anthony-gomez-fastly:
### Change summary

Implements a new resource to manage ACL entries with support for both explicit
  version management and bulk operations. This resource provides fine-grained
  control over ACL entries separate from the ACL definition itself.

  Features:
  - Bulk ACL entry management with create, update, and delete operations
  - manage_entries flag to control whether entries should be reconciled
  - Support for IP, subnet, negated, and comment fields per entry
  - Import support using service_id/acl_id format
  - Plan modifiers for service_id and acl_id (RequiresReplace)
  - Comprehensive acceptance tests covering 250+ entry scenarios

 All Submissions:

### New Feature Submissions:

* [ ] Does your submission pass tests?
* [ ] Post the output of your test runs
Screenshot 2026-07-08 at 10 19 04 AM
Screenshot 2026-07-08 at 10 46 52 AM


examples/cdn-acl-entries/README.md
```diff
@@ -0,0 +1,36 @@
+# ACL Entries Example
+
+This example demonstrates managing ACL entries for a Fastly service using the `fastly_service_acl_entries` resource.
```
philippschulte: nit: this README refers to `fastly_service_acl_entries`, but the new resource implemented in this PR is `fastly_service_cdn_acl_entries`.

internal/resources/cdnaclentries/list.go
```diff
@@ -0,0 +1,157 @@
...
+		stream.Results = list.ListResultsStreamDiagnostics(diag.Diagnostics{
+			diag.NewErrorDiagnostic("Error listing Fastly services", err.Error()),
+		})
+		return
+	}
+
+	stream.Results = func(push func(list.ListResult) bool) {
+		var count int64
+		for _, svc := range services {
+			if svc == nil || svc.Type == nil || !service.TypeSupported(*svc.Type, service.TypeVCL) {
```
philippschulte, line 66: The current SDK resource is generic `fastly_service_acl_entries` and manages entries by `service_id + acl_id`. It does not appear to restrict entries to CDN services only.

internal/resources/cdnaclentries/list.go
```diff
@@ -0,0 +1,157 @@
...
+				count++
+
+				aclID := fastly.ToValue(a.ACLID)
+
+				result := req.NewListResult(ctx)
+				result.DisplayName = service.ToGeneratedResourceName(fastly.ToValue(svc.Name), serviceID, *a.Name)
+
+				result.Diagnostics.Append(result.Identity.SetAttribute(ctx, path.Root("service_id"), serviceID)...)
+				result.Diagnostics.Append(result.Identity.SetAttribute(ctx, path.Root("version"), int64(version))...)
+				result.Diagnostics.Append(result.Identity.SetAttribute(ctx, path.Root("acl_id"), aclID)...)
```
philippschulte: I think this list resource is still using structured identity, even though we decided not to support structured imports for these resources.

`fastly_service_cdn_acl_entries` imports through the normal string import ID:

`service_id/acl_id`

But this list resource sets structured identity attributes:

`service_id`
`version`
`acl_id`

`version` is also not part of the resource schema or import contract.

Since we intentionally removed `ResourceWithIdentity` support, can we avoid setting structured identity here and keep the list/query behavior aligned with the normal import ID contract? Otherwise I’m worried generated imports from the list resource may not work with the actual resource importer.

internal/resources/cdnaclentries/schema.go
```diff
@@ -0,0 +1,104 @@
...
+					"negated": schema.BoolAttribute{
+						Optional:    true,
+						Computed:    true,
+						Description: "A boolean that will negate the match if true.",
+					},
+					"comment": schema.StringAttribute{
+						Optional:    true,
+						Computed:    true,
+						Description: "A personal freeform descriptive note.",
+					},
```
philippschulte, line 100: Can we add a custom schema validator to reject duplicate entries with the same `ip + subnet`?

The update logic correctly treats `ip + subnet` as the Fastly ACL entry identity, because comment-only or negated-only changes should update the existing entry. But the schema can still allow two Terraform `entry` blocks with the same `ip + subnet` if another field differs, such as `comment` or `negated`.

I think this would be best handled as a custom validator on the `entry` collection, rather than inside `Create()` / `Update()`, so users get a plan-time diagnostic and we can unit test the validation directly.

philippschulte (dismissed):
Thanks for adding this. The overall shape looks good to me.

I left a few comments that I think we should address before merging.

examples/cdn-acl-entries/README.md
anthony-gomez-fastly: fixed in `2b007e91`

internal/resources/cdnaclentries/schema.go
anthony-gomez-fastly: addressed in `e1df3150`

internal/resources/cdnaclentries/list.go
anthony-gomez-fastly: addressed in `e1df3150`

internal/resources/cdnaclentries/list.go
anthony-gomez-fastly: addressed in `e1df3150`

anthony-gomez-fastly:
Screenshot 2026-07-08 at 12 04 52 PM
Screenshot 2026-07-08 at 12 06 41 PM

docs/resources/service_cdn_acl_entries.md
```diff
@@ -0,0 +1,128 @@
...
+}
+
+resource "fastly_service_cdn_acl_entries" "example" {
+  service_id     = fastly_service_cdn.example.id
+  acl_id         = fastly_service_cdn_acl.example.acl_id
+  manage_entries = true
+
+  entry {
+    ip      = "192.0.2.1"
+    subnet  = "32"
```
philippschulte: nit: `subnet` is defined as a numeric attribute, but the docs and examples use quoted strings which is the same in the current provider docs!

Fastly [docs](https://www.fastly.com/documentation/reference/api/acls/acl-entry/):

```
subnet	integer	Number of bits for the subnet mask applied to the IP address. For IPv4 addresses, a value of 32 represents the smallest subnet mask (1 address), 24 represents a class C subnet mask (256 addresses), 16 represents a class B subnet mask (65k addresses), and 8 is class A subnet mask (16m addresses). If not provided, no mask is applied.
```

Can we update these examples to use numeric values instead? That matches the schema and the acceptance-test configs.

philippschulte (dismissed):
Thanks for the updates. This looks great now.

I left one small docs/example nit around using numeric values for `subnet` instead of quoted strings. No blocking comments from me.

docs/resources/service_cdn_acl_entries.md
anthony-gomez-fastly: done in `5ec19e90`

philippschulte (approved):
The broader removal of `ResourceWithIdentity` from the top-level service resources makes sense to me for consistency.

Thank you!

PR #1334, opened 2026-07-08
[CDTOOL-1547] Add support for ACLs in Compute

anthony-gomez-fastly:
### Change summary

Ports the SDKv2 ACL resources to the Plugin Framework as versionless, account-level resources:
- fastly_acl — standalone ACL resource (id, name)
- fastly_acls — data source listing all account Compute ACLs
- Unit/acceptance tests, generated Registry docs, and a runnable example
- Extends the Compute lifecycle test script to cover ACL create, entries update, and destroy alongside existing service/version/action checks

 All Submissions:

### New Feature Submissions:

Screenshot 2026-07-09 at 2 55 03 PM
Screenshot 2026-07-09 at 2 57 20 PM

### DIFF
--- a/docs/data-sources/acls.md
+++ b/docs/data-sources/acls.md
@@ -1,0 +1,43 @@
+---
+page_title: "fastly_acls Data Source - fastly"
+subcategory: ""
+description: |-
+  Use this data source to retrieve a list of Fastly ACLs.
+---
+
+# fastly_acls (Data Source)
+
+Use this data source to retrieve a list of [Fastly ACLs](https://www.fastly.com/documentation/reference/api/compute-acls/).
+
+## Example Usage
+
+```terraform
+data "fastly_acls" "example" {}
+
+output "fastly_acls_all" {
+  value = data.fastly_acls.example.acls
+}
+
+output "fastly_acls_filtered" {
+  # Example: get the ID of the ACL named "my_acl"
+  value = one([
+    for acl in data.fastly_acls.example.acls :
+    acl.id if acl.name == "my_acl"
+  ])
+}
+```
+
+## Schema
+
+### Read-Only
+
+- `acls` (Attributes Set) List of all ACLs. (see [below for nested schema](#nestedatt--acls))
+- `id` (String) Terraform data source identifier.
+
+<a id="nestedatt--acls"></a>
+### Nested Schema for `acls`
+
+Read-Only:
+
+- `id` (String) Identifier of the ACL.
+- `name` (String) Name of the ACL.
--- a/docs/resources/acl.md
+++ b/docs/resources/acl.md
@@ -1,0 +1,38 @@
+---
+page_title: "fastly_acl Resource - fastly"
+subcategory: ""
+description: |-
+  Provides an Access Control List (ACL) that defines CIDR-based access rules and is accessible to Compute services during request processing.
+---
+
+# fastly_acl (Resource)
+
+Provides an Access Control List (ACL) that defines CIDR-based access rules (e.g., allow/block IP ranges) and is accessible to Compute services during request processing.
+
+This resource is versionless: it is not tied to a service version and is managed independently of any `fastly_service_compute` resource.
+
+## Example Usage
+
+```terraform
+resource "fastly_acl" "example" {
+  name = "my_acl"
+}
+```
+
+## Schema
+
+### Required
+
+- `name` (String) A unique name to identify the ACL. Changing this attribute will delete and recreate the ACL, discarding its current entries. Any `resource_link` referencing this ACL must be removed first.
+
+### Read-Only
+
+- `id` (String) The ID of this ACL.
+
+## Import
+
+Fastly ACLs can be imported using their ACL ID, e.g.
+
+```shell
+terraform import fastly_acl.example <acl_id>
+```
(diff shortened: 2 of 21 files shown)


internal/resources/cdnaclentries/resource.go
```diff
@@ -312,9 +312,19 @@ func (r *Resource) ImportState(ctx context.Context, req resource.ImportStateRequ
...
+	}
+	entries := flattenEntries(ctx, remote, types.SetNull(types.ObjectType{AttrTypes: entryAttrTypes}), &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_id"), serviceID)...)
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("acl_id"), aclID)...)
-	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("manage_entries"), true)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("entry"), entries)...)
```
anthony-gomez-fastly: I put this in here but am happy to split it out into it's own PR. Previously importing set manage_entries to true, which could potentially blow away un imported entries. LMK what you think

rcaril: This is a good catch, but perhaps to your point it may be better to wrap with the full ACL entries implementation?

internal/acceptance_tests/acls_test.go
```diff
@@ -0,0 +1,83 @@
+package acceptancetests
```
rcaril: Is the purpose of this additional file to test the interaction of HCL with multiple ACLs? If so, did we need a separate file for this?

anthony-gomez-fastly: good call, consolidating

anthony-gomez-fastly: you could argue that having the data source test be separate is reasonable but since it's small it is ok to pop it into the same file

internal/resources/cdnaclentries/resource.go
anthony-gomez-fastly: sure, i'll remove it from this and put it in with the entries

docs/resources/acl.md
```diff
@@ -0,0 +1,38 @@
+---
+page_title: "fastly_acl Resource - fastly"
```
philippschulte, line 2: Can we keep the Terraform resource/data source names aligned with the current provider’s Compute ACL naming, and also make sure the related package, folder, file, docs, and template names include compute as well?

docs/data-sources/acls.md
```diff
@@ -0,0 +1,43 @@
+---
+page_title: "fastly_acls Data Source - fastly"
+subcategory: ""
+description: |-
+  Use this data source to retrieve a list of Fastly ACLs.
+---
+
+# fastly_acls (Data Source)
```
philippschulte, line 8: I think we should keep `compute` in the naming throughout this implementation.

Fastly supports both service/VCL ACLs and Compute ACLs, and this PR is specifically for the account-level Compute ACL API. Using generic names like `fastly_acl` / `fastly_acls` is ambiguous next to the existing service ACL resources.

Can we keep the Terraform resource and data source names aligned with the current provider’s `fastly_compute_acl` naming, and also make sure the related package, folder, file, docs, and template names include `compute` as well?

internal/acceptance_tests/service_resource_link_test.go
```diff
@@ -0,0 +1,144 @@
...
+func TestAccFastlyServiceResourceLink_ACL(t *testing.T) {
+	t.Parallel()
+	serviceName := fmt.Sprintf("tf-test-%s", acctest.RandString(10))
+	aclName := fmt.Sprintf("tf_test_acl_%s", acctest.RandString(10))
+	linkName := fmt.Sprintf("tf_test_link_%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { PreCheck(t) },
+		ProtoV6ProviderFactories: ProtoV6ProviderFactories(),
+		CheckDestroy:             CheckServiceDestroy("fastly_service_compute"),
```
philippschulte: Non-blocking test cleanup suggestion: these explicit resource-link tests create an account-level Compute ACL, but `CheckDestroy` only verifies the Compute service cleanup.

At the end of the test, Terraform should destroy the `fastly_service_resource_link`, the `fastly_acl`, and the `fastly_service_compute` resources. Since the Compute ACL is account-level and not owned by the service, deleting the service does not prove the ACL was deleted.

The current SDKv2 test has the same limitation, so I don’t think this needs to block the PR. But since this PR already adds `CheckServiceAndACLDestroy(...)` for the `_auto` tests, it may be worth using the same helper here too so leaked Compute ACLs are caught consistently.

philippschulte (dismissed):
Thanks for adding this. The implementation looks good overall, and I like that the PR includes resource-link test coverage for Compute ACLs as part of the lifecycle.

I left one main naming comment: since Fastly supports both service ACLs and Compute ACLs, I think the resource/data source names, docs, packages, folders, and file names should include `compute` consistently to avoid ambiguity.

The only other note is a non-blocking test cleanup suggestion around verifying Compute ACL cleanup in the explicit resource-link tests.

docs/resources/acl.md
anthony-gomez-fastly: This was intentional, @kpfleming mentioned that there are reasonably imminent plans to make this type of ACL available to all services and recommended we have the name be generic ahead of this coming change

internal/acceptance_tests/service_resource_link_test.go
anthony-gomez-fastly: good point, fixed in `89a9798f`

docs/resources/acl.md
philippschulte: Got it! Thanks for the explanation.

philippschulte (approved):
Thank you!

PR #1335, opened 2026-07-08
Add resource_link support for explicit and auto Compute services

anthony-gomez-fastly:
### Change summary

Port the SDKv2 resource_link block to the Plugin Framework so shared resources (KV Stores, Config Stores, etc.) can be attached to a Compute service version and read from Wasm code via resource_id.

- Add internal/resources/resourcelink, wrapping Fastly's generic resource-link API (Create/Get/Update/Delete/ListResources) with the same NestedModel/reconcile pattern used by backend and domain.
- Wire a resource_link nested block into fastly_service_compute_auto, participating in its clone/validate/activate lifecycle.
- Add fastly_service_resource_link as the standalone explicit-service counterpart, restricted to Compute services via TypeCompute validation.
- Add unit tests and acceptance tests covering create, rename, retarget, remove, and import.
- Add examples and generated Registry docs for both variants.

Refs: CDTOOL-1538

 All Submissions:

### New Feature Submissions:

Screenshot 2026-07-08 at 5 48 57 PM
Screenshot 2026-07-08 at 6 00 16 PM


internal/resources/resourcelink/schema.go
```diff
@@ -0,0 +1,182 @@
...
+		},
+		"resource_id": schema.StringAttribute{
+			Required:    true,
+			Description: "The ID of the shared resource to link (e.g. the ID of a KV Store or Config Store).",
+		},
+		"link_id": schema.StringAttribute{
+			Computed:    true,
+			Description: "An alphanumeric string identifying this resource link.",
+		},
+	}
```
philippschulte, line 43: Can we add schema-level validation on the `resource_link` collection to reject duplicate keys?

I think this should live on the nested block collection rather than on the individual `name` or `resource_id` attributes, because the validation needs to compare all configured blocks.

At minimum, I think we should reject duplicate `resource_id` values because the reconciliation logic keys resource links by `resource_id`.

I also think we should reject duplicate `name` values, since `name` is the alias Compute code uses to open the linked resource. Two links with the same alias would be ambiguous even if they point at different underlying resources.

This seems similar to the custom collection-level uniqueness validator we added for ACL entries.

internal/resources/resourcelink/resource.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+}
+
+func (r *Resource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	serviceID, version, name, err := importutil.ParseCompositeID(req.ID)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Invalid Import ID",
+			"Expected import ID in format: service_id/version/name\n"+
+				"For example: service123/3/my-acl\n\n"+
```
philippschulte: nit: this example still says `my-acl`.

Since this is the resource link importer, can we update the example to use a resource link alias instead?

For example:

`service123/3/my-store-link`

philippschulte (dismissed):
Thanks for adding this. The implementation looks good overall.

I left a couple of smaller comments.

internal/resources/resourcelink/schema.go
anthony-gomez-fastly: added in `df7c757e`

internal/resources/resourcelink/resource.go
anthony-gomez-fastly: fixed in `df7c757e`

philippschulte (dismissed):
Sorry I found one more nit for the standalone explicit resource: `fastly_service_resource_link` supports direct import, but the generated docs do not show the import format.

Can we add an Import section to `docs/resources/service_resource_link.md`, for example:

`terraform import fastly_service_resource_link.example SERVICE_ID/VERSION/NAME`

philippschulte (approved):
Thanks for the updates. This looks good to me now.

PR #1341, opened 2026-07-10
[CDTOOL-1546] Add support for compute ACL entries

anthony-gomez-fastly:
### Change summary

Port the SDKv2 Compute ACL entries resource to the Plugin Framework as a versionless, account-level resource for batch-managing CIDR prefix -> ALLOW/BLOCK entries on an existing Compute ACL.

- fastly_acl_entries: manage_entries drift-suppression flag, CIDR/action validation, batch update/delete against the API
- unit and acceptance tests, generated Registry docs, and a runnable example
- extend the Compute lifecycle test script to cover ACL entries create, update, and destroy against an ACL and attached to a service via fastly_service_resource_link

 

 All Submissions:

### New Feature Submissions:

Screenshot 2026-07-10 at 2 04 09 PM
Screenshot 2026-07-10 at 2 06 15 PM
Screenshot 2026-07-10 at 2 07 47 PM

### DIFF
--- a/docs/resources/acl_entries.md
+++ b/docs/resources/acl_entries.md
@@ -1,0 +1,76 @@
+---
+page_title: "fastly_acl_entries Resource - fastly"
+subcategory: ""
+description: |-
+  Manages CIDR-based allow/block entries within a Fastly ACL.
+---
+
+# fastly_acl_entries (Resource)
+
+Manages CIDR-based allow/block entries within a Fastly ACL.
+
+By default, Terraform does not continue to manage the entries after the initial `terraform apply`. This allows you to make changes to ACL entries outside of Terraform using the [Fastly API](https://developer.fastly.com/reference/api/) or [Fastly CLI](https://developer.fastly.com/learning/tools/cli/) without Terraform resetting them.
+
+To have Terraform continue managing the entries after creation (e.g., deleting any entries not defined in the config), set `manage_entries = true`.
+
+~> **Note:** Use `manage_entries = true` cautiously. Terraform will overwrite external changes and delete any unmanaged entries.
+
+## Example Usage
+
+Basic usage (with seeded values, unmanaged after initial apply):
+
+```terraform
+resource "fastly_acl" "example" {
+  name = "my_acl"
+}
+
+resource "fastly_acl_entries" "example" {
+  acl_id = fastly_acl.example.id
+  entries = {
+    "192.0.2.0/24"    = "ALLOW"
+    "198.51.100.0/24" = "BLOCK"
+  }
+}
+```
+
+Terraform-managed usage (where Terraform controls entries long-term):
+
+```terraform
+resource "fastly_acl" "example" {
+  name = "my_acl"
+}
+
+resource "fastly_acl_entries" "example" {
+  acl_id = fastly_acl.example.id
+  entries = {
+    "203.0.113.0/24"  = "BLOCK"
+    "198.51.100.0/24" = "ALLOW"
+  }
+  manage_entries = true
+}
+```
+
+## Schema
+
+### Required
+
+- `acl_id` (String) The ID of the ACL that the entries belong to.
+- `entries` (Map of String) A map representing the entries in the ACL, where the keys are CIDR prefixes and the values are actions (`ALLOW` or `BLOCK`).
+
+### Optional
+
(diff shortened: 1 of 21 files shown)


docs/resources/acl_entries.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Import
+
+Fastly ACL entries can be imported using the format `<acl_id>/entries`, e.g.
+
+```shell
+terraform import fastly_acl_entries.example <acl_id>/entries
+```
+
+When imported, `manage_entries` is automatically set to `true`.
```
philippschulte: The Compute ACL entries importer does not appear to set it to 'true'.

internal/resources/cdnaclentries/resource.go
```diff
@@ -312,9 +312,19 @@ func (r *Resource) ImportState(ctx context.Context, req resource.ImportStateRequ
...
+	}
+	entries := flattenEntries(ctx, remote, types.SetNull(types.ObjectType{AttrTypes: entryAttrTypes}), &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_id"), serviceID)...)
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("acl_id"), aclID)...)
-	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("manage_entries"), true)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("entry"), entries)...)
```
philippschulte: I think we should make the import behavior consistent between service/VCL ACL entries and Compute ACL entries.

The current service/VCL ACL entries importer sets `manage_entries = true` on import, while the current Compute ACL entries importer does not. That seems like a mismatch in the SDKv2 provider rather than behavior we should preserve in the dual-model implementation.

For an imported ACL entries resource, Terraform is adopting the existing remote entries into state, so I think it is more intuitive for imported entries to become managed by default. That also matches the existing service/VCL ACL entries behavior and the docs in this PR.

Can we set `manage_entries = true` during Compute ACL entries import as well, and avoid changing the CDN/service ACL entries importer away from that behavior?

philippschulte (dismissed):
Thanks for adding this. The core implementation looks good to me.

I left one main comment around import behavior. I think imported ACL entries should consistently set `manage_entries = true` for both service/VCL and Compute ACL entries. The current SDKv2 difference looks like a mismatch rather than something we should carry forward.

docs/resources/acl_entries.md
anthony-gomez-fastly: missed that, fixing.

internal/resources/cdnaclentries/resource.go
anthony-gomez-fastly: I agree that it's more intuitive, the problem i was hoping to avoid was where people have entries that are outside of terraform that will be blown away. I will set managed entries to `true` and add documentation to warn users that when importing, ACL entries will become managed for that ACL and non terraformed entries will be deleted

anthony-gomez-fastly: done in `24d5b1e1`

internal/acceptance_tests/acl_entries_test.go
```diff
@@ -0,0 +1,218 @@
...
+						"192.0.2.0/24":    "ALLOW",
+						"198.51.100.0/24": "BLOCK",
+					}),
+				),
+			},
+			{
+				ResourceName:            "fastly_acl_entries.acl_entries",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"manage_entries"},
```
philippschulte: Since import now explicitly sets `manage_entries = true`, can we remove `manage_entries` from `ImportStateVerifyIgnore`?

philippschulte (dismissed):
Thanks for the updates. This looks good to me overall.

The import behavior is now consistent across ACL entries resources, and I’m good with the `manage_entries = false` refresh behavior since it matches the dual-model service/CDN ACL entries implementation.

I left one small inline test comment around removing `manage_entries` from `ImportStateVerifyIgnore`, but I don’t have any blocking concerns beyond that.

internal/acceptance_tests/acl_entries_test.go
anthony-gomez-fastly: fixed in `d9c1924c`

philippschulte (approved):
Thanks for the update. This looks good to me now.

PR #1345, opened 2026-07-14
Expose certificate_id on tls_subscription_validation to fix empty cert ID in same-apply activations (#1197)

### LINKED ISSUE #1197: fastly_tls_activation fails with 400 when using certificate_id from fastly_tls_subscription in same apply
# fastly_tls_activation fails with 400 when using certificate_id from fastly_tls_subscription in same apply

## Description

When using `fastly_tls_subscription` and then passing its `certificate_id` into `fastly_tls_activation` in the same Terraform apply, the activation is created with an empty `certificate_id`, and the Fastly API returns:

```
400 - Bad Request:
  Title:  Missing required attribute
  Detail: param is missing or the value is empty: tls_certificate_id
```

## Minimal example (anonymous)

```hcl
# Certificates map: key = domain (common_name), value = subscription config
variable "certificates" {
  type = map(object({
    authority     = optional(string, "lets-encrypt")
    common_name   = optional(string, "")
    domains       = set(string)
    force_destroy = optional(bool, true)
  }))
}

data "fastly_tls_configuration" "tls_1_3" {
  name = "HTTP/3 & TLS v1.3"
}

resource "fastly_tls_subscription" "for" {
  for_each = var.certificates

  certificate_authority = each.value.authority
  common_name          = each.value.common_name
  domains              = each.value.domains
  force_destroy        = each.value.force_destroy
}

resource "fastly_tls_activation" "tls_1_3_activation" {
  for_each = var.certificates

  certificate_id   = fastly_tls_subscription.for[each.key].certificate_id
  domain           = each.value.common_name
  configuration_id = data.fastly_tls_configuration.tls_1_3.id

  depends_on = [fastly_tls_subscription.for]
}
```

With input like:

```hcl
certificates = {
  "alias.example.com" = {
    common_name = "alias.example.com"
    domains     = ["alias.example.com"]
  }
}
```

On first apply (new subscription), the activation fails with 400 because `certificate_id` is empty until Fastly issues the cert after DNS validation.

## Steps to reproduce

1. Create a `fastly_tls_subscription` (e.g. with `for_each` over a set of domains).
2. Create `fastly_tls_activation` with:
   - `certificate_id = fastly_tls_subscription.for[each.key].certificate_id`
   - `depends_on = [fastly_tls_subscription.for]`
3. Run `terraform apply`.

## Expected behavior

- Activation is created only when `certificate_id` is available (e.g. after subscription validation), **or**
- The provider/docs clearly require using `fastly_tls_subscription_validation` (or a second apply) before using `certificate_id` in `fastly_tls_activation`.

## Actual behavior

- The activation is created in the same apply as the subscription.
- At that time `certificate_id` is still empty (certificate is issued asynchronously by Fastly after DNS validation).
- The API is called with an empty `tls_certificate_id` and returns 400.

## Root cause

- `fastly_tls_subscription.certificate_id` is read-only and only populated by Fastly **after** the certificate is issued (asynchronously, after DNS/ACME validation).
- Terraform does not re-read resource attributes mid-apply, so the activation is planned and applied with the initial (empty) `certificate_id`.

## Environment

- Terraform: ~1.14
- Provider: fastly/fastly ~7.1

marceloboeira:
### Change summary

Fixes #1197 `fastly_tls_activation` (or anything else) consuming
`fastly_tls_subscription.certificate_id` in the same apply receives an empty
value, because certificates are issued asynchronously after DNS/ACME
validation, and fails with `400 … tls_certificate_id is missing or empty`.

Changes:

1. **`fastly_tls_subscription_validation` now exports `certificate_id`**
   (computed). The resource already blocks until the subscription is issued;
   exposing the certificate ID lets users chain issuance-dependent resources
   in a single apply — same pattern as `aws_acm_certificate_validation.certificate_arn`.
2. **Renewal churn guard:** the validation resource's Read now keys validity
   on the certificate's presence rather than `state == "issued"`, so a
   subscription in `renewing` state no longer destroys/recreates the
   validation resource (and cascades to consumers) on every renewal refresh.
3. **Fail fast:** `fastly_tls_activation` Create returns an actionable error
   when `certificate_id` is empty instead of the opaque API 400.
4. **Docs corrected:** managed subscriptions are auto-activated by Fastly on
   issuance — pairing `fastly_tls_activation` with `fastly_tls_subscription`
   fails with `400 domain_id has already been taken`. Docs now direct users to
   set `configuration_id` on the subscription and read the auto-created
   activation via the `fastly_tls_activation` data source; a complete
   multi-domain example was added.

All Submissions:

* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?

### New Feature Submissions:

`go build ./...`, `go vet ./fastly/` clean; new schema unit test passes
(`TestResourceFastlyTLSSubscriptionValidation_CertificateIDSchema`).

Acceptance-testing the full flow is not possible in CI (completing an ACME
DNS challenge requires a real, publicly resolvable domain), so it was
verified manually end-to-end against real domains:

* **Before** (released provider, issue's config): apply fails with
  `400 … param is missing or the value is empty: tls_certificate_id`.
* **After** (this branch, `certificate_id` chained through the validation
  resource): one single apply — subscription → validation blocks (~38s after
  challenge records existed) → certificate IDs populated → downstream
  data sources read the auto-created activations. Second `terraform plan`
  shows no changes.

### User Impact

* New computed attribute `fastly_tls_subscription_validation.certificate_id`; no breaking changes.
* Users previously stuck on the 400 need no state surgery: point the consumer
  at the validation resource's `certificate_id` and re-apply.
* Subscriptions in `renewing` state no longer recreate the validation resource.

None; additive change, existing configurations are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,13 @@
 
 ### ENHANCEMENTS:
 
+- feat(tls_subscription_validation): expose computed `certificate_id`, allowing issuance-dependent resources to be chained in a single apply ([#1197](https://github.com/fastly/terraform-provider-fastly/issues/1197))
+
 ### BUG FIXES:
+
+- fix(tls_subscription_validation): key resource validity on certificate presence instead of `issued` state, so subscriptions in `renewing` state no longer destroy/recreate the validation resource on refresh ([#1197](https://github.com/fastly/terraform-provider-fastly/issues/1197))
+
+- fix(tls_activation): fail fast with an actionable error when `certificate_id` is empty (certificate not yet issued) instead of an opaque API 400; document that managed subscription domains are auto-activated by Fastly and must not be paired with `fastly_tls_activation` ([#1197](https://github.com/fastly/terraform-provider-fastly/issues/1197))
 
 - fix(logging): Validate logging `format` length (max 12288 characters) at plan/validate time instead of failing at apply time ([#1340](https://github.com/fastly/terraform-provider-fastly/issues/1340))
 
--- a/docs/resources/tls_activation.md
+++ b/docs/resources/tls_activation.md
@@ -12,6 +12,8 @@ Enables TLS on a domain using a specified custom TLS certificate.
 
 ~> **Note:** The Fastly service must be provisioned _prior_ to enabling TLS on it. This can be achieved in Terraform using [`depends_on`](https://www.terraform.io/docs/configuration/meta-arguments/depends_on.html).
 
+~> **Warning:** This resource is for custom (uploaded) certificates. Do not pair it with a `fastly_tls_subscription` — Fastly automatically activates TLS on a subscription's domains once the managed certificate is issued, so creating an activation for those domains fails with `400 domain_id has already been taken` (and referencing `fastly_tls_subscription.<name>.certificate_id` in the same apply fails earlier with an empty certificate ID). For managed certificates, set `configuration_id` on the subscription and use the `fastly_tls_activation` data source to read the automatically-created activation — see the `fastly_tls_subscription_validation` documentation for a complete example.
+
 ## Example Usage
 
 Basic usage:
--- a/docs/resources/tls_subscription.md
+++ b/docs/resources/tls_subscription.md
@@ -28,6 +28,8 @@ The workflow configures a `fastly_tls_subscription` resource, then a `aws_route5
 
 We configure the `fastly_tls_subscription_validation` resource, which blocks other resources until the challenge DNS records have been validated by Fastly.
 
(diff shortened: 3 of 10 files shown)


kpfleming:
I've just done a first review of this and it looks very good. The only comment I have so far is that we should not refer to 'DNS' or 'ACME' in the text referring to validation, but should instead just refer to 'domain validation'. There are some customers using non-DNS-based validation today, and there could be others in the future, and we don't want to have to update the help text here when those changes happen :)

marceloboeira:
> * New computed attribute `fastly_tls_subscription_validation.certificate_id`; no breaking changes.
> * Users previously stuck on the 400 need no state surgery: point the consumer
>   at the validation resource's `certificate_id` and re-apply.
> * Subscriptions in `renewing` state no longer recreate the validation resource.

@kpfleming  Thanks for the quick review, and good point! I've updated the wording in the docs.

kpfleming:
I've resolved the merge conflict in CHANGELOG.md. Please note that the issue link in CHANGELOG entries should refer to the Pull Request which made the changes, not the issue that the PR is resolving.

kpfleming (approved):

PR #1346, opened 2026-07-14
Render in-place plan diffs for snippet blocks (#802)

### LINKED ISSUE #802: fastly_service_vcl Overly verbose Snippets diff
### Terraform Version

```shell
Terraform v1.6.3
on darwin_arm64
```

- `fastly_service_vcl`

```hcl
resource "fastly_service_vcl" "this" {
  name     = "some-service"
  activate = true

  snippet {
    name     = "my_snippet"
    type     = "init"
    content  =  <<EOF
      table my_table {
        "foo" = "bar"
        "bar" = "foo"
      }
      # Old line
    EOF
  }
}

```

### Expected Behavior

```diff
...
        snippet {
          name     = "my_snippet"
          content  = <-EOF
            ... lots of suff ...
-           # Old Line
+           # Changed Line
          EOF
        }
```

### Actual Behavior

```diff
...
-       snippet {
-         name     = "my_snippet"
-         type     = "init"
-         content  = <-EOF
-           ... lots of suff ...
-         EOF
-       }

+       snippet {
+         name     = "my_snippet"
+         type     = "init"
+         content  = <-EOF
+           ... lots of suff ...
+           # Changed Line
+         EOF
+       }
```

### Steps to Reproduce

1. `terraform plan/apply`
1. make a change to snippet content
1. `terraform plan` — see the "wrong" diff (more verbose than necessary)

### Important Factoids (optional)

* I'm guessing other types of snippets (dynamic) and also raw VCL should have the same issue
* Is that because of the use of `Type.Set` in the resource schema? Would it be possible to use something like `Type.Map` and eventually use the name of the snippet as a map (I'm guessing if the name changes it is okay to show it as a new snippet)

marceloboeira:
### Change summary

Fixes #802.

Editing one line of a `snippet` block's `content` used to render as a remove+add of the whole block (or worse ... with multiple snippets, Terraform cross-matches set elements and shows one snippet "becoming" another). 

"Root cause" is `snippet` is a `schema.TypeSet`, and since Terraform 0.12 set element identity is pure value equality computed by core: no provider-side hook (custom `Set:` hash, `DiffSuppressFunc`, `CustomizeDiff`) can change that rendering. Notably, apply behavior was already correct: the provider matches snippets by `name` and issues in-place `UpdateSnippet` calls. The noise was purely a display artifact.

This PR converts the container to `schema.TypeList`, with two guards so that ordering (meaningless to the Fastly API: execution order is driven by `priority`) never produces spurious diffs:

1. `Read` reorders remote results to match prior state order, keyed by `name` (the list API does not guarantee stable ordering); snippets created out-of-band are appended sorted by name.
2. Apply still matches by `name`: `SetDiff` gained list-based `DiffLists`/`FilterList` (the set methods delegate to them ... all other blocks are behavior-identical), and the shared CRUD `Process` handler accepts both containers via `toResourceList`. A pure config reorder applies with **zero** snippet API calls.

State migration: set and list serialize identically (JSON array), so data is compatible as-is; `SchemaVersion` is bumped 1 → 2 with a pass-through upgrader to make the change explicit.

**Before** (v9.x: one line of `my_snippet` edited; Terraform cross-matches the two set elements, so each snippet's entire content is shown replaced by the other's):

```hcl
  # fastly_service_vcl.this will be updated in-place
  ~ resource "fastly_service_vcl" "this" {
      ~ active_version     = 2 -> (known after apply)
      ~ cloned_version     = 2 -> (known after apply)
        id                 = "<REDACTED>"
        name               = "<REDACTED>"
        # (12 unchanged attributes hidden)

      ~ snippet {
          ~ content  = <<-EOT
              - table my_table {
              -   "foo": "bar",
              -   "bar": "foo",
              - }
              - # line 1
              - # line 2
              - # line 3 -- REMOVE ME
              - # line 4
              - # line 5
            EOT
            # (1 unchanged attribute hidden)
        }

    ~ snippet {
          ~ content  = <<-EOT
              + table my_table {
              +   "foo": "bar",
              +   "bar": "foo",
              + }
              + # line 1
              + # line 2
              + # line 4
              + # line 5
            EOT
            # (1 unchanged attribute hidden)
        }

        # (1 unchanged block hidden)
    }
```

**After** (this PR, same edit: only the changed line shows):

```hcl
  # fastly_service_vcl.this will be updated in-place
  ~ resource "fastly_service_vcl" "this" {
      ~ active_version     = 2 -> (known after apply)
      ~ cloned_version     = 2 -> (known after apply)
        id                 = "<REDACTED>"
        name               = "<REDACTED>"
        # (12 unchanged attributes hidden)

      ~ snippet {
          ~ content  = <<-EOT
                table my_table {
                  "foo": "bar",
                  "bar": "foo",
                }
                # line 1
                # line 2
              - # line 3 -- REMOVE ME
                # line 4
                # line 5
            EOT
            name     = "my_snippet"
            # (2 unchanged attributes hidden)
        }

        # (3 unchanged blocks hidden)
    }
```

Scope is deliberately snippet-only; `vcl` and `dynamicsnippet` share the identical pattern and can follow in separate PRs if this approach is accepted (`toResourceList` already supports them).

All Submissions:

* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?

### Changes to Core Features:

* [x] Have you written new tests for your core changes, as applicable? (`TestSetDiff_DiffLists`, `TestReorderSnippetsByPriorState`, `TestUpgradeServiceVCLStateV1toV2_*`)
* [x] Have you successfully run tests with your changes locally? `go build ./...`, `go vet ./fastly/`, `go test ./fastly/`: 222 pass. Verified end-to-end against a real service: one-line edit renders an in-place diff (outputs above); no-op refresh plans "No changes."; config reorder applies with zero snippet API calls.

### User Impact

- Plans for content-heavy snippets become reviewable: a 1-line change shows as a 1-line diff.
- The exported `snippet` attribute changes type from set-of-objects to list-of-objects; expressions feeding it to set functions may need `tolist()`/`toset()` adjustments.

- Semver: should ship in a **major** release (attribute type change + schema version bump; state is not downgradeable afterwards).
- Release notes should call out the one-time, order-only `snippet` diff on the first plan after upgrading (state carries the old set's hash ordering); applying it makes no API changes and subsequent plans are clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ### BREAKING:
 
+- breaking(service_vcl): the `snippet` block is now a list instead of a set. Existing state upgrades automatically (schema v1 -> v2); the first plan after upgrading may show a one-time, order-only diff for `snippet` blocks — applying it makes no changes to the service. Expressions that relied on the attribute being a set (e.g. passing `fastly_service_vcl.x.snippet` to set functions) may need a `tolist()`/`toset()` adjustment ([#802](https://github.com/fastly/terraform-provider-fastly/issues/802))
+
 ### ENHANCEMENTS:
+
+- enhancement(service_vcl): render in-place attribute diffs for `snippet` blocks — a one-line `content` edit now plans as a `~ content` line diff instead of a whole-block remove/add pair ([#802](https://github.com/fastly/terraform-provider-fastly/issues/802))
 
 ### BUG FIXES:
 
--- a/docs/resources/service_vcl.md
+++ b/docs/resources/service_vcl.md
@@ -272,7 +272,7 @@ $ terraform import fastly_service_vcl.demo xxxxxxxxxxxxxxxxxxxx@2
 - `request_setting` (Block Set) (see [below for nested schema](#nestedblock--request_setting))
 - `response_object` (Block Set) (see [below for nested schema](#nestedblock--response_object))
 - `reuse` (Boolean) Services that are active cannot be destroyed. If set to `true` a service Terraform intends to destroy will instead be deactivated (allowing it to be reused by importing it into another Terraform project). If `false`, attempting to destroy an active service will cause an error. Default `false`
-- `snippet` (Block Set) (see [below for nested schema](#nestedblock--snippet))
+- `snippet` (Block List) (see [below for nested schema](#nestedblock--snippet))
 - `stage` (Boolean) Conditionally enables new service versions to be staged. If set to `true`, all changes made by an `apply` step will be staged, even if `apply` did not create a new draft version. Default `false`
 - `stale_if_error` (Boolean) Enables serving a stale object if there is an error
 - `stale_if_error_ttl` (Number) The default time-to-live (TTL) for serving the stale object for the version
--- a/fastly/block_fastly_service_snippet.go
+++ b/fastly/block_fastly_service_snippet.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -35,7 +36,9 @@ func (h *SnippetServiceAttributeHandler) Key() string {
 // GetSchema returns the resource schema.
 func (h *SnippetServiceAttributeHandler) GetSchema() *schema.Schema {
 	return &schema.Schema{
-		Type:     schema.TypeSet,
+		// TypeList (not TypeSet) so plans render in-place attribute diffs.
+		// Order carries no meaning for the API; Read keeps it stable by name.
+		Type:     schema.TypeList,
 		Optional: true,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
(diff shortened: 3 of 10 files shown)


philippschulte (commented):
Thanks for the contribution and for the detailed investigation.

Because this changes the public Terraform schema from `TypeSet` to `TypeList`, we consider it a breaking change and need to discuss the approach with the team before proceeding.

The same underlying issue also applies to the `vcl` block: it is currently modeled as a `TypeSet`, so content changes can produce the same remove/add-style plan rendering. If we adopt this approach, we should consider including the corresponding `vcl` changes in the same breaking release rather than introducing another schema-breaking migration later.

We will review the proposal and its release scope in more detail with the team and follow up soon.

marceloboeira:
@philippschulte thanks for the quick review and no worries. Also, I'm happy to help testing/validating this at scale, so far I ran a branch on a local setup with dozens of services and hundreds of VCL snippets and there has been no problems.

philippschulte:
Thank you again for the time and effort you put into this contribution. We appreciate both the detailed investigation and the work to address this long-standing plan-rendering issue.

After reviewing the pull request with the broader team, we agree with your diagnosis and overall approach. However, we have decided not to introduce this change into the current provider for timing and release-planning reasons.

The same underlying issue affects both VCL snippets and VCL configurations, and your work helped confirm the appropriate schema-level direction for addressing it. We plan to carry that approach forward into our upcoming dual-model Terraform provider rather than make this compatibility-sensitive change to the current provider so close to its transition.

Given that the new provider is already in active development and approaching its initial beta release, we believe it is better to deliver this improvement there as part of a consistent solution across both resource types.

More information about the provider rewrite and the two-resource-family model is available here:

https://community.fastly.com/t/update-two-resource-families-one-provider-your-call/4426

We currently plan to release the new provider functionality later this quarter.

We genuinely value external contributions, and your work helped validate both the importance of this issue and the technical direction for solving it. Based on the product and release strategy described above, we are going to close this pull request, with the intention of carrying its approach forward into the new provider.

Thank you again for pushing this forward. Your contribution has made a real difference to how we are approaching this area in the new provider, and we would welcome your feedback on that work once it is further along.

marceloboeira:
Thanks for the context @philippschulte, makes sense, less (state) migrations!

PR #1348, opened 2026-07-16
[CDTOOL-1555] Add fastly_kvstore resource and fastly_kvstores data source

anthony-gomez-fastly:
### Change summary

Introduces a versionless resource for managing KV Stores and a companion data source for listing them, following the standard resource package layout. Updates resource_link acceptance tests to provision their KV Store fixtures via the new fastly_kvstore resource instead of raw go-fastly API calls, removing the now-unneeded PreCheckAcc/ CreateTestKVStore helpers and the static resource_link_single.tf fixture in favor of Terraform-managed, referenceable config.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-07-16 at 4 11 25 PM


internal/datasources/idhash/idhash_test.go
```diff
@@ -1,4 +1,4 @@
-package acls
+package idhash
```
anthony-gomez-fastly, line 1: This is going to be used in a bunch of places so it made sense to generalize it

internal/acceptance_tests/fixtures/packages/valid.tar.gz
anthony-gomez-fastly, line 1: Something upstream made me have to recompile this

internal/acceptance_tests/test_helpers.go
```diff
@@ -32,22 +32,6 @@ func PreCheck(t *testing.T) {
...
-// PreCheckAcc mirrors resource.Test's own TF_ACC gate before running PreCheck. Use this
-// (instead of PreCheck) in any test that needs to perform live setup - e.g. creating a fixture
-// via CreateTestKVStore - before constructing its resource.TestCase, since that setup would
-// otherwise run unconditionally: resource.Test only checks TF_ACC after its TestCase argument,
-// including any Config strings built from that setup, has already been fully evaluated.
-//
-// TODO: once a fastly_kvstore resource exists and the resource_link tests build their KV Store
-// fixture via Terraform config instead of CreateTestKVStore, this early gate is no longer
-// needed - those tests can go back to calling PreCheck through TestCase.PreCheck like the rest.
-func PreCheckAcc(t *testing.T) {
```
anthony-gomez-fastly, line 44: removed since the original check function covers kv stores now

internal/acceptance_tests/test_helpers.go
```diff
@@ -57,34 +41,6 @@ func NewFastlyClient() (*fastly.Client, error) {
...
-// CreateTestKVStore creates a KV Store fixture directly against the Fastly API for use as a
-// resource_link target, and registers its cleanup via t.Cleanup. Returns the store's ID.
-//
-// This provider doesn't yet have a dedicated resource for KV Stores, so resource_link
-// acceptance tests need a real linkable resource created out-of-band from Terraform.
-//
-// TODO: once a fastly_kvstore resource exists, replace this raw go-fastly client call with a
-// Terraform-managed fixture (e.g. via config_builder.go) so the fixture participates in the
-// same state/plan lifecycle as the rest of the test config.
-func CreateTestKVStore(t *testing.T, name string) string {
```
anthony-gomez-fastly, line 69: removed because kv stores are now implemented

internal/acceptance_tests/kvstore_test.go
```diff
@@ -0,0 +1,218 @@
...
+	t.Parallel()
+	storeName := fmt.Sprintf("tf_test_kvstore_%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { PreCheck(t) },
+		ProtoV6ProviderFactories: ProtoV6ProviderFactories(),
+		CheckDestroy:             CheckKVStoreDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: ConfigKVStoreForceDestroy(storeName),
```
philippschulte, line 102: Can we add acceptance coverage for the non-empty deletion guard when `force_destroy = false`?

The current KV Store test covers successful cleanup with `force_destroy = true`, but not the failure path. I think we should add the same pattern used by `TestAccFastlyServiceCDNAuto_withACLForceDestroy`: create a non-empty resource with `force_destroy = false`, verify deletion/replacement fails, then enable `force_destroy` and verify it succeeds.

internal/acceptance_tests/fixtures/packages/valid.tar.gz
philippschulte: That's strange! Do the tests fail if you revert that change?

philippschulte (dismissed):
Thanks for adding this. The core KV Store resource and data source implementations look good.

I left one test-coverage comment and one question about the unrelated binary package fixture change.

internal/acceptance_tests/fixtures/packages/valid.tar.gz
anthony-gomez-fastly: yup

internal/acceptance_tests/kvstore_test.go
anthony-gomez-fastly: Added `TestAccFastlyKVStore_forceDestroyGuard`, following the `TestAccFastlyServiceCDNAuto_withACLForceDestroy` pattern: insert a key, attempt a name change (forces replacement) with `force_destroy = false` and confirm it's rejected with "KV Store is not empty", then set `force_destroy = true` and confirm the rename/replace succeeds.

anthony-gomez-fastly: fixed in `febb0632`

philippschulte (approved):
Thanks for the updates. This looks good to me now.

PR #1349, opened 2026-07-16
add flag to make test-acc to accept a keyword string

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

Screenshot 2026-07-16 at 5 08 18 PM


rcaril (dismissed):
This looks solid! 

Can you also update the `CLAUDE.md` testing section to reference the new `KEYWORD=<word>` filter for `make test-acc`?

philippschulte (commented):
Could you please clarify the intended behavior of `KEYWORD`?

The help text says it runs tests whose names contain a word, which suggests literal substring matching. However, `go test -run` interprets the value as a regular expression, so characters such as `.`, `*`, `+`, or `[` will have special behavior.

Should `KEYWORD` be treated as a literal string, or is regex matching intentional? The implementation and help text should make that explicit.

rcaril (approved):
Thank you!

PR #1350, opened 2026-07-21
[CDTOOL-1504] Add gzip nested block support to fastly_service_cdn_auto

anthony-gomez-fastly:
### Change summary

Add the internal/resources/gzip package (schema, CRUD ops, reconciler helpers) and wire it into servicecdnauto's Create/Read/Update flows so gzip configurations can be managed as a nested block. Includes unit tests, acceptance tests, HCL example blocks, and updated docs.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-07-21 at 2 35 39 PM


internal/resources/gzip/schema.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	if api.CacheCondition != nil && *api.CacheCondition != "" {
+		model.CacheCondition = types.StringValue(*api.CacheCondition)
+	} else {
+		model.CacheCondition = types.StringNull()
+	}
+	return model
+}
+
+// joinStringList converts a Terraform list of strings into the space-separated
```
rcaril, line 181: I did some testing and I do not believe it's true that `list explicitly clears the remote value rather` is accurate.  Sending `content_types= (empty string)` makes the API substitute its full default list rather then clearing out the remote value. 

It may be also worth noting that that once these fields have been set, removing them from the config does not clear the remote - `ops-Equal` skips the `Update` so the previous values persist. This seems like behavior we do not need to change, but the comment here maybe be misleading and worth adjusting.

internal/resources/gzip/schema.go
```diff
@@ -0,0 +1,250 @@
...
+	})
+}
+
+// Equal ignores content_types/extensions on the remote side when desired leaves them
+// unset, since the Fastly API silently substitutes a large default list for either
+// field when it's omitted or empty. Comparing against that default would otherwise
+// produce a permanent diff for any gzip config that doesn't set them explicitly.
+func (o ops) Equal(desired NestedModel, remote *fastly.Gzip) bool {
+	remoteModel := o.ToModel(remote)
+	if desired.ContentTypes.IsNull() || desired.ContentTypes.IsUnknown() {
```
rcaril: From local testing - 
`content_types = []` (explicit empty list) fails apply with:
```
 Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to fastly_service_cdn_auto.example, provider
│ "provider[\"registry.terraform.io/fastly/fastly\"]" produced an unexpected new value: .gzip[0].content_types:
│ was cty.ListValEmpty(cty.String), but now null.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
``` 

This _null / unknown_ check and the matching one in `ReadForVersionWithPlan` does not cover an explicit empty list, so `[]` is not preserved and gets normalized to `null`. It may be worth extending these to consider a zero length list as unset as well, matching `stringListEqual`.

rcaril (dismissed):
Looks solid! 

One issue and comment note to consider below.

internal/resources/gzip/schema.go
anthony-gomez-fastly: fixed in `6b2525e` and added test in `e190642`

internal/resources/gzip/schema.go
anthony-gomez-fastly: fixed in `6b2525e`

internal/resources/gzip/schema.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+// Equal ignores content_types/extensions on the remote side when desired leaves them
+// unset (null, unknown, or an explicit empty list), since the Fastly API silently
+// substitutes a large default list for either field when it's omitted or empty.
+// Comparing against that default would otherwise produce a permanent diff for any
+// gzip config that doesn't set them explicitly.
+func (o ops) Equal(desired NestedModel, remote *fastly.Gzip) bool {
+	remoteModel := o.ToModel(remote)
+	if listUnset(desired.ContentTypes) {
```
philippschulte, line 122: Could we handle and test the transition from a configured list to an unset or empty list?

For example, if `content_types = ["text/html"]` is applied and then removed from the same gzip block, this equality normalization replaces the non-empty remote value with the desired empty representation before comparison. The reconciler can therefore treat the objects as equal and skip `Update`, leaving `"text/html"` configured in Fastly while the subsequent read hides that remote value from Terraform state.

The current SDKv2 implementation handles this differently: when `content_types` or `extensions` is removed, it still calls `UpdateGzip` and explicitly sends an empty string so the field is reset.

Can we preserve that behavior here, or recreate the gzip configuration if the API no longer supports clearing it in place? Please add an acceptance test that:

1. Creates a gzip block with populated `content_types` and `extensions`.
2. Removes `content_types` and verifies the previous remote value is cleared.
3. Removes `extensions` and verifies that value is cleared as well.
4. Includes a final plan-only step to confirm there is no hidden or perpetual diff.

philippschulte (dismissed):
Thanks for adding `gzip` support. The implementation looks well structured.

I found one blocking issue in the equality normalization: removing previously configured `content_types` or `extensions` can be treated as no change, causing the provider to skip the API update and leave the old remote value in place.

internal/resources/gzip/schema.go
rcaril: I was mistaken when I pointed out similar behavior [here](https://github.com/fastly/terraform-provider-fastly/pull/1350#discussion_r3625519614) and said the behavior didn't need to be changed. 

From another test I did, if you do send null values to the API for `content_types` and `extensions`, it will reset to the API default - so we should in fact allow the same behavior, most likely with the `fastly.Nullable` via `go-fastly`.

philippschulte: Thanks for confirming.

With the current implementation it might be that the current equality normalization may prevent reconciliation from running when a previously configured `content_types` or `extensions` value is removed. The desired transition should result in an update.

An acceptance test covering configured values followed by removal should verify both that the API value is reset and that the following plan is clean.

anthony-gomez-fastly: addressed in `cacd3191`

rcaril (approved):
Looks good to me, thanks!

internal/resources/gzip/schema.go
philippschulte: I might be overthinking this but I believe we should extend the test to verify the value directly through `ListGzips`, rather than only checking Terraform state.

The state assertion confirms that Terraform sees the field as unset, but because the read path suppresses API values for unset fields, it does not prove that Fastly actually cleared the previous value. Could we also verify the remote gzip configuration directly through `ListGzips`?

philippschulte (changes requested):
Thanks for the adding the test.

The current state assertions confirm that Terraform sees the fields as unset, but the read path can intentionally suppress API-returned values for unset attributes. A direct API check would confirm that the previously configured `content_types` and `extensions` values were actually cleared in Fastly.

philippschulte (approved):
The concern from my previous review is addressed. No further blocking comments from me. Thank you!

PR #1352, opened 2026-07-21
add polling to ACL entries to prevent reads after from showing a diff

anthony-gomez-fastly:
Screenshot 2026-07-21 at 11 37 05 AM

there was a case where tests would fail because the batch update functions would return 202 from the API but the reads after wouldn't reflect the new changes because the operations weren't completed. Added a wait and timeout to handle those cases.

All Submissions:

### New Feature Submissions:

Screenshot 2026-07-21 at 11 40 13 AM

### DIFF
--- a/internal/resources/aclentries/resource.go
+++ b/internal/resources/aclentries/resource.go
@@ -288,7 +288,8 @@ func (r *Resource) updateEntries(ctx context.Context, aclID string, batch []*com
 // waitForEntries polls listEntries until the remote state reflects every
 // operation in batch, since the batch update endpoint applies asynchronously.
 func (r *Resource) waitForEntries(ctx context.Context, aclID string, batch []*computeacls.BatchComputeACLEntry) ([]computeacls.ComputeACLEntry, error) {
-	deadline := time.Now().Add(entriesPollTimeout)
+	ctx, cancel := context.WithTimeout(ctx, entriesPollTimeout)
+	defer cancel()
 
 	for {
 		remote, err := r.listEntries(ctx, aclID)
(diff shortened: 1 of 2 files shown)


internal/resources/aclentries/resource.go
```diff
@@ -275,3 +284,55 @@ func (r *Resource) updateEntries(ctx context.Context, aclID string, batch []*com
 		Entries:      batch,
 	})
 }
+
+// waitForEntries polls listEntries until the remote state reflects every
+// operation in batch, since the batch update endpoint applies asynchronously.
+func (r *Resource) waitForEntries(ctx context.Context, aclID string, batch []*computeacls.BatchComputeACLEntry) ([]computeacls.ComputeACLEntry, error) {
+	deadline := time.Now().Add(entriesPollTimeout)
```
philippschulte: Could we enforce the timeout through a derived context and pass that context to `listEntries`?

A slow or blocked listEntries request can exceed 30 seconds, and a converged response arriving after the deadline would still be accepted because convergence is checked before the deadline.

Using `context.WithTimeout(ctx, entriesPollTimeout)` for both the list calls and the polling wait would make the timeout apply to the complete operation.

internal/resources/aclentries/resource.go
```diff
@@ -275,3 +284,55 @@ func (r *Resource) updateEntries(ctx context.Context, aclID string, batch []*com
...
+			}
+			continue
+		}
+		if !exists || action != *op.Action {
+			return false
+		}
+	}
+
+	return true
+}
```
philippschulte, line 335: Can we add unit coverage for `entriesConverged`?

It would be useful to cover create, update, delete, mixed operations, and unrelated remote entries so that the conditions which stop polling are explicitly verified.

It would be great to have a table-driven unit test for this function.

philippschulte (dismissed):
Thanks for addressing the eventual-consistency issue. The polling approach looks reasonable.

I left one blocking comment about making the timeout cover the API requests themselves, since the current deadline can still be exceeded by a slow `listEntries` call. I also suggested adding focused unit coverage for the convergence logic.

internal/resources/aclentries/resource.go
anthony-gomez-fastly: addressed in `e8ed1cd4`

internal/resources/aclentries/resource.go
anthony-gomez-fastly: addressed in `e8ed1cd4`

philippschulte (approved):
Thanks for the update. No further blocking comments from me. This looks great!

PR #1355, opened 2026-07-22
[CDTOOL-1507] Port Image Optimizer default settings for the automatic-lifecycle service family

anthony-gomez-fastly:
### Change summary
- Add internal/resources/imageoptimizerdefaultsettings package: a singleton nested block (schema/expand/flatten)
- Wire the image_optimizer_default_settings block into fastly_service_cdn_auto (Model, Schema, Create/Read/Update reconciliation).
- Create and update are treated as the same full-replace operation, since Image  Optimizer default settings always exist once the product is enabled; removing  the block resets settings back to API defaults, but only when it was previously configured, to avoid surfacing phantom drift for services that never declared this block.
- Add unit tests for the new package, and acceptance tests covering create, update, reset-on-removal, and import

 All Submissions:

### New Feature Submissions:

Screenshot 2026-07-27 at 2 41 57 PM

### DIFF
--- a/examples/orchestration-cdn-auto/main.tf
+++ b/examples/orchestration-cdn-auto/main.tf
@@ -25,8 +25,8 @@ locals {
 }
 
 resource "fastly_service_cdn_auto" "service_1" {
-  name     = var.service_1_name
-  comment  = "Managed by Terraform"
+  name    = var.service_1_name
+  comment = "Managed by Terraform"
   domain {
     name = "www.service1.example.com"
   }
@@ -50,11 +50,24 @@ resource "fastly_service_cdn_auto" "service_1" {
     content_types = ["text/html", "text/css", "application/javascript"]
     extensions    = ["css", "js", "html"]
   }
+
+  # Requires the Image Optimizer product to already be enabled on this service
+  # (via the Fastly UI, API, or product enablement tooling). Remove this block
+  # to reset Image Optimizer default settings back to their API defaults.
+  image_optimizer_default_settings {
+    resize_filter = "lanczos3"
+    webp          = false
+    webp_quality  = 85
+    jpeg_type     = "auto"
+    jpeg_quality  = 85
+    upscale       = false
+    allow_video   = false
+  }
 }
 
 resource "fastly_service_cdn_auto" "service_2" {
-  name     = var.service_2_name
-  comment  = "Managed by Terraform"
+  name    = var.service_2_name
+  comment = "Managed by Terraform"
   domain {
     name = "www.service2.example.com"
   }
--- a/examples/orchestration-cdn-auto/main.tf
+++ b/examples/orchestration-cdn-auto/main.tf
@@ -50,19 +50,17 @@ resource "fastly_service_cdn_auto" "service_1" {
     content_types = ["text/html", "text/css", "application/javascript"]
     extensions    = ["css", "js", "html"]
   }
+}
 
-  # Requires the Image Optimizer product to already be enabled on this service
-  # (via the Fastly UI, API, or product enablement tooling). Remove this block
-  # to reset Image Optimizer default settings back to their API defaults.
-  image_optimizer_default_settings {
-    resize_filter = "lanczos3"
-    webp          = false
-    webp_quality  = 85
-    jpeg_type     = "auto"
-    jpeg_quality  = 85
-    upscale       = false
-    allow_video   = false
-  }
+# Image Optimizer must be enabled on service_1 before an
+# image_optimizer_default_settings block can be added to fastly_service_cdn_auto.service_1.
+# This resource can be applied together with the service's own creation, since it only
+# depends on the service's id. But image_optimizer_default_settings is reconciled inside
+# the service resource's own create step, which runs before this resource - so the settings
+# block itself must be added in a later apply, once this resource has enabled the product.
+# See README.md's "Configure Image Optimizer default settings" section.
+resource "fastly_service_product_image_optimizer" "service_1" {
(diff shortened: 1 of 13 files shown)


internal/acceptance_tests/test_helpers.go
```diff
@@ -43,6 +43,34 @@ func NewFastlyClient() (*fastly.Client, error) {
...
+// CreateTestKVStore creates a KV Store fixture directly against the Fastly API for use as a
+// resource_link target, and registers its cleanup via t.Cleanup. Returns the store's ID.
+//
+// This provider doesn't yet have a dedicated resource for KV Stores, so resource_link
+// acceptance tests need a real linkable resource created out-of-band from Terraform.
+//
+// TODO: once a fastly_kvstore resource exists, replace this raw go-fastly client call with a
+// Terraform-managed fixture (e.g. via config_builder.go) so the fixture participates in the
+// same state/plan lifecycle as the rest of the test config.
+func CreateTestKVStore(t *testing.T, name string) string {
```
rcaril: Was this meant to be in another PR?

rcaril (commented):
Looks solid! Just one question about the added helper.

internal/acceptance_tests/image_optimizer_default_settings_test.go
```diff
@@ -0,0 +1,86 @@
...
+				),
+			},
+			{
+				// Removing the block resets Image Optimizer default settings back to API defaults.
+				Config: ConfigCDNAutoImageOptimizerEnabled(serviceName, domainName),
+				Check: resource.ComposeTestCheckFunc(
+					CheckServiceExists("fastly_service_cdn_auto.test"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "image_optimizer_default_settings.#", "0"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "active_version", "4"),
+					resource.TestCheckResourceAttr("fastly_service_cdn_auto.test", "managed_version", "4"),
```
philippschulte, line 81: Could we add a direct `GetImageOptimizerDefaultSettings` assertion here?

The test verifies only that the block is absent from Terraform state and that a new version was activated. Since the read path does not surface remote settings once the block is removed, this can pass even if the previous non-default values remain in Fastly.

Please verify that all remote fields match the expected API defaults after removal.

examples/orchestration-cdn-auto/main.tf
```diff
@@ -50,11 +50,24 @@ resource "fastly_service_cdn_auto" "service_1" {
...
+  image_optimizer_default_settings {
+    resize_filter = "lanczos3"
+    webp          = false
+    webp_quality  = 85
+    jpeg_type     = "auto"
+    jpeg_quality  = 85
+    upscale       = false
+    allow_video   = false
+  }
 }
```
philippschulte, line 64: Can we remove this block from the bootstrap example or restructure the example as an explicit staged workflow?

This resource creates a new service, but the block requires Image Optimizer to already be enabled and the docs correctly note that a dependent `product-enablement` resource cannot satisfy that requirement during the same initial apply.

The acceptance test uses the workable sequence: create the service and enable Image Optimizer first, then add the settings block in a later apply. The example should demonstrate that same sequence rather than include the block in the initial configuration.

philippschulte (dismissed):
Thanks for adding Image Optimizer default settings support. I left two comments.

internal/acceptance_tests/test_helpers.go
anthony-gomez-fastly: ah, i made this branch before kvstore existed 😅 let me remove that

examples/orchestration-cdn-auto/main.tf
anthony-gomez-fastly: done in `7106e529`

internal/acceptance_tests/image_optimizer_default_settings_test.go
anthony-gomez-fastly: done in `7106e529`

internal/acceptance_tests/test_helpers.go
anthony-gomez-fastly: done in `7106e529`

rcaril (approved):
LGTM

internal/acceptance_tests/test_helpers.go
```diff
@@ -182,6 +183,71 @@ func CheckGzipFieldClearedRemotely(resourceName, gzipName, field, staleValue str
...
+
+		settings, err := client.GetImageOptimizerDefaultSettings(context.Background(), &fastly.GetImageOptimizerDefaultSettingsInput{
+			ServiceID:      rs.Primary.ID,
+			ServiceVersion: version,
+		})
+		if err != nil {
+			return fmt.Errorf("error fetching Image Optimizer default settings: %w", err)
+		}
+		if settings == nil {
+			return nil
```
philippschulte: I am not sure if a `nil` settings response should count as success after removal. Since Image Optimizer remains enabled in this test shouldn't we expect from the GET endpoint to return fully populated settings?

philippschulte (approved):
Thanks for the update. Both previous concerns are addressed.

I just left a question regarding `CheckImageOptimizerDefaultSettingsMatchAPIDefaults()` because I am not sure if the behavior for a `settings == nil` is what we should expect.

internal/acceptance_tests/test_helpers.go
anthony-gomez-fastly: good point, added an error instead in `3623336d`

PR #1359, opened 2026-07-23
[CDTOOL-1580] Add fastly_service_product resource blocks for each product

anthony-gomez-fastly:
### Change summary
Adds Terraform resources for all 11 Fastly Product Enablement features. Resource presence is the enablement — creating enables the product on service_id, destroying disables it (no separate enabled attribute).

New resources (internal/resources/productenablement/)
- simple.go — one generic resource.Resource implementation driven by a spec table, shared by the 8 boolean/no-config products: Fanout, Brotli Compression, Image Optimizer, Origin Inspector, Domain Inspector, WebSockets, Log Explorer & Insights, API Discovery. Replaces the old shared Model/schema/expand/flatten files.
- bot_management.go, ddos_protection.go, ngwaf.go — bespoke implementations for the 3 products with extra config (content_guard; mode; workspace_id + traffic_ramp), using UpdateConfiguration for in-place edits instead of round-tripping through Enable/Disable.
- shared.go — common id/service_id attribute builders and isEntitlementError helper used by all 11 resources.
- shared_test.go — unit tests for the shared helpers.

Behavior preserved from before
- Service-type restrictions enforced via ModifyPlan (with a Create/Update fallback): Fanout is Compute-only; Brotli Compression and Image Optimizer are CDN-only; NGWAF's traffic_ramp only takes effect on CDN.
- Confirmed against the live API that Image Optimizer is rejected outright for wasm services, while Origin Inspector is valid on both.

Docs
- 11 new pages under docs/resources/ (old single-resource page/template removed), regenerated via tfplugindocs.
- examples/product-enablement/main.tf — worked example covering all 11 resources.

Tests
- internal/acceptance_tests/product_enablement_test.go — create/import coverage for every resource, service-type validation both directions, schema validators, in-place update paths for the 3 configurable resources, and service_id replace semantics.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-07-23 at 1 37 30 PM


internal/resources/productenablement/bot_management.go
```diff
@@ -0,0 +1,156 @@
...
+	resp.TypeName = req.ProviderTypeName + "_product_enablement_bot_management"
+}
+
+func (r *BotManagementResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Enables Bot Management on a service. Product Enablement operates on the service directly rather than a specific service version, so this resource is not tied to a `version` and applies immediately.",
+		Attributes: map[string]schema.Attribute{
+			"id":         idAttribute(),
+			"service_id": serviceIDAttribute("Bot Management"),
+			"content_guard": schema.StringAttribute{
```
anthony-gomez-fastly: should this be 1 word or two?

rcaril: This should be one, we refer to this as `contentguard` in our docs: https://www.fastly.com/documentation/reference/api/products/bot_management/

kpfleming (dismissed):
This will need some naming changes:

1. Each resource block should have a name which starts with `fastly_service`, because these enablements are service-specific. Future enablements will be account-specific, or possibly domain-specific, so the naming will need to provide a way to distinguish between them.

2. I don't think we need `enablement` in the resource block names. `fastly_service_product_bot_management` should be sufficient, since this design means that product is enabled if the block exists and disabled if it does not exist.

There will also be complications related to the differences between CDN and Compute services; as an example, NGWAF requires supplying a `traffic_ramp` value for CDN services, but that value cannot be supplied for a Compute service (the API will return an error). I'm not sure if that means we need separate resources for CDN and Compute services or if a different strategy would work.

internal/resources/productenablement/simple.go
```diff
@@ -0,0 +1,325 @@
...
+		return
+	}
+
+	plan.ID = plan.ServiceID
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
+
+// Read treats any error from the product's Get endpoint - whether because
+// the product was disabled outside Terraform or because the service itself
+// is gone - as "no longer enabled" and removes the resource from state.
```
kpfleming: Is this safe to do for *any* error? What if the error is 5xx, indicating some sort of temporary API issue unrelated to the service or product?

I suspect this may have to be restricted to only `HTTPNotFound` errors, and leave other errors to cause their normal behavior.

kpfleming (dismissed):
Overall this looks fantastic, I just left one question about error-handling.

internal/resources/productenablement/simple.go
anthony-gomez-fastly: good call, added a check for 404's in `a98d0026`

kpfleming (approved):

PR #1362, opened 2026-07-27
add baseline acceptance test suite

anthony-gomez-fastly:
All Submissions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,9 +56,38 @@ test-acc:
 # One canonical happy-path test per resource family/feature, picked to fail fast (~2.5m)
 # on a broken build without running the full acceptance suite. See "Baseline regression
 # suite" in CLAUDE.md for the criteria used to decide whether a new test belongs here.
-BASELINE_TESTS := TestAccProvider_ConfigureWithAPIToken$$|TestAccFastlyServiceCDN_basic$$|TestAccFastlyServiceCompute_basic$$|TestAccFastlyServiceCDNAuto_basic$$|TestAccFastlyServiceComputeAuto_basic$$|TestAccFastlyServiceBackend_basic$$|TestAccFastlyServiceDomain_basic$$|TestAccFastlyACL_basic$$|TestAccFastlyKVStore_basic$$|TestAccFastlyServiceLoggingS3_basic$$|TestAccFastlyServiceCDNAuto_withLoggingS3$$|TestAccFastlyProductEnablement_cdnBasic$$|TestAccFastlyProductEnablement_computeBasic$$|TestAccFastlyServiceResourceLink_ACL$$
+empty :=
+space := $(empty) $(empty)
+
+BASELINE_TEST_NAMES := \
+	TestAccProvider_ConfigureWithAPIToken \
+	TestAccFastlyServiceCDN_basic \
+	TestAccFastlyServiceCompute_basic \
+	TestAccFastlyServiceCDNAuto_basic \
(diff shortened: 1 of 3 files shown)


anthony-gomez-fastly:
adding @rcaril for claude.md review

Makefile
```diff
@@ -51,6 +53,16 @@ test-acc:
 	@echo "    Note: This requires FASTLY_API_TOKEN to be set"
 	@TF_ACC=1 $(GO_BIN) test -count=1 -v -timeout 30m ./internal/acceptance_tests -run 'TestAcc.*$(KEYWORD)'
 
+# One canonical happy-path test per resource family/feature, picked to fail fast (~2.5m)
+# on a broken build without running the full acceptance suite. See "Baseline regression
+# suite" in CLAUDE.md for the criteria used to decide whether a new test belongs here.
```
rcaril, line 58: Does it make sense to add this information on this baseline test to `[TESTING.md](https://github.com/fastly/terraform-provider-fastly/blob/dual-model-framework-rewrite/TESTING.md)` instead of the `CLAUDE.md` file (or at least the reference here)?

Makefile
```diff
@@ -51,6 +53,16 @@ test-acc:
...
+BASELINE_TESTS := TestAccProvider_ConfigureWithAPIToken$$|TestAccFastlyServiceCDN_basic$$|TestAccFastlyServiceCompute_basic$$|TestAccFastlyServiceCDNAuto_basic$$|TestAccFastlyServiceComputeAuto_basic$$|TestAccFastlyServiceBackend_basic$$|TestAccFastlyServiceDomain_basic$$|TestAccFastlyACL_basic$$|TestAccFastlyKVStore_basic$$|TestAccFastlyServiceLoggingS3_basic$$|TestAccFastlyServiceCDNAuto_withLoggingS3$$|TestAccFastlyProductEnablement_cdnBasic$$|TestAccFastlyProductEnablement_computeBasic$$|TestAccFastlyServiceResourceLink_ACL$$
```
philippschulte: I would highly recommend to validate that every baseline test name still exists before running the suite. I believe that `go test -run` succeeds when a pattern matches no tests but I can be wrong. If I am correct then renaming or deleting one test would silently remove that coverage while `make test-baseline` continued to pass.

kpfleming: `Makefile` syntax supports `+=`, which allows constructs like this to be spread across multiple lines with one entry per line. This increases readability and maintainability.

However it also puts a 'space' character between entries, which may cause the regex matching to fail. If you haven't already tried that, it's likely worth a test to see if it produces the desired result.

philippschulte (dismissed):
The baseline suite is a useful addition. Please add a check that every configured baseline test name resolves before running the suite.

Makefile
anthony-gomez-fastly: good call, i'll add a note there

anthony-gomez-fastly: addressed in `8d79e877`

Makefile
anthony-gomez-fastly: addressed in `8d79e877`

anthony-gomez-fastly: addressed in `8d79e877`

kpfleming (approved):
Very nice.

philippschulte (approved):
Thanks for the update. This addresses my previous review comment.

Makefile
rcaril: Thanks! I think I would still prefer the reference to point to the Testing MD file instead of the `CLAUDE.md` if possible.

PR #1365, opened 2026-07-28
[CDTOOL-1500] Add fastly_service_condition resource and wire into servicecdnauto

anthony-gomez-fastly:
### Change summary

Scaffold the condition package following backend's template (schema, resource, expand, flatten, list) and register it as a standalone explicit resource plus list resource. Wire condition.NestedModel into servicecdnauto's Create/Read/ Update so it's usable as a nested block in the _auto family (CDTOOL-1500).

The Fastly API doesn't support updating a condition's type via PUT, so ops.Update deletes and recreates the condition when type changes instead of issuing a no-op PUT.

Fix reconcile ordering in servicecdnauto: conditions must be reconciled before backend/gzip, since both can reference a condition by name and the API rejects a backend/gzip create naming a condition that doesn't exist yet in that version.

Add acceptance tests for the standalone resource, the nested auto-family block, and end-to-end wiring through backend.request_condition and gzip.cache_condition (including a regression test for the reconcile-order fix). Add example usage docs and update the orchestration-cdn-auto example to demonstrate a condition driving a gzip cache_condition.

 All Submissions:

### New Feature Submissions:

* [x] Post the output of your test 
Screenshot 2026-07-28 at 1 32 21 PM
runs

### Changes to Core Features:


internal/resources/condition/schema.go
```diff
@@ -0,0 +1,177 @@
...
+	Statement types.String `tfsdk:"statement"`
+	Priority  types.Int64  `tfsdk:"priority"`
+}
+
+func (n NestedModel) ModelsEqual(other NestedModel) bool {
+	return service.StringValue(n.Name) == service.StringValue(other.Name) &&
+		service.StringValue(n.Type) == service.StringValue(other.Type) &&
+		service.StringValue(n.Statement) == service.StringValue(other.Statement) &&
+		service.Int64Value(n.Priority) == service.Int64Value(other.Priority)
+}
```
philippschulte, line 34: After comparing this with the current provider implementation, both providers have inconsistent whitespace handling. The current provider trims on create but reads the API value back directly, and normal updates can send the untrimmed modified value. This implementation improves request construction by consistently trimming outgoing statements, but `ModelsEqual` still compares the raw configured value against the API value.

For example, a heredoc such as:

```hcl
statement = <<-VCL
  req.url.path ~ "^/admin/"
VCL
```

typically contains a trailing newline in configuration. The provider sends the trimmed statement, and Fastly returns it without that newline, so raw equality can report a difference even though the condition is semantically unchanged.

Could we normalize both sides in `ModelsEqual` and add an acceptance test using a heredoc statement followed by a clean `PlanOnly` step? I'll have to do the same for the vcl and snippet implementations.

philippschulte (dismissed):
Thanks for adding condition support. The implementation look solid.

I left one blocking comment on statement whitespace normalization. Both the current and new providers can drift when heredoc statements include trailing whitespace, so this should be addressed in the rewrite and covered with a clean `PlanOnly` acceptance test.

internal/resources/condition/schema.go
anthony-gomez-fastly: done in `4ae4f528`

philippschulte (approved):
Thanks for the update. This addresses my previous blocking comment. No further issues from my side.

PR #1367, opened 2026-07-28
[CDTOOL-1542] Add custom VCL support to dual-model provider

philippschulte:
### Change summary

This PR adds custom VCL support to the dual-model provider.

It introduces:

- a new explicit `fastly_service_vcl` resource
- a nested `vcl` block for `fastly_service_cdn_auto`
- reconciliation of custom VCL files by `name`
- validation for duplicate VCL names and invalid `main` selection
- support for both inline VCL content and `file(...)`-backed VCL content
- unit and acceptance test coverage for explicit and automatic resource families
- generated documentation and documentation templates for `fastly_service_vcl`

## Improved VCL diffs

The current SDKv2 provider models custom VCL blocks as a set. Because set element
identity includes the VCL content, changing one line in a VCL file causes
Terraform to render the whole block as removed and re-added.

The dual-model implementation uses named nested VCL blocks and reconciles them by
`name`, which produces a focused in-place content diff.

Local test output after changing one line in a VCL file:

```diff
Terraform will perform the following actions:

  # fastly_service_cdn_auto.vcl_diff_demo will be updated in-place
  ~ resource "fastly_service_cdn_auto" "vcl_diff_demo" {
      ~ active_version  = 1 -> (known after apply)
        id              = "LhwYo8IyY5cuqeA6DerybP"
      ~ managed_version = 1 -> (known after apply)
        name            = "tf-vcl-diff-demo"
        # (3 unchanged attributes hidden)

      ~ vcl {
          ~ content = <<-EOT
                sub vcl_recv {
                #FASTLY recv
                  return(lookup);
                }

                sub vcl_hash {
                  set req.hash += req.url;
                  set req.hash += req.http.host;
                #FASTLY hash
                  return(hash);
                }

                sub vcl_hit {
                #FASTLY hit
                  return(deliver);
                }

                sub vcl_miss {
                #FASTLY miss
                  return(fetch);
                }

                sub vcl_pass {
                #FASTLY pass
                  return(pass);
                }

                sub vcl_fetch {
                #FASTLY fetch
                  return(deliver);
                }

                sub vcl_error {
                #FASTLY error
                  return(deliver);
                }

                sub vcl_deliver {
                #FASTLY deliver
              -   set resp.http.X-VCL-Diff-Demo = "before";
              +   set resp.http.X-VCL-Diff-Demo = "after";
                  return(deliver);
                }

                sub vcl_log {
                #FASTLY log
                }
            EOT
            name    = "main"
            # (1 unchanged attribute hidden)
        }

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

### New Feature Submissions:

```
=== RUN   TestAccFastlyServiceVCL_basicAndUpdate
=== PAUSE TestAccFastlyServiceVCL_basicAndUpdate
=== RUN   TestAccFastlyServiceCDNAuto_withVCLAndContentUpdate
=== PAUSE TestAccFastlyServiceCDNAuto_withVCLAndContentUpdate
=== RUN   TestAccFastlyServiceCDNAuto_withMultipleVCLFiles
=== PAUSE TestAccFastlyServiceCDNAuto_withMultipleVCLFiles
=== RUN   TestAccFastlyServiceCDNAuto_VCLValidation
=== PAUSE TestAccFastlyServiceCDNAuto_VCLValidation
=== RUN   TestAccFastlyServiceCDNAuto_VCLValidationNoMain
=== PAUSE TestAccFastlyServiceCDNAuto_VCLValidationNoMain
=== RUN   TestAccFastlyServiceCDNAuto_VCLValidationInvalidSyntax
=== PAUSE TestAccFastlyServiceCDNAuto_VCLValidationInvalidSyntax
=== CONT  TestAccFastlyServiceVCL_basicAndUpdate
=== CONT  TestAccFastlyServiceCDNAuto_VCLValidation
=== CONT  TestAccFastlyServiceCDNAuto_withMultipleVCLFiles
=== CONT  TestAccFastlyServiceCDNAuto_VCLValidationInvalidSyntax
=== CONT  TestAccFastlyServiceCDNAuto_VCLValidationNoMain
=== CONT  TestAccFastlyServiceCDNAuto_withVCLAndContentUpdate
--- PASS: TestAccFastlyServiceCDNAuto_VCLValidation (0.51s)
--- PASS: TestAccFastlyServiceCDNAuto_VCLValidationNoMain (0.50s)
--- PASS: TestAccFastlyServiceCDNAuto_VCLValidationInvalidSyntax (39.40s)
--- PASS: TestAccFastlyServiceVCL_basicAndUpdate (54.97s)
--- PASS: TestAccFastlyServiceCDNAuto_withMultipleVCLFiles (71.42s)
--- PASS: TestAccFastlyServiceCDNAuto_withVCLAndContentUpdate (100.49s)
```

### DIFF
--- a/internal/resources/servicecdnauto/resource.go
+++ b/internal/resources/servicecdnauto/resource.go
@@ -42,6 +42,7 @@ const imageOptimizerImportedPrivateKey = "image_optimizer_default_settings_impor
 var _ resource.Resource = &Resource{}
 var _ resource.ResourceWithConfigure = &Resource{}
 var _ resource.ResourceWithImportState = &Resource{}
+var _ resource.ResourceWithValidateConfig = &Resource{}
 
 func NewResource() resource.Resource {
 	return &Resource{}
@@ -136,6 +137,22 @@ func (r *Resource) Configure(_ context.Context, req resource.ConfigureRequest, r
 	r.providerData = data
 }
 
+func (r *Resource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	var config Model
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if err := vcl.ValidateConfig(config.VCL); err != nil {
+		resp.Diagnostics.AddAttributeError(
+			path.Root("vcl"),
+			"Invalid custom VCL configuration",
+			err.Error(),
+		)
+	}
+}
+
 func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
 	var plan Model
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
@@ -144,7 +161,11 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp 
 	}
 
 	if err := vcl.Validate(plan.VCL); err != nil {
-		resp.Diagnostics.AddError("Invalid custom VCL configuration", err.Error())
+		resp.Diagnostics.AddAttributeError(
+			path.Root("vcl"),
+			"Invalid custom VCL configuration",
+			err.Error(),
+		)
 		return
 	}
 
@@ -266,13 +287,13 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp 
 	plan.ImageOptimizerDefaultSettings = imageOptimizerDefaultSettings
 
 	if err := vcl.Reconcile(ctx, r.providerData.AutoClient(), serviceID, version, plan.VCL); err != nil {
-		resp.Diagnostics.AddError("Error reconciling custom VCL", err.Error())
+		resp.Diagnostics.AddError("Error reconciling custom VCL files", err.Error())
 		return
 	}
 
 	vcls, err := vcl.ReadForVersion(ctx, r.providerData.AutoClient(), serviceID, version)
 	if err != nil {
-		resp.Diagnostics.AddError("Error reading custom VCL", err.Error())
+		resp.Diagnostics.AddError("Error reading custom VCL files", err.Error())
 		return
 	}
 	plan.VCL = vcl.MatchOrderPreservePlanContent(vcls, plan.VCL)
@@ -380,12 +401,6 @@ func (r *Resource) Read(ctx context.Context, req resource.ReadRequest, resp *res
 		resp.Diagnostics.AddError("Error reading New Relic OTLP logging endpoints", err.Error())
 		return
 	}
-	vcls, err := vcl.ReadForVersion(ctx, r.providerData.AutoClient(), state.ID.ValueString(), readVersion)
-	if err != nil {
-		resp.Diagnostics.AddError("Error reading custom VCL", err.Error())
-		return
-	}
-
 	state.Domain = domain.MatchOrder(domains, state.Domain)
 	state.Backend = backend.MatchOrder(backends, state.Backend)
 	state.ACL = cdnacl.MatchOrder(acls, state.ACL)
@@ -393,6 +408,12 @@ func (r *Resource) Read(ctx context.Context, req resource.ReadRequest, resp *res
 	state.Gzip = gzip.MatchOrder(gzips, state.Gzip)
 	state.LoggingS3 = loggings3.MatchOrder(loggingS3s, state.LoggingS3)
 	state.LoggingNewRelicOTLP = loggingnewrelicotlp.MatchOrder(loggingNewRelicOTLPs, state.LoggingNewRelicOTLP)
+
+	vcls, err := vcl.ReadForVersion(ctx, r.providerData.AutoClient(), state.ID.ValueString(), readVersion)
+	if err != nil {
+		resp.Diagnostics.AddError("Error reading custom VCL files", err.Error())
+		return
+	}
 	state.VCL = vcl.MatchOrder(vcls, state.VCL)
 
 	importedBytes, diags := req.Private.GetKey(ctx, imageOptimizerImportedPrivateKey)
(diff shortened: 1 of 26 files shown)


internal/resources/vcl/schema.go
```diff
@@ -0,0 +1,200 @@
...
+		boolplanmodifier.RequiresReplace(),
+	}
+	attrs["main"] = mainAttr
+
+	return attrs
+}
+
+func NestedBlockSchema() schema.ListNestedBlock {
+	return schema.ListNestedBlock{
+		Description: "Custom VCL files attached to this service. This is modeled as a list so Terraform can render changes to `content` as an in-place string diff instead of a whole set element delete/add.",
```
rcaril: Does it make sense to keep this last bit as a customer facing description instead of a internal comment? 
```
 This is modeled as a list so Terraform can render changes to `content` as an in-place string diff instead of a whole set element delete/add.
```

philippschulte: I removed the sentence entirely in 8b908a3f0ab7db808146488e1668e765f73866c8  because the implementation to accomplish a clean rendered diff is not important to our customers.

internal/resources/vcl/schema.go
```diff
@@ -0,0 +1,200 @@
...
+
+func CommonAttributes() map[string]schema.Attribute {
+	return map[string]schema.Attribute{
+		"name": schema.StringAttribute{
+			Required:    true,
+			Description: "A unique name for this custom VCL file. Included VCL files must be referenced by this exact name from the main VCL file.",
+		},
+		"content": schema.StringAttribute{
+			Required:    true,
+			Description: "The custom VCL source code to upload. Commonly configured with file(\"${path.module}/main.vcl\") or templatefile(...).",
```
rcaril: What you have is fine, but without knowing what every user does - I think the phrase `Can be` also works. 
```suggestion
			Description: "The custom VCL source code to upload. Can be configured with file(\"${path.module}/main.vcl\") or templatefile(...).",
```

rcaril (changes requested):
Most of my local testing had great results. However, I found when passing a newline character via file `\n` - the newline was not interpreted correctly and resulted in direct characters renders instead of sending the newline itself: 

IE:
```
  vcl {
    name = "my custom VCL - from a file"
    main = false
    content = file("myVCL.txt")
  }
```
_myVCL.txt_
```
## this is my VCL\n##on two lines\n##from a external file
```
Results as:

Image

Where as sending the newlines directly via the `content` attributem the newline appears as expected:
```
  vcl {
    name = "my custom VCL - two"
    main = true
    content = "## This is my whole VCL block - again!\n#this is my second line I am adding"
  }
```

Results as:

Image

internal/resources/vcl/resource.go
```diff
@@ -0,0 +1,243 @@
...
+				"name":       state.Name.ValueString(),
+			})
+			resp.State.RemoveResource(ctx)
+			return
+		}
+		resp.Diagnostics.AddError("Error reading custom VCL", err.Error())
+		return
+	}
+
+	flatten(ctx, v, &state)
```
anthony-gomez-fastly, line 128: is there a reason you don't wrap this in the `plannedContent := plan.Content` `plan.Content = plannedContent` block? wouldn't that help prevent whitespace issues that the other calls catch?

internal/resources/servicecdnauto/resource.go
```diff
@@ -137,6 +140,11 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp
 		return
 	}
 
+	if err := vcl.Validate(plan.VCL); err != nil {
```
anthony-gomez-fastly, line 163: do we want to also add plan validation? i believe this only validates on apply?

philippschulte:
> Most of my local testing had great results. However, I found when passing a newline character via file `\n` - the newline was not interpreted correctly and resulted in direct characters renders instead of sending the newline itself:
> 
> IE:
> 
> ```
>   vcl {
>     name = "my custom VCL - from a file"
>     main = false
>     content = file("myVCL.txt")
>   }
> ```
> 
> _myVCL.txt_
> 
> ```
> ## this is my VCL\n##on two lines\n##from a external file
> ```
> 
> Results as:
> Image
> 
> Where as sending the newlines directly via the `content` attributem the newline appears as expected:
> 
> ```
>   vcl {
>     name = "my custom VCL - two"
>     main = true
>     content = "## This is my whole VCL block - again!\n#this is my second line I am adding"
>   }
> ```
> 
> Results as:
> Image

Thanks for testing this. I believe this is expected Terraform behavior rather than provider-specific behavior, and it matches the current provider.

`file("myVCL.txt")` reads file contents literally. If the file contains the two characters `\` and `n`, Terraform passes those characters through unchanged. Inline HCL strings interpret escapes such as `\n`, which is why the inline example produces actual line breaks.

I don't think the provider should automatically decode literal `\n` sequences because valid VCL may intentionally contain those characters inside strings. Users who want multiline VCL from a file should put actual newlines in the file. For inline configuration, quoted strings and HEREDOCs work as expected.

I added HEREDOC acceptance coverage to confirm multiline VCL content works correctly in the dual-model implementation.

internal/resources/vcl/schema.go
philippschulte: Changed in f4de915a79eb7f4f0e5c910a8d0b0418c13078e8.

internal/resources/vcl/resource.go
philippschulte: Good question. The planned-content preservation is needed in Create/Update because Terraform requires the post-apply state for configured attributes to match the planned value.

Read does not have access to the plan/config, only prior state. During refresh, the provider should report the remote API value. Preserving prior state content in Read would risk hiding real remote drift, so I kept Read as a plain `flatten(ctx, v, &state)`.

internal/resources/servicecdnauto/resource.go
philippschulte: Good catch! Added in da4c3679b160991f9e78c2ef8bdfcf138ef14903.

rcaril (approved):
Looks good from my side, I agree that `\n` behavior is Terraform specific. Thanks for double checking.

anthony-gomez-fastly (approved):
LGTM! ty

PR #1374, opened 2026-08-03
[CDTOOL-1691] add support for datadog, jira, jsm, opsgenie, and splunkoncall integrations

anthony-gomez-fastly:
### Change summary
add support for datadog, jira, jsm, opsgenie, and splunkoncall integrations

All Submissions:

### New Feature Submissions:

Screenshot 2026-08-03 at 4 35 27 PM

### DIFF
--- a/fastly/resource_fastly_integration.go
+++ b/fastly/resource_fastly_integration.go
@@ -42,9 +42,21 @@ func resourceFastlyIntegration() *schema.Resource {
 			"type": {
 				Type:        schema.TypeString,
 				Required:    true,
-				Description: "Type of the integration. One of: `mailinglist`, `microsoftteams`, `newrelic`, `pagerduty`, `slack`, `webhook`.",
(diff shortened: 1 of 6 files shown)


fastly/resource_fastly_integration.go
```diff
@@ -42,9 +42,21 @@ func resourceFastlyIntegration() *schema.Resource {
...
+						gofastly.IntegrationTypeJiraIssue,
+						gofastly.IntegrationTypeJSM,
+						"mailinglist",
+						"microsoftteams",
+						"newrelic",
+						gofastly.IntegrationTypeOpsGenie,
+						"pagerduty",
+						"slack",
+						gofastly.IntegrationTypeSplunkOnCall,
+						"webhook",
```
kpfleming: We should probably use string literals here instead of using constants from go-fastly, to be consistent and also because we have to use them in the `Description` attribute just above so they have to match.

Alternatively this schema object could be code-generated and only use the constants from go-fastly for both the `Description` and `ValidateDiagFunc` attributes.

kpfleming (dismissed):
One small change requested, otherwise this looks good.

kpfleming (approved):

PR #1383, opened 2026-08-06
[CDTOOL-1508] Add Support for BigQuery Logging to the Dual-Model Provider

rcaril:
### Change summary

This PR adds support for BigQuery Logging to the Dual-Model Provider. 

 All Submissions:

### New Feature Submissions:

```
--- PASS: TestAccFastlyServiceLoggingBigQuery_computeRejectsVCLOnlyFields (30.88s)
--- PASS: TestAccFastlyServiceLoggingBigQuery_importBasic (47.09s)
--- PASS: TestAccFastlyServiceLoggingBigQuery_computeConsistentAfterApply (50.30s)
--- PASS: TestAccFastlyServiceLoggingBigQuery_formatDefault (54.75s)
--- PASS: TestAccFastlyServiceLoggingBigQuery_basic (59.24s)
--- PASS: TestAccFastlyServiceLoggingBigQuery_clearToDefaults (65.42s)
--- PASS: TestAccFastlyServiceLoggingBigQuery_update (71.86s)
--- PASS: TestAccFastlyServiceLoggingBigQuery_versionUpdateInPlace (76.19s)
--- PASS: TestAccFastlyServiceLoggingBigQuery_placementUnsetVsNone (80.36s)
--- PASS: TestAccFastlyServiceCDNAuto_withBackendAndLoggingBigQuery (82.07s)
--- PASS: TestAccFastlyServiceComputeAuto_withLoggingBigQuery (82.60s)
--- PASS: TestAccFastlyServiceCDNAuto_withMultipleLoggingBigQuery (90.73s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingBigQuery (91.64s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingBigQueryRemoved (92.24s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingBigQueryUpdate (97.91s)
--- PASS: TestAccFastlyServiceCDNAuto_loggingBigQueryPlacementUnsetVsNone (109.87s)
PASS
ok 
```

### Changes to Core Features:

### Additional Notes:

The prior env variable for the `account` attribute has been updated from ` FASTLY_GCS_ACCOUNT_NAME` to `FASTLY_GOOGLE_SERVICE_ACCOUNT_NAME`, which more closely relates to our language in the UI and API docs.


internal/resources/loggingbigquery/schema.go
```diff
@@ -0,0 +1,491 @@
...
+	return "value defaults to the FASTLY_GOOGLE_SERVICE_ACCOUNT_NAME, FASTLY_BQ_EMAIL, and FASTLY_BQ_SECRET_KEY environment variables"
+}
+
+func (d authenticationEnvDefault) MarkdownDescription(ctx context.Context) string {
+	return d.Description(ctx)
+}
+
+func (authenticationEnvDefault) DefaultObject(ctx context.Context, _ fwdefaults.ObjectRequest, resp *fwdefaults.ObjectResponse) {
+	resp.PlanValue = NewAuthenticationObject(
+		envStringDefault(ctx, "FASTLY_GOOGLE_SERVICE_ACCOUNT_NAME"),
```
rcaril: Updated to the more clear naming of `FASTLY_GOOGLE_SERVICE_ACCOUNT_NAME` from the prior `FASTLY_GCS_ACCOUNT_NAME` variable in the live provider. This more closely relates to our language in the UI and API docs.

internal/resources/loggingbigquery/schema.go
```diff
@@ -0,0 +1,491 @@
...
+			//
+			// On a VCL service the API resolves this to exactly what was configured.
+			// Compute (wasm) services are the exception — the API forces "none" there
+			// regardless of what was sent, which can never match the planned null, so
+			// ResetVCLOnlyToDefaults discards it after apply for those services.
+			Optional: true,
+			Validators: []validator.String{
+				stringvalidator.OneOf("none"),
+			},
+			Description: "Where in the generated VCL the logging call should be placed. If not set, endpoints with format_version of 2 are placed in vcl_log and those with format_version of 1 are placed in vcl_deliver. Valid value is `none`.",
```
kpfleming:
```suggestion
			Description: "Where in the generated VCL the logging call should be placed. If not set, endpoints with `format_version` of `2` are placed in `vcl_log` and those with `format_version` of `1` are placed in `vcl_deliver`. Valid value is `none`.",
```

internal/resources/servicecomputeauto/resource.go
```diff
@@ -387,7 +408,7 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
 		return
 	}
 
-	nestedChanged := !domain.Equal(plan.Domain, state.Domain) || !backend.Equal(plan.Backend, state.Backend) || !resourcelink.Equal(plan.ResourceLink, state.ResourceLink) || !computepackage.Equal(plan.Package, state.Package) || !loggings3.ComputeEqual(plan.LoggingS3, state.LoggingS3) || !loggingnewrelicotlp.ComputeEqual(plan.LoggingNewRelicOTLP, state.LoggingNewRelicOTLP) || !loggingdatadog.ComputeEqual(plan.LoggingDatadog, state.LoggingDatadog)
+	nestedChanged := !domain.Equal(plan.Domain, state.Domain) || !backend.Equal(plan.Backend, state.Backend) || !resourcelink.Equal(plan.ResourceLink, state.ResourceLink) || !computepackage.Equal(plan.Package, state.Package) || !loggings3.ComputeEqual(plan.LoggingS3, state.LoggingS3) || !loggingnewrelicotlp.ComputeEqual(plan.LoggingNewRelicOTLP, state.LoggingNewRelicOTLP) || !loggingdatadog.ComputeEqual(plan.LoggingDatadog, state.LoggingDatadog) || !loggingbigquery.ComputeEqual(plan.LoggingBigQuery, state.LoggingBigQuery)
```
kpfleming, line 411: I wish there was a better way to write this in Go but I can't think of one :-(

kpfleming (dismissed):
One small text markup change.

internal/resources/loggingbigquery/schema.go
rcaril: Oops, good catch. I'm normally on top of the back ticks 😓

internal/resources/loggingbigquery/schema.go
philippschulte: The new name is definitely clearer, but should we preserve `FASTLY_GCS_ACCOUNT_NAME` as a fallback for compatibility with the current provider?

Users may already rely on that environment variable without having it represented anywhere in their Terraform configuration, so changing the name as part of the rewrite could silently break authentication after upgrading. We could prefer `FASTLY_GOOGLE_SERVICE_ACCOUNT_NAME` while falling back to the legacy variable when the new one is unset.

internal/resources/loggingbigquery/expand.go
```diff
@@ -0,0 +1,134 @@
...
+		NewName:        new(service.StringValue(m.Name)),
+		ProjectID:      new(service.StringValue(m.ProjectID)),
+		Dataset:        new(service.StringValue(m.Dataset)),
+		Table:          new(service.StringValue(m.Table)),
+	}
+
+	// Unlike email/secret_key below, account_name can't be cleared via update:
+	// the API rejects an explicit empty string ("Invalid account_name ''"), so
+	// it must be omitted (fastly.NullString) rather than sent as "" (new()).
+	input.AccountName = fastly.NullString(service.StringValue(m.AccountName()))
```
philippschulte, line 71: What happens when an existing endpoint changes from `account_name` authentication to `email`/`secret_key`?

When the desired `account_name` is empty we correctly omit it here because the API rejects an explicit empty value, but omitting it also means Fastly retains the previously configured account name. The subsequent read would then return the old value while Terraform planned it as unset, which can produce drift/inconsistent state.

Could we handle the non-empty -> empty transition explicitly, likely by delete/recreate, and cover it with an acceptance test?

internal/resources/loggingbigquery/schema.go
```diff
@@ -0,0 +1,491 @@
...
+				"secret_key": schema.StringAttribute{
+					Optional:  true,
+					Computed:  true,
+					Sensitive: true,
+					Default:   defaults.EnvString("FASTLY_BQ_SECRET_KEY", ""),
+					Validators: []validator.String{
+						notTrimmed{},
+					},
+					Description: "The `private_key` field in your service account authentication JSON. Not required if `account_name` is provided. Can be set via the `FASTLY_BQ_SECRET_KEY` environment variable.",
+				},
```
philippschulte, line 287: The docs say authentication must be either `account_name` or `both email` + `secret_key`, while all three fields are optional/defaultable.  It would be great if we could add cross-field validation enforcing that combination.

philippschulte (changes requested):
BigQuery support looks well integrated with the existing dual-model logging resources!

I left one blocking comment around switching away from `account_name` authentication and two comments for suggestions.

internal/resources/loggingbigquery/schema.go
rcaril: Thinking about this more over the weekend, i'm a bit split on this. One may consider the `FASTLY_GCS_ACCOUNT_NAME` to actually be incorrect usage. While I understand the ease of use by settings a fallback, I wonder if we should instead use one of these strategies: 

- A helper that errors out suggesting to move to the new env var when `FASTLY_GCS_ACCOUNT_NAME` is found
- A helper that let's the user know that `FASTLY_GCS_ACCOUNT_NAME` is deprecated and they should use `FASTLY_GOOGLE_SERVICE_ACCOUNT_NAME` instead 

If we don't agree with either of these, I think it's okay to use a fallback, but I think considering this is a new provider, we can make sure we clearly document breaking changes and consider adding a script that helps find items that have changed - such as this.

internal/resources/servicecomputeauto/resource.go
rcaril: Indeed =/

internal/resources/loggingbigquery/expand.go
rcaril: Good catch, I see the same when testing:

```
  ~ logging_bigquery {
          ~ authentication     = {
              ~ account_name = "123" -> ""
              + email        = (sensitive value)
              + secret_key   = (sensitive value)
```

`apply` ->

```

│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to fastly_service_cdn_auto.example, provider
│ "provider[\"registry.terraform.io/fastly/fastly\"]" produced an unexpected new
│ value: .logging_bigquery[0].authentication: inconsistent values for sensitive
│ attribute.
│ 
│ This is a bug in the provider, which should be reported in the provider's own
│ issue tracker.
```

internal/resources/loggingbigquery/schema.go
rcaril: Yup, this was a gap from `main`. Added validation back in [ae623de](https://github.com/fastly/terraform-provider-fastly/pull/1383/commits/ae623de27f401a37464cc57d2af7c632b34c2ace).

internal/resources/loggingbigquery/expand.go
rcaril: I'm determining if we should recreate this resource or do an upstream change and leverage the `nullable` helper here. I'll follow up here once we decide.

rcaril: Went with the recreate approach in [dd89aac](https://github.com/fastly/terraform-provider-fastly/pull/1383/commits/dd89aac5764bb251581069822f3a4be1ffbb68bc).

rcaril:
```
=== RUN   TestAccFastlyServiceCDNAuto_loggingBigQueryAccountNameToEmailSecretKey
=== PAUSE TestAccFastlyServiceCDNAuto_loggingBigQueryAccountNameToEmailSecretKey
=== CONT  TestAccFastlyServiceCDNAuto_loggingBigQueryAccountNameToEmailSecretKey
--- PASS: TestAccFastlyServiceCDNAuto_loggingBigQueryAccountNameToEmailSecretKey (43.43s)
PASS
ok    
```

internal/resources/loggingbigquery/schema.go
philippschulte: @verythorough please weigh in on this. Thanks in advance!

philippschulte (approved):
Thanks for the update.

My only remaining concern is preserving the existing `FASTLY_GCS_ACCOUNT_NAME` environment variable as a fallback when introducing `FASTLY_GOOGLE_SERVICE_ACCOUNT_NAME`, so upgrading users don't silently lose their authentication configuration.

internal/resources/loggingbigquery/schema.go
verythorough: On the one hand, this is a perfect time for breaking changes, and it would be great not to have to keep deprecated variable names in the new provider. However, we also want to make the migration as smooth and easy as possible for our current users, and I'm not sure this is worthy of the speed bump it could create.

Looking at the suggestions so far, I think this is the resulting experience:
- Fallback only: everything works smoothly, but the user has no idea they're using a deprecated variable name.
- Helper that errors: user can't complete their migration until they change their variable name; whether they choose to search for external uses of it is up to them, but at least they know.
- Helper that warns: user can complete their migration even with the deprecated variable name, but they may not notice the warning and may have incomplete context on any external effects that break

To be honest, I'm not sure whether the warning option included the fallback already, but I'm thinking that "fallback + warning" would provide the smoothest, best informed experience for the user overall, in exchange for a small about of code cruft.

rcaril: > "fallback + warning" would provide the smoothest, best informed experience for the user overall, in exchange for a small about of code cruft.

I agree, i'll implement this. Thanks for your thoughts and input!

rcaril: Added the fallback behavior in [2015123](https://github.com/fastly/terraform-provider-fastly/pull/1383/commits/201512334fa67e97921d8d3f3095c305c8070563). 

Using `FASTLY_GCS_ACCOUNT_NAME` will now result in the following deprecation message:
image

PR #1384, opened 2026-08-06
[CDTOOL-1706] Ensure that Compute and VCL attribute blocks are sequenced correctly

rcaril:
### Change summary

This PR addresses an issue where adding a `healthcheck` block to a Compute resources while assigning said healthcheck to a `backend` block results in the incorrect sequencing order where the backend changes are processed first - resulting in a API `400` with a `  Detail: No healthcheck named ...` error. 

We are addressing this with a simple changing of lowering the order of which we process `backend` blocks for Compute services, ensuring that healthchecks fire off first. 

-- 

Upon further codebase review, another discrepancy was found with VCL services: `rate_limiter` blocks were processed before the `dictionary` / `response_object` blocks in which they referenced by name, so referencing one created in the same apply would also fail. This has also been addressed. 

 All Submissions:

### New Feature Submissions:

```
TF_ACC=1  go test ./fastly/ -run 'TestAccFastlyServiceCompute_basic|TestAccFastlyServiceCompute_stage'
ok      github.com/fastly/terraform-provider-fastly/fastly      51.495s
```
```
TF_ACC=1  go test ./fastly/ -run 'TestAccFastlyServiceVCLRateLimiter_orderingRegression'
ok      github.com/fastly/terraform-provider-fastly/fastly      15.441s
```

### Changes to Core Features:

### User Impact

- Users can now define and assign healthchecks on Compute services in the same apply without facing errors. 
- Users can now define rate_limiter blocks that reference a dictionary or response_object created in the same apply on VCL services without encountering errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### ENHANCEMENTS:
 
 ### BUG FIXES:
+- fix(service_compute): process `healthcheck` blocks before `backend` blocks so backends referencing a healthcheck created in the same apply do not fail with `No healthcheck named '<name>'`
 
 ### Dependencies
 - build(deps): `github.com/fastly/go-fastly/v17` from 17.1.0 to 17.2.0 ([#1382](https://github.com/fastly/terraform-provider-fastly/pull/1382))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### BUG FIXES:
 - fix(service_compute): process `healthcheck` blocks before `backend` blocks so backends referencing a healthcheck created in the same apply do not fail with `No healthcheck named '<name>'` ([#1384](https://github.com/fastly/terraform-provider-fastly/pull/1384))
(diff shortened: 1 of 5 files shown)


rcaril:
Validated locally that this fixes the reported behavior. 

Added specific tests to capture this behavior as well.

kpfleming:
It appears that the CDN service resource already did this in the correct order, but it could be useful to compare the orders of all the other nested resources between the two in order to find any other differences.

rcaril:
> It appears that the CDN service resource already did this in the correct order, but it could be useful to compare the orders of all the other nested resources between the two in order to find any other differences.


Good call, I found a VCL specific issue with `rate_limiter` blocks; they were processed before the `dictionary` / `response_object` blocks they can reference by name, so referencing one created in the same apply would also fail. 

Addressed in [9c953d1](https://github.com/fastly/terraform-provider-fastly/pull/1384/commits/9c953d10e4c4fa2005581b318ab8a6bebd6e1ba5).

anthony-gomez-fastly (approved):
LGTM

PR #1385, opened 2026-08-06
Condition acc test fixes

anthony-gomez-fastly:
### Change summary

- Fix fastly_service_condition's statement attribute, which was incorrectly trimmed before being sent to the API, causing provider produced inconsistent result after apply on Create when the statement came from a HEREDOC with a trailing newline. Verified against the live API that Fastly returns the statement byte-for-byte as submitted — no normalization needed on our side; removed the now-unnecessary plan modifier and trimming logic entirely.

- Fix a latent bug in the acceptance-test helper (normalizeNestedBlockIndent) that uniformly indents every line of a rendered nested block, including HEREDOC content — this was baked into actual VCL/snippet content sent to the API whenever a non-dash HEREDOC marker (<<VCL/<<SNIPPET) was used. Switched those block templates to the dash form (<<-VCL/<<-SNIPPET) so Terraform's own indentation-stripping cancels out the helper's added indent.

All Submissions:

### New Feature Submissions:

Screenshot 2026-08-06 at 6 31 45 PM

Screenshot 2026-08-06 at 6 33 33 PM


anthony-gomez-fastly:
I found these while running the full suite for another change

it looks like the api _doesn't_ strip out newlines as we had thought?

philippschulte (approved):
The condition fix looks correct. Thank you!

PR #1386, opened 2026-08-06
fix(client): retry idempotent Fastly API requests on transient 5xx/429

anthony-gomez-fastly:
### Change summary

The go-fastly client has no built-in retry/backoff, so a transient 503 from the Fastly API immediately failed the whole apply/plan. Wrap the client transport with go-retryablehttp, scoped to HTTP-idempotent methods (GET/HEAD/PUT/DELETE) so retries can't resubmit a POST/PATCH write that may have already succeeded server-side.

Other major providers (google, AWS, and hashicorp) do similar things, so this is not unusual.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-08-06 at 6 59 16 PM


kpfleming (approved):
Nice!

rcaril:
While I understand the `5xx` retry, I don't think retrying on a `429` will help in any way - once a request is rate limited, I don't think we should be attempting so add to said limit as it won't recover until the following hour.

philippschulte (dismissed):
Thank you for adding this! The implementation looks very good but I would like to see a small unit test around the method-routing behavior before approving.

We don't need to re-test `go-retryablehttp` itself, but it would be useful to verify our contract that `GET/HEAD/PUT/DELETE` go through the retry path while `POST/PATCH` bypass it and are only attempted once.

kpfleming:
> While I understand the `5xx` retry, I don't think retrying on a `429` will help in any way - once a request is rate limited, I don't think we should be attempting so add to said limit as it won't recover until the following hour.

That's a great point - ideally the 'retry' wrapper would understand the standard rate-limiting headers in the response and use them to schedule the retry.

anthony-gomez-fastly:
Screenshot 2026-08-07 at 11 15 14 AM

kpfleming (approved):

rcaril (approved):
Thank you!

glerchundi:
Fantastic fix, thanks!

We found this issue in our CI pipelines, two follow-up questions:
1. When it's planned to release the "dual-model-framework-rewrite" branch?
2. If it's still in the long-ride development cycle, does it make sense to backport this fix?

anthony-gomez-fastly:
@glerchundi 
1. We are working on getting a beta out for the new provider framework soon!
2. We are not currently planning to make this change for the current provider.

glerchundi:
Thanks for your response!

PR #1391, opened 2026-08-07
add optional enabled attribute to toggle products in-place

anthony-gomez-fastly:
### Change summary

Adds an `enabled` bool (default true) to bot_management, ddos_protection, ngwaf, and the simple product resources so they can be disabled/re-enabled via update instead of requiring create/destroy.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-08-07 at 2 18 11 PM

### DIFF
--- a/internal/resources/productenablement/bot_management.go
+++ b/internal/resources/productenablement/bot_management.go
@@ -121,7 +121,7 @@ func (r *BotManagementResource) Read(ctx context.Context, req resource.ReadReque
 	tflog.Debug(ctx, "Reading Fastly Product Enablement (bot_management)", map[string]any{"service_id": serviceID})
 
 	if _, err := botmanagement.Get(ctx, r.client, serviceID); err != nil {
-		if errors.IsNotFound(err) {
+		if errors.IsNotFound(err) || isProductDisabledError(err) {
 			tflog.Debug(ctx, "bot_management is disabled on service", map[string]any{"service_id": serviceID})
 			state.Enabled = types.BoolValue(false)
 		} else {
--- a/internal/resources/productenablement/bot_management.go
+++ b/internal/resources/productenablement/bot_management.go
@@ -157,6 +157,10 @@ func (r *BotManagementResource) Update(ctx context.Context, req resource.UpdateR
 	}
 
 	if enabled {
+		if _, err := botmanagement.Enable(ctx, r.client, serviceID); err != nil {
+			resp.Diagnostics.AddError("Error enabling bot_management", err.Error())
+			return
+		}
 		if _, err := botmanagement.UpdateConfiguration(ctx, r.client, serviceID, botmanagement.ConfigureInput{
 			ContentGuard: plan.ContentGuard.ValueString(),
 		}); err != nil {
--- a/internal/resources/productenablement/bot_management.go
+++ b/internal/resources/productenablement/bot_management.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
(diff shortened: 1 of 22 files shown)


internal/resources/productenablement/bot_management.go
```diff
@@ -128,14 +151,27 @@ func (r *BotManagementResource) Update(ctx context.Context, req resource.UpdateR
...
-	}); err != nil {
-		resp.Diagnostics.AddError("Error configuring bot_management", err.Error())
-		return
+	enabled := true
+	if !plan.Enabled.IsNull() {
+		enabled = plan.Enabled.ValueBool()
+	}
+
+	if enabled {
+		if _, err := botmanagement.UpdateConfiguration(ctx, r.client, serviceID, botmanagement.ConfigureInput{
```
philippschulte, line 164: This breaks the `false -> true` transition. After `enabled = false` disables Bot Management, changing it back to `true` only calls `UpdateConfiguration`, so the product is never enabled again.

The current provider calls `botmanagement.Enable` before `UpdateConfiguration`, which also matches the API's separate PUT enable and PATCH configuration endpoints. Could we preserve that behavior here and cover `true -> false -> true` in an acceptance test?

internal/resources/productenablement/ddos_protection.go
```diff
@@ -124,14 +147,27 @@ func (r *DDoSProtectionResource) Update(ctx context.Context, req resource.Update
...
-		Mode: plan.Mode.ValueString(),
-	}); err != nil {
-		resp.Diagnostics.AddError("Error configuring ddos_protection", err.Error())
-		return
+	enabled := true
+	if !plan.Enabled.IsNull() {
+		enabled = plan.Enabled.ValueBool()
+	}
+
+	if enabled {
```
philippschulte, line 155: Same issue here as Bot Management: after `enabled = false` calls `Disable`, switching back to `true` goes straight to `UpdateConfiguration` instead of enabling the product again first.

The current provider uses the product's `Enable` API whenever the desired state is enabled. Could we do the same here? NGWAF appears to have the same `false -> true` problem as well.

philippschulte (dismissed):
The `enabled` addition addresses the original drift-management gap well, and the simple false/true product lifecycle looks good.

Requesting changes for the configurable products: after `enabled = false` disables the product, switching back to `true` currently only calls `UpdateConfiguration` and never re-enables it. Bot Management, DDoS Protection, and NGWAF should preserve the existing provider behavior of calling the product's `Enable` API before applying configuration, with `true -> false -> true` acceptance coverage.

internal/resources/productenablement/bot_management.go
anthony-gomez-fastly: nice catch! fixed in `afb2f248`

internal/resources/productenablement/ddos_protection.go
anthony-gomez-fastly: fixed in `afb2f248`

internal/resources/productenablement/bot_management.go
```diff
@@ -101,19 +122,21 @@ func (r *BotManagementResource) Read(ctx context.Context, req resource.ReadReque
 
 	if _, err := botmanagement.Get(ctx, r.client, serviceID); err != nil {
 		if errors.IsNotFound(err) {
```
philippschulte: Should this also use `isProductDisabledError(err)` like the simple-product `Read` path?

The existing provider notes that these product APIs can return 400 when the product is not enabled. After `enabled = false`, that response should map to `state.Enabled = false`, not a read error. DDoS Protection and NGWAF appear to have the same pattern.

anthony-gomez-fastly:
Screenshot 2026-08-07 at 5 07 07 PM

internal/resources/productenablement/bot_management.go
anthony-gomez-fastly: fixed in `6e444a46`

philippschulte (approved):
Thanks for the updates. LGTM.

PR #1400, opened 2026-08-12
[CDTOOL-1509]  Add support for Blob Storage Logging to the Dual Model Provider

rcaril:
### Change summary

This PR adds support for the `logging_blobstorage` resource to the Dual Model Provider. 

 All Submissions:

### New Feature Submissions:

```
--- PASS: TestAccFastlyServiceLoggingBlobStorage_codecConflict (0.20s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_computeRejectsVCLOnlyFields (16.41s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_formatDefault (42.08s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_allAttr (45.03s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_basic (46.32s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_importBasic (48.35s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_computeConsistentAfterApply (52.16s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_clearToDefaults (62.73s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_versionUpdateInPlace (67.45s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingBlobStorageGzipCodec (69.71s)
--- PASS: TestAccFastlyServiceComputeAuto_withLoggingBlobStorage (72.70s)
--- PASS: TestAccFastlyServiceLoggingBlobStorage_placementUnsetVsNone (73.02s)
--- PASS: TestAccFastlyServiceCDNAuto_withMultipleLoggingBlobStorage (74.83s)
--- PASS: TestAccFastlyServiceCDNAuto_withBackendAndLoggingBlobStorage (75.33s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingBlobStorage (77.71s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingBlobStorageUpdate (82.09s)
PASS
ok  
```

### Changes to Core Features:


internal/resources/loggingblobstorage/expand.go
```diff
@@ -0,0 +1,182 @@
+package loggingblobstorage
+
+import (
+	fastly "github.com/fastly/go-fastly/v17/fastly"
+
+	"github.com/fastly/terraform-provider-fastly/internal/service"
+)
+
+func BuildCreateInput(serviceID string, version int, m NestedModel) *fastly.CreateBlobStorageInput {
+	input := &fastly.CreateBlobStorageInput{
```
anthony-gomez-fastly, line 13: i feel like there's a lot of duplication in here, is it possible to reuse some of these setters?

internal/resources/loggingblobstorage/expand.go
```diff
@@ -0,0 +1,182 @@
...
+	input.AccountName = fastly.NullString(service.StringValue(m.AccountName()))
+	input.SASToken = fastly.NullString(service.StringValue(m.SASToken()))
+	input.Path = new(service.StringValue(m.Path))
+	input.Period = fastly.NullInt(int(service.Int64Value(m.Period)))
+	input.CompressionCodec = new(service.StringValue(m.CompressionCodec))
+	// Only send an explicitly configured gzip_level. DefaultGzipLevel (-1) means
+	// unset: the API rejects requests that set both compression_codec and
+	// gzip_level, and it auto-manages the level when omitted.
+	if gl := service.Int64Value(m.GzipLevel); gl != DefaultGzipLevel {
+		input.GzipLevel = new(int(gl))
```
anthony-gomez-fastly: do we need to do the same thing for this as placement?

rcaril: `gzip_level` being `0` or `null` equivalates to a  specific level of compression. There wouldn't be a specific scenario where a user would "unset" `gzip_level`, as the `CompressionCodec` field handles this. 

For a in depth explanation of why we are keeping this logic from the prior provider, you can see this, very long. thread: https://fastly.slack.com/archives/C01E7FV8P5H/p1758555293782509 😃

internal/resources/loggingblobstorage/expand.go
rcaril: Good catch, updated in [aea3344](https://github.com/fastly/terraform-provider-fastly/pull/1400/commits/aea334432043aea9146eed1ea4424cbf91637d04).

anthony-gomez-fastly (approved):
LGTM!

internal/resources/loggingblobstorage/schema.go
```diff
@@ -0,0 +1,573 @@
...
+		"gzip_level": schema.Int64Attribute{
+			Optional: true,
+			Computed: true,
+			Default:  int64default.StaticInt64(DefaultGzipLevel),
+			// compression_codec and gzip_level are mutually exclusive; the API
+			// rejects a request that sets both. Validation runs against config,
+			// where an unset gzip_level is null rather than the -1 default, so
+			// this correctly fires only when both are set.
+			Validators: []validator.Int64{
+				gzipLevelCodecConflict{},
```
philippschulte, line 283: The current/legacy provider doesn't validate the `gzip_level` range, so I realize adding this would be a small validation behavior change. That said, since valid user/API values are `0–9` and `-1` is specifically an internal sentinel for “unset”, would it make sense to add `int64validator.Between(0, 9)` as a second validator here?

Validation runs against config before the `-1` default is applied, so an omitted `gzip_level` would continue to work as it does today, while explicitly configured out-of-range values (including the internal `-1` sentinel) would fail early instead of being accepted and interpreted internally as “unset”.

I also checked the S3 implementation, which currently doesn't apply a range validator either. I'm not sure whether we intentionally want to preserve the ability for users to explicitly configure `gzip_level = -1` for backwards compatibility. Semantically, though, if the intent is for `gzip_level` to be unset, it seems cleaner for the user to omit the attribute and let the provider apply its internal `-1` sentinel rather than exposing that implementation detail as a valid configurable value.

philippschulte (approved):
LGTM. I left one non-blocking comment regarding `gzip_level` validation, but otherwise this looks good to me.

internal/resources/loggingblobstorage/schema.go
rcaril: This is fair, I had some hesitations about this but it looks like the sentinel `-1` still works with the `0-9` validation. All of this is added as well as further coverage in tests in [2547ca8](https://github.com/fastly/terraform-provider-fastly/pull/1400/commits/2547ca8d2bc13d4bcb1ef89f1a780b6a8a240b7d).

PR #1401, opened 2026-08-12
[CDTOOL-1501] Add Edge Dictionary support to auto service resources

anthony-gomez-fastly:
### Change summary

Adds a `dictionary` nested block to `fastly_service_cdn_auto` and `fastly_service_compute_auto`, backed by a new internal/resources/dictionary package following the NestedModel reconciler pattern (see backend/cdnacl).

- write_only changes are applied via delete+recreate, since the Fastly API rejects updating write_only in place on an existing dictionary.
- force_destroy guards both removing a dictionary and changing write_only, since either discards the dictionary's items and they are not recoverable.
- Adds unit tests, acceptance tests for both service families, HCL fixtures, regenerated Registry docs, and an example walkthrough.
- make reconciliation generic with ACL's
 All Submissions:

### New Feature Submissions:

Screenshot 2026-08-12 at 7 42 01 PM
Screenshot 2026-08-12 at 7 47 54 PM
Screenshot 2026-08-12 at 7 48 35 PM


internal/resources/dictionary/schema.go
```diff
@@ -0,0 +1,225 @@
...
+// force_destroy, check emptiness, error) are shared with cdnacl via reconcile.GuardedRun.
+func ReconcileWithPrevious(ctx context.Context, client *fastly.Client, serviceID string, version int, previous, desired []NestedModel) error {
+	return reconciler.GuardedRun(
+		ctx, client, serviceID, version, previous, desired,
+		func(m NestedModel) bool { return service.BoolValue(m.ForceDestroy) },
+		func(prev, desired NestedModel, stillPresent bool) bool {
+			return !stillPresent || service.BoolValue(prev.WriteOnly) != service.BoolValue(desired.WriteOnly)
+		},
+		func(ctx context.Context, prev NestedModel) (bool, error) {
+			return isDictionaryEmpty(ctx, client, serviceID, service.StringValue(prev.DictionaryID))
```
philippschulte, line 198: What happens when the existing dictionary is `write_only = true`? We can't reliably inspect its items, but this path calls `ListDictionaryItems` whenever `force_destroy` is false. The current provider explicitly requires `force_destroy` for write-only dictionaries instead of trying the emptiness check. Could we preserve that behavior here and cover a `write_only: true -> false` (or removal) case without `force_destroy`?

philippschulte (dismissed):
The dictionary integration looks solid overall. I found one blocking edge case around existing `write-only` dictionaries.

internal/resources/dictionary/schema.go
anthony-gomez-fastly: good catch, addressed in `73df429a`

philippschulte (approved):
Thanks for the update. LGTM.

PR #1404, opened 2026-08-13
[CDTOOL-1526] Add Support for New Relic Logging in the Dual Model Provider

rcaril:
### Change summary

This PR adds support for the New Relic logging resource to the Dual Model Provider. 

 All Submissions:

### New Feature Submissions:

```
--- PASS: TestAccFastlyServiceLoggingNewRelic_clearToDefaults (66.53s)
--- PASS: TestAccFastlyServiceLoggingNewRelic_placementUnsetVsNone (81.40s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingNewRelic (140.41s)
--- PASS: TestAccFastlyServiceComputeAuto_withLoggingNewRelic (111.02s)
--- PASS: TestAccFastlyServiceCDNAuto_loggingNewRelicPlacementUnsetVsNone (170.84s)
--- PASS: TestAccFastlyServiceCDNAuto_withMultipleLoggingNewRelic (81.46s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingNewRelicUpdate (133.06s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingNewRelicRemoved (95.73s)
PASS
ok  
```

### Changes to Core Features:


internal/acceptance_tests/logging_newrelic_auto_test.go
anthony-gomez-fastly, line 1: is there a reason these are in 2 different files?

rcaril: That is a fantastic question, let me fix that :)

anthony-gomez-fastly (approved):
LGTM!

internal/resources/loggingnewrelic/list.go
```diff
@@ -0,0 +1,132 @@
...
+	id := serviceID + "-" + fmt.Sprintf("%d", version) + "-" + fastly.ToValue(n.Name)
+
+	model := Model{
+		NestedModel: FlattenToNestedModel(n),
+		ID:          types.StringValue(id),
+		Service:     types.StringValue(serviceID),
+		Version:     types.Int64Value(int64(version)),
+	}
+	diags.Append(result.Resource.Set(ctx, &model)...)
+	return diags
```
philippschulte, line 141: Should Compute list results normalize the VCL-only fields here? The standalone resource explicitly rejects `format/format_version/placement/response_condition` on Compute. As written, the list resource can emit resource data that cannot be used by the resource itself. Could we apply the same Compute normalization used by Create/Read/Import and cover it with a list test?

BigQuery appears to use a very similar list implementation so if you agree that we should address this then we might need to update the List() for other loggers as well.

philippschulte (changes requested):
The implementation looks solid. I left one blocking comment on the list resource.

internal/resources/loggingnewrelic/list.go
rcaril: You are correct, this was a gap that the previously implemented logging resources also shared. Fixed in [1dd5e3d](https://github.com/fastly/terraform-provider-fastly/pull/1404/commits/1dd5e3da12e5dfce329df1ac51e268f09f8aeedb) as well as added a List test. Let me know if you think the test should live elsewhere.

internal/acceptance_tests/logging_newrelic_auto_test.go
rcaril: @anthony-gomez-fastly I think I initially misunderstood what you were asking. The non `auto` test is for explicit mode, when it's fully implemented. This has been the pattern I have been using for the other logging endpoints in attempt to get as much done in one PR. 

Let me know your thoughts or i I wasn't following your question.

anthony-gomez-fastly: ah, i see. I think we should reverse the labeling then, since most (if not all) tests in this folder are auto mode unless specified AFAIK

rcaril: Ok, i'm going to open a new PR then for this and address this for all of the previously implemented logging resources as well.

philippschulte (approved):
Thanks for the update. LGTM.

PR #1406, opened 2026-08-13
[CDTOOL-1499] Add cache setting support to auto service resources

anthony-gomez-fastly:
### Change summary

Ports Cache Setting nested configuration to fastly_service_cdn_auto (CDTOOL-1499). ttl/stale_ttl default to 0 (Computed) so removing them from config reliably clears the remote value via an explicit Update, avoiding gzip's previous-state tracking for this simpler case.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-08-13 at 3 45 01 PM

### DIFF
--- a/internal/resources/cachesetting/schema.go
+++ b/internal/resources/cachesetting/schema.go
@@ -2,6 +2,7 @@ package cachesetting
 
 import (
 	"context"
+	"strings"
 
 	"github.com/fastly/terraform-provider-fastly/internal/reconcile"
 	"github.com/fastly/terraform-provider-fastly/internal/service"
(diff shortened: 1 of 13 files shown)


internal/resources/cachesetting/schema.go
```diff
@@ -0,0 +1,188 @@
...
+	return map[string]schema.Attribute{
+		"name": schema.StringAttribute{
+			Required:    true,
+			Description: "Unique name for this Cache Setting. Changing this attribute will delete and recreate the resource.",
+		},
+		"action": schema.StringAttribute{
+			Optional:    true,
+			Description: "One of `cache`, `pass`, or `restart`, as defined on Fastly's documentation under [\"Caching action descriptions\"](https://docs.fastly.com/en/guides/controlling-caching#caching-action-descriptions).",
+			Validators: []validator.String{
+				stringvalidator.OneOf("cache", "pass", "restart"),
```
philippschulte: nit: the current provider lowercased action during Create, so values like PASS could previously be accepted.

philippschulte (approved):
LGTM. Thank you!

internal/resources/cachesetting/schema.go
anthony-gomez-fastly: added in `7789a2b7`

PR #1408, opened 2026-08-14
[CDTOOL-1536] Add rate limiter support to fastly_service_cdn_auto

anthony-gomez-fastly:
### Change summary

Ports the VCL-only rate limiter (edge rate limiting) nested configuration from the legacy SDKv2 provider to the automatic-lifecycle service resource, following the existing NestedModel/reconciler pattern used by cache_setting and dictionary.

Behavior differs from the legacy provider in three ways, adopted for consistency with recently-ported resources:
- client_key and http_methods are native list(string) attributes instead of comma-separated strings, matching the gzip content_types/extensions convention
- action and logger_type validate case-insensitively (lowercased before hitting the API), matching cache_setting's action handling
- the computed ID attribute is named rate_limiter_id instead of the legacy ratelimiter_id, matching this codebase's underscore-everywhere convention

Update/Delete re-list rate limiters by name to resolve the current rate_limiter_id before calling the API, since Fastly only accepts an ID (never name+version) and cloning a service version assigns new IDs - this mirrors the legacy provider's documented approach.

Reconciled after dictionary in Create/Update, since uri_dictionary_name can reference a dictionary by name.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-08-14 at 1 41 34 PM
Screenshot 2026-08-14 at 1 44 37 PM


internal/resources/ratelimiter/schema.go
```diff
@@ -0,0 +1,507 @@
...
+		},
+		"penalty_box_duration": schema.Int64Attribute{
+			Required:    true,
+			Description: "Length of time in minutes that the rate limiter is in effect after the initial violation is detected.",
+		},
+		"rate_limiter_id": schema.StringAttribute{
+			Computed:    true,
+			Description: "Alphanumeric string identifying the rate limiter.",
+		},
+		"response": schema.SingleNestedAttribute{
```
rcaril, line 173: It looks like we may have lost the max item limit of `1` for this attribute here.

internal/resources/ratelimiter/schema.go
```diff
@@ -0,0 +1,507 @@
...
+					Required:    true,
+					Description: "HTTP Content-Type (e.g. `application/json`).",
+				},
+				"status": schema.Int64Attribute{
+					Required:    true,
+					Description: "HTTP response status code (e.g. `429`).",
+				},
+			},
+		},
+		"response_object_name": schema.StringAttribute{
```
rcaril, line 191: It looks like we don't have any validation on `main` for this, but the description does state that this is required if:

> `action` is `response_object`

If may be worth considering to add validation here.

internal/resources/ratelimiter/schema.go
```diff
@@ -0,0 +1,507 @@
...
+					"ftp", "gcs", "googleanalytics", "heroku", "honeycomb", "http", "https", "kafka",
+					"kinesis", "logentries", "loggly", "logshuttle", "newrelic", "openstack", "papertrail",
+					"pubsub", "s3", "scalyr", "sftp", "splunk", "stackdriver", "sumologic", "syslog",
+				),
+			},
+			PlanModifiers: []planmodifier.String{
+				caseInsensitiveState{},
+			},
+		},
+		"penalty_box_duration": schema.Int64Attribute{
```
rcaril, line 162: We should add validation here that was missed on `main`:
Min: `1`
Max: `60`

Source: https://github.com/fastly/api-documentation/blob/764415c037b59acf3051e41c5dde3ab95e65a282/src/rate_limiter.yaml#L182

internal/resources/ratelimiter/schema.go
```diff
@@ -0,0 +1,507 @@
...
+		Action:             actionPointer(desired.Action),
+		ClientKey:          listToStringSlice(desired.ClientKey),
+		FeatureRevision:    new(int(service.Int64Value(desired.FeatureRevision))),
+		HTTPMethods:        listToStringSlice(desired.HTTPMethods),
+		LoggerType:         loggerTypePointer(desired.LoggerType),
+		PenaltyBoxDuration: new(int(service.Int64Value(desired.PenaltyBoxDuration))),
+		Response:           responseType(desired.Response),
+		ResponseObjectName: optionalStringPointer(desired.ResponseObjectName),
+		RpsLimit:           new(int(service.Int64Value(desired.RpsLimit))),
+		URIDictionaryName:  optionalStringPointer(desired.URIDictionaryName),
```
rcaril, line 318: When `URIDictionaryName` is empty we correctly omit it here because the API rejects an explicit empty value, but omitting it also means Fastly retains the previously configured account name. The subsequent read would then return the old value while Terraform planned it as unset, which can produce drift/inconsistent state.

Could we handle the non-empty -> empty transition explicitly, likely by delete/recreate, and cover it with an acceptance test?

I confirmed this behavior locally as well when you remove a `uri_dictionary_name` entry from HCL:

```
  ~ rate_limiter {
            name                 = "my ERL"
          ~ rate_limiter_id      = "2JoCYl4V6yZ4llkPpPcsnE" -> (known after apply)
          - uri_dictionary_name  = "test" -> null
            # (8 unchanged attributes hidden)
        }
```
->

```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to fastly_service_cdn_auto.example, provider "provider[\"registry.terraform.io/fastly/fastly\"]" produced an unexpected new
│ value: .rate_limiter[0].uri_dictionary_name: was null, but now cty.StringVal("test").
```

internal/resources/ratelimiter/schema.go
```diff
@@ -0,0 +1,507 @@
...
+		ERLID:              id,
+		Name:               new(name),
+		Action:             actionPointer(desired.Action),
+		ClientKey:          listToStringSlice(desired.ClientKey),
+		FeatureRevision:    new(int(service.Int64Value(desired.FeatureRevision))),
+		HTTPMethods:        listToStringSlice(desired.HTTPMethods),
+		LoggerType:         loggerTypePointer(desired.LoggerType),
+		PenaltyBoxDuration: new(int(service.Int64Value(desired.PenaltyBoxDuration))),
+		Response:           responseType(desired.Response),
+		ResponseObjectName: optionalStringPointer(desired.ResponseObjectName),
```
rcaril, line 316: Same scenario here as for `URIDictionaryName`.

rcaril (dismissed):
Good start! A few items below that are blockers for now.

internal/resources/ratelimiter/schema.go
rcaril: The same solution was applied here if you would like some code to reference: https://github.com/fastly/terraform-provider-fastly/pull/1383#discussion_r3760329659

internal/resources/ratelimiter/schema.go
```diff
@@ -0,0 +1,507 @@
...
+			},
+			PlanModifiers: []planmodifier.String{
+				caseInsensitiveState{},
+			},
+		},
+		"penalty_box_duration": schema.Int64Attribute{
+			Required:    true,
+			Description: "Length of time in minutes that the rate limiter is in effect after the initial violation is detected.",
+		},
+		"rate_limiter_id": schema.StringAttribute{
```
rcaril, line 169: Can we add min / max values here too? 

Ref: https://github.com/fastly/api-documentation/blob/764415c037b59acf3051e41c5dde3ab95e65a282/src/rate_limiter.yaml#L160 

```
minimum: 10
maximum: 10000
```

rcaril: (I validated that the API still enforces this limit with a local test)

internal/resources/servicecdnauto/resource.go
```diff
@@ -855,6 +888,21 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
 		}
 		plan.Dictionary = dictionary.MatchOrder(dictionaries, plan.Dictionary)
 
+		// Rate limiters must be reconciled after dictionaries: uri_dictionary_name can reference
+		// a dictionary by name, and the Fastly API rejects a create that names one which doesn't
+		// exist yet in this version.
+		if err := ratelimiter.Reconcile(ctx, r.providerData.AutoClient(), serviceID, targetVersion, plan.RateLimiter); err != nil {
```
rcaril, line 927: Question on ordering the other direction: `dictionary.ReconcileWithPrevious` above (line 879) deletes any dictionary  present in `state.Dictionary` but missing from `plan.Dictionary`, and that runs before this call. If a single apply removes a `dictionary` block and updates / removes the `rate_limiter` that references it via `uri_dictionary_name` in the same config change, doesn't the dictionary get deleted while the not-yet-updated remote rate limiter still references it by name? 

The comment above explains why rate limiters must come after dictionaries for creation, but I  don't see the reverse handled for removal. Might be worth an acceptance test that deletes a dictionary referenced by an existing rate limiter to confirm this doesn't error (or silently leave a stale reference).

internal/resources/ratelimiter/schema.go
anthony-gomez-fastly: `SingleNestedAttribute` enforces this automatically i believe

rcaril: Confirmed that this is enforced with a local test, thanks for pointing that out!

internal/resources/ratelimiter/schema.go
anthony-gomez-fastly: added in `77e5fe2d`

internal/resources/ratelimiter/schema.go
anthony-gomez-fastly: added in `77e5fe2d`

internal/resources/ratelimiter/schema.go
anthony-gomez-fastly: great catch, added in `77e5fe2d`

internal/resources/ratelimiter/schema.go
anthony-gomez-fastly: added in `77e5fe2d`

internal/resources/ratelimiter/schema.go
anthony-gomez-fastly: added in `77e5fe2d`

internal/resources/servicecdnauto/resource.go
anthony-gomez-fastly: missed that, added in `77e5fe2d`

internal/resources/ratelimiter/schema.go
rcaril: I'm still able to replicate the `produced inconsistent ` note. I'll review this further Monday.

rcaril: Tested again - confirmed this is resolved now.

internal/resources/ratelimiter/schema.go
rcaril: I believe we still need a third clause here; `Update` seems to pass response through `responseType(desired.Response)`, which returns `nil` when the object is cleared - and `UpdateERLInput.Response` is the same `omitempty` tagged pointer field as the other two.  

I also confirmed that after a clean build, the same error still persists in my testing. 

The last clause can look something like this:

```
clearsResponse := desired.Response.IsNull() && remote.Response != nil return clearsURIDictionaryName || clearsResponseObjectName || clearsResponse
```

(Or perhaps cleaner with a helper)

internal/resources/servicecdnauto/resource.go
rcaril: Great, thank you!

internal/resources/ratelimiter/schema.go
anthony-gomez-fastly: added in `6e8ed7ab`

rcaril: Perfect, thank you!

rcaril (approved):
Thanks for all of the updates, this looks great now!

PR #1410, opened 2026-08-17
[CDTOOL-1502] Add director support to fastly_service_cdn_auto

anthony-gomez-fastly:
### Change summary

Implements CDTOOL-1502: nested director blocks reconciled through the provider-managed version lifecycle, including create/update/delete ordering against backend references and acceptance test coverage for create, update, backend-swap, and import.
 All Submissions:

### New Feature Submissions:

Screenshot 2026-08-17 at 2 57 26 PM


internal/resources/director/schema.go
```diff
@@ -0,0 +1,550 @@
...
+		service.Int64Value(n.Retries) == service.Int64Value(other.Retries) &&
+		service.StringValue(n.Shield) == service.StringValue(other.Shield) &&
+		service.StringValue(n.Type) == service.StringValue(other.Type)
+}
+
+func CommonAttributes() map[string]schema.Attribute {
+	return map[string]schema.Attribute{
+		"name": schema.StringAttribute{
+			Required:    true,
+			Description: "Unique name for this Director.",
```
rcaril: Can we keep the additional text from `main`? 
```suggestion
			Description: "Unique name for this Director. It is important to note that changing this attribute will delete and recreate the resource",
```

internal/resources/director/schema.go
```diff
@@ -0,0 +1,550 @@
...
+		"shield": schema.StringAttribute{
+			Optional:    true,
+			Computed:    true,
+			Default:     stringdefault.StaticString(DefaultShield),
+			Description: "Selected POP to serve as a \"shield\" for backends. Valid values for `shield` are included in the [`GET /datacenters`](https://developer.fastly.com/reference/api/utils/datacenter/) API response.",
+		},
+		"type": schema.StringAttribute{
+			Optional:    true,
+			Computed:    true,
+			Description: "Type of load balance group to use. One of `random`, `hash`, or `client`. Default `random`.",
```
rcaril: I don't believe the enums listed here are correct for the desc nor validator. 

https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_vcl#type-2

https://github.com/fastly/api-documentation/blob/0f731789b74e90b12809d763ca4218c9809c6a8e/src/director.yaml#L163

rcaril (dismissed):
Looks good! Some minor items in `schema.go` pending below

internal/resources/director/schema.go
anthony-gomez-fastly: for the first link there are the types for director https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_vcl#nested-schema-for-director

for the old provider we used numbers instead of the naming, but they are the same values as here

rcaril: I agree that the string value is a more clear user experience, but we may want to maintain what aligns with the API and prior behavior in `main`? 

In my mind, this goes along with "minimal breaking changes" approach that was discussed here:
https://github.com/fastly/terraform-provider-fastly/pull/1383#discussion_r3731394806 

@philippschulte - what are your thoughts?

anthony-gomez-fastly: ah i missed that convo. happy to do either tbh

anthony-gomez-fastly: i'll change it to numbers

rcaril: Would it be too much to use a fallback for the numeric values to maintain the "lets user changes per HCL" approach?

kpfleming: That's my preference too - improve the user experience by using values with human-relatable meanings, but don't break existing configurations any more than necessary.

philippschulte: I believe this is a little bit of a different use case as in https://github.com/fastly/terraform-provider-fastly/pull/1383#discussion_r3731394806 . A fallback for the numeric values is a great idea if possible otherwise I always support if something improves the rewrite. So the strings are an improvement and it is fine that's a breaking change. The other discussion was about an env variable and my concern was that that env var could have been used elsewhere and therefore I recommended a fallback. Here I don't believe we need it but it's fine if we implement it however, I believe it adds unnecessary complexity. I hope that makes sense.

anthony-gomez-fastly:
Screenshot 2026-08-18 at 3 34 48 PM

internal/resources/servicecdnauto/resource.go
```diff
@@ -943,9 +984,57 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
...
+		// before the director that references it, but a backend delete must wait until any
+		// director no longer referencing it has already been updated - otherwise the API rejects
+		// deleting a backend still named by a director's association. So: create/update backends
+		// first (satisfies the create direction), reconcile directors fully (any director losing
+		// a backend reference is updated/deleted here), then delete backends no longer desired
+		// (now safe - no director still references them). Services with no directors configured
+		// before or after this update skip straight to the single-pass Reconcile, which deletes
+		// before creating and so never transiently holds an extra backend against the service's
+		// backend-count limit.
+		if len(plan.Director) > 0 || len(state.Director) > 0 {
```
rcaril, line 999: This may need adjustment when we are completed with with `_auto` resources. I believe this defers  backend deletes whenever any director exist - however we'll most likely continue to update this list, so we can tackle this later when we are validating the full order of operations.

internal/resources/director/schema.go
anthony-gomez-fastly: i'll see if i can fall back

internal/resources/servicecdnauto/resource.go
anthony-gomez-fastly: i have a follow up PR that does larger reconcile changes now that i've done a few of these 😂

anthony-gomez-fastly:
Screenshot 2026-08-18 at 3 50 03 PM

internal/resources/director/schema.go
anthony-gomez-fastly: done in `905ba4f6`

internal/resources/director/schema.go
```diff
@@ -0,0 +1,550 @@
...
+		return
+	}
+
+	newList, diags := types.ListValue(req.PlanValue.ElementType(ctx), newElems)
+	resp.Diagnostics.Append(diags...)
+	if diags.HasError() {
+		return
+	}
+	resp.PlanValue = newList
+}
```
philippschulte, line 270: In the current provider, `type` had `Default: 1`, so removing `type = 3` from config would reset the director back to the default random type. With the new list-level plan modifier, an omitted `type` carries forward the prior state value by name, and the acceptance test codifies that dropping `type = 3` from config keeps `3`.

That seems like a user-visible behavior change from the current provider. Could we limit the sticky preservation to prior `type == 2` for imported/out-of-band round_robin directors, and otherwise default omitted config to `1`?

anthony-gomez-fastly: handled in `db2ba50d`

internal/resources/director/schema.go
rcaril: Thanks!

rcaril (approved):
Looks good to me, thanks for all of the updates!

philippschulte (approved):
This looks great. Thank you!

PR #1414, opened 2026-08-18
[CDTOOL-1520] Add support for HTTPS Logging to Dual Model Provider

rcaril:
### Change summary

This PR adds support for HTTPS logging to the Dual Model Provider.

 All Submissions:

### New Feature Submissions:

```
--- PASS: TestAccFastlyServiceLoggingHTTPS_gzipCodecConflict (0.38s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_computeRejectsVCLOnlyFields (22.89s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_importBasic (34.54s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_formatDefault (35.54s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_computeConsistentAfterApply (36.23s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_basic (40.63s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_update (51.21s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_clearToDefaults (56.49s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_versionUpdateInPlace (64.99s)
--- PASS: TestAccFastlyServiceLoggingHTTPS_placementUnsetVsNone (68.37s)
--- PASS: TestAccFastlyServiceComputeAuto_withLoggingHTTPS (75.03s)
--- PASS: TestAccFastlyServiceCDNAuto_withBackendAndLoggingHTTPS (78.11s)
--- PASS: TestAccFastlyServiceCDNAuto_withMultipleLoggingHTTPS (80.48s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingHTTPSRemoved (95.36s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingHTTPS (96.53s)
--- PASS: TestAccFastlyServiceCDNAuto_withLoggingHTTPSUpdate (104.03s)
PASS
ok   
```

### Changes to Core Features:


internal/resources/logginghttps/validators.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// httpsURLRe requires url to use the https:// scheme, matching the Fastly
+// API's documented requirement ("Must use HTTPS").
+var httpsURLRe = regexp.MustCompile(`^https://`)
```
philippschulte: Could we make the URL validator equivalent to the SDK provider’s `validation.IsURLWithHTTPS` behavior?

The current regex only checks that the value starts with `https://`, so malformed values such as `https://` can pass provider validation and fail later at the API. The SDK provider used `validation.IsURLWithHTTPS`, which validates both URL shape and HTTPS scheme. We should either reuse an equivalent framework validator or add a custom validator that parses the URL and requires `scheme == "https"` plus a non-empty host.

internal/resources/logginghttps/schema.go
```diff
@@ -0,0 +1,603 @@
...
+			Default:  stringdefault.StaticString(DefaultMethod),
+			Validators: []validator.String{
+				stringvalidator.OneOf("POST", "PUT"),
+			},
+			Description: "HTTP method used for request. Can be either `POST` or `PUT`. Default `POST`.",
+		},
+		"period": schema.Int64Attribute{
+			Optional:    true,
+			Computed:    true,
+			Default:     int64default.StaticInt64(DefaultPeriod),
```
philippschulte: Could we add lower-bound validation for the numeric fields here?

`period`, `request_max_bytes`, and `request_max_entries` are all non-negative API values. The public docs allow `0` for `period` and the request limit defaults, while the SDK provider was stricter for `period` with `IntAtLeast(1)`. Either way, the current framework schema appears to allow negative values through plan validation and defer the failure to the API.

I think we should add at least `int64validator.AtLeast(0)` for these fields, and decide separately whether `period = 0` should be allowed for public-doc parity or rejected for SDK parity.

philippschulte (changes requested):
The implementation looks solid! I only found two small things we should improve.

internal/resources/logginghttps/schema.go
rcaril: I agree, I added the `.AtLeast(0)` validator to be at minimum compliant with the API spec - good catch!

internal/resources/logginghttps/validators.go
rcaril: Makes sense, updated in [82887c7](https://github.com/fastly/terraform-provider-fastly/pull/1414/commits/82887c7a25ce8f820508557173957e10fdd2e03e).

philippschulte (approved):
LGTM. Thanks!

PR #1416, opened 2026-08-18
[CDTOOL-1505] Add header support to fastly_service_cdn_auto

anthony-gomez-fastly:
### Change summary

Ports the legacy header block (fastly/block_fastly_service_header.go) as a nested config block on the automatic-lifecycle CDN service resource. Header is VCL-only in the legacy provider (registered only for vclAttributes, never for Compute), so it is wired into servicecdnauto only, not servicecomputeauto.

Reconciled after ACL/condition in Create and Update, since a header's cache_condition/request_condition/response_condition can reference a condition by name.

 All Submissions:

### New Feature Submissions:

Screenshot 2026-08-18 at 4 35 19 PM

### DIFF
--- a/internal/resources/header/schema.go
+++ b/internal/resources/header/schema.go
@@ -1,0 +1,295 @@
+package header
+
+import (
+	"context"
+
+	"github.com/fastly/terraform-provider-fastly/internal/reconcile"
+	"github.com/fastly/terraform-provider-fastly/internal/service"
+
+	fastly "github.com/fastly/go-fastly/v17/fastly"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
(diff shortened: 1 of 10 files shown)


internal/resources/header/schema.go
```diff
@@ -0,0 +1,295 @@
...
+			Optional:    true,
+			Computed:    true,
+			Default:     stringdefault.StaticString(DefaultRequestCondition),
+			Description: "Name of already defined `condition` to apply. This `condition` must be of type `REQUEST`.",
+		},
+		"response_condition": schema.StringAttribute{
+			Optional:    true,
+			Computed:    true,
+			Default:     stringdefault.StaticString(DefaultResponseCondition),
+			Description: "Name of already defined `condition` to apply. This `condition` must be of type `RESPONSE`.",
```
rcaril: Can we keep the rest of the description from `main`? 

https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_vcl#response_condition-1

rcaril (approved):
Great work! Just one nit below.

internal/resources/header/schema.go
anthony-gomez-fastly: fixed in `3fb33e5f`

PR #1420, opened 2026-08-26
Bcunningham/ngwaf resource fixes

BrooksCunningham:
### Change summary

`fastly-terraformer` generated `import.tf`/`generated.tf` for existing `fastly_ngwaf_thresholds` resources, then `terraform apply` failed with:

```
Error: expected interval to be one of [60 600 3600], got 0
Error: expected limit to be in the range (1 - 10000), got 0
Error: expected length of name to be in the range (3 - 50), got
```

The NGWAF API legitimately returns zero-value defaults (`0` / `""`) for `interval`, `limit`, and `name` on Read for thresholds where those fields were never explicitly set — the same pattern `duration` already handles by defaulting `0` to `86400` in `resourceFastlyNGWAFThresholdsRead`. Unlike `duration`, the schema's `ValidateDiagFunc` for `interval`, `limit`, and `name` didn't accept these zero values, so Terraform rejected its own generated config (and would reject a hand-written config or `refresh`-driven plan for any resource in that state).

Fix widens the three validators to accept the API's zero-value defaults alongside the existing valid ranges:
- `interval`: `IntInSlice([60, 600, 3600])` → `IntInSlice([0, 60, 600, 3600])`
- `limit`: `IntBetween(1, 10000)` → `IntBetween(0, 10000)`
- `name`: `StringLenBetween(3, 50)` → `StringLenBetween(0, 50)`

Descriptions updated to document the zero-value case. Docs regenerated via `make generate-docs` (only `docs/resources/ngwaf_thresholds.md` changed). No changes to Create/Update/Read logic — the API round-trip was already correct; only the schema's acceptable-value contract was too strict.

All Submissions:

### Changes to Core Features:

Added `TestFastlyNGWAFThresholds_zeroValueValidators` (unit-level, no live API) locking in that `interval=0`, `limit=0`, `name=""` now validate, while out-of-range values (`interval=1234`, `limit=10001`, 51-char `name`) still fail. Existing `TestAccFastlyNGWAFThresholds_validate` steps are untouched — still cover the normal non-zero path.

Output:
```
=== RUN   TestFastlyNGWAFThresholds_zeroValueValidators
--- PASS: TestFastlyNGWAFThresholds_zeroValueValidators (0.00s)
PASS
ok      github.com/fastly/terraform-provider-fastly/fastly     0.851s
```
`go build ./...` and `go vet ./fastly/...` also clean. `TestAccFastlyNGWAFThresholds_validate` requires `FASTLY_API_KEY`/live NGWAF workspace — not run in this environment; unchanged by this diff.

### User Impact

Terraform no longer rejects valid, API-returned state for `fastly_ngwaf_thresholds` resources whose `interval`, `limit`, or `name` are at their zero-value default. Unblocks `terraform apply`/`plan -generate-config-out` and `refresh` for any threshold in that state, notably ones imported via `fastly-terraformer`.

No breaking changes — this only widens (never narrows) accepted values. Note: widening `name`'s minimum from 3 to 0 also permits 1–2 character names going forward, not just empty string; flagging in case maintainers want the stricter `validation.Any(StringLenBetween(0,0), StringLenBetween(3,50))` variant instead — happy to switch if preferred.


kpfleming:
We made an explicit choice when building this resource to disallow the '0' value because it is not useful for users of the provider. A person reading the HCL and seeing `interval` set to '0' would have no idea what that means and would be forced to read documentation to see what the effective value actually is. This is an API design flaw but it is what we have to deal with.

As a result, we will not modify the provider to allow '0' values for these attributes. If there are changes required in the provider to avoid these '0' values being written out by `terraform plan -generate-config-out` then we may consider making those changes. I suspect that you might be able to work around this problem when using `fastly-terraformer` by running `terraform apply` with the import blocks in place *before* running `terraform plan -generate-config-out`, as that may result in Terraform fully populating the state correctly before generating the HCL. It would also be informative to see what value is in the Terraform state file after the apply operation which imported the resource.

BrooksCunningham:
> This is an API design flaw but it is what we have to deal with.
@kpfleming Thank you for comment and the review. 

I do agree with what you are saying about this being an overall API design flaw. I would personally rather see the fix implemented in the API since that would fix this issue and maybe other challenges. What needs to happen to flag this flaw in the API? I do not know even what team this would go to.

kpfleming:
It's the Security API team - ask in Slack if you need a specific person's name.

rcaril:
Closing this PR as per the above conversation.

PR #1424, opened 2026-09-05
feat(tls/certificate): support allow_untrusted_root

leoromanovsky:
## Motivation

The Fastly Custom TLS Certificate API accepts `allow_untrusted_root` on create and update requests. The `fastly_tls_certificate` resource does not expose that field, so Terraform cannot upload a certificate whose chain is not signed by a root trusted by Fastly, even when the caller explicitly wants the API to accept that chain.

## Changes

- Add the optional `allow_untrusted_root` argument to `fastly_tls_certificate`, defaulting to `false`.
- Pass the argument to both create and update API requests.
- Add unit coverage for create/update requests with both `true` and `false` values.
- Update the existing self-signed acceptance fixtures, example, generated resource documentation, and changelog.

Validation:

- `go test ./fastly -run '^TestResourceFastlyTLSCertificateAllowUntrustedRoot$' -count=1`
- `make test`
- `make tfproviderlintx`
- `make generate-docs`
- `git diff --check`

## Decisions

- Keep the argument write-only because the Fastly API client exposes it only on create/update inputs and does not return it when reading a certificate.
- Preserve the current API request shape when disabled: the Go client omits `false`, while Terraform records the explicit default.
- Reuse the existing self-signed certificate acceptance paths instead of adding a separate fixture. Acceptance tests were not run because they require Fastly credentials.
- `make validate-docs` currently reports legacy frontmatter errors across the existing generated documentation, and `make lint` is blocked by the repository-pinned golangci-lint v2.4.0 being built with Go 1.24 while this checkout targets Go 1.26.5; neither failure is introduced by this diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### ENHANCEMENTS:
 
+- feat(tls/certificate): add `allow_untrusted_root` support to `fastly_tls_certificate` ([#1424](https://github.com/fastly/terraform-provider-fastly/pull/1424))
+
 ### BUG FIXES:
 
 ### Dependencies
--- a/docs/resources/tls_certificate.md
+++ b/docs/resources/tls_certificate.md
@@ -47,9 +47,10 @@ resource "fastly_tls_private_key" "key" {
 }
 
 resource "fastly_tls_certificate" "example" {
-  name = "tf-demo"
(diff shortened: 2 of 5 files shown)


rcaril:
Hi @leoromanovsky  - thanks for your contribution! 
The feature associated with `allow_untrusted_root` has not been made generally available yet, so we'll be aiming to tackle this as a larger effort in the future.

